PR #9, opened 2015-01-20
Implement Itertools.merge(other).

Stebalien:
`Itertools.merge` merges two sorted iterators (with items implementing Ord) into a single sorted iterator.

### DIFF
--- a/src/adaptors.rs
+++ b/src/adaptors.rs
@@ -6,7 +6,8 @@
 
 use std::mem;
 use std::num::Int;
-use std::iter::Fuse;
+use std::usize;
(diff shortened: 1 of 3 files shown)


bluss:
Thank you, looks great.  What do you think about relaxing it to PartialOrd? We only need one comparator, `<=`, so we can give the user the flexibility.

Stebalien:
Good point. I used Ord because everything else in the standard library that depends on sort-order appears to use it however, PartialOrd is explicitly for types that can be compared for sort-order.

bluss:
Thank you!

PR #12, opened 2015-02-03
refactor tuple zip macros and traits

oli-obk:
i simplified the macros a little and got rid of mutable variables.
if this commit is doing too much at once, i can split it up

### DIFF
--- a/src/ziptuple.rs
+++ b/src/ziptuple.rs
@@ -37,13 +37,8 @@ impl<T> Zip<T> where Zip<T>: Iterator
     }
 }
 
-// Macro to turn `t.$idx`, where $idx:tt, into an expression.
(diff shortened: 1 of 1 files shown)


bluss:
Guess what I have locally that I didn't push yet

```
commit daa1dd86fbf546f890e4198b89cbd9047b8b0880 (HEAD, master)
Author: root <root@localhost>
Date:   Tue Feb 3 01:28:59 2015 +0100

    Simplify the Zip macro impl, using ideas from ker on the forum.
    
    Simplification following ker, see
    http://internals.rust-lang.org/t/pre-rfc-replace-iteratorext-zip-with-tuple-iteration/1518
```

but I'll use your commit. Thank you!

PR #14, opened 2015-02-05
Fix Merge adaptor

aepsil0n:
Peekable is no longer generic over its item type, but it is inferred from the iterator instead.


bluss:
Thank you! I've already fixed this and I'm preparing push & version bump.

aepsil0n:
Oh, that's fine. Only 4 lines of code. ^^

PR #15, opened 2015-02-13
Add algorithm for inplace-mutation and dropping without reallocation

oli-obk:
this was created from http://stackoverflow.com/questions/28318886/how-do-i-partially-fold-a-vec-in-place

name-bikeshedding welcome.

instead of implementing it for `&mut [T]` i could probably implement it for `RandomAccessIterator` not sure on that

### DIFF
--- a/src/inplacemut.rs
+++ b/src/inplacemut.rs
@@ -1,0 +1,118 @@
+use std::vec::Vec;
+
+pub enum Combine<T> {
+    Keep(usize), // n >= 1
+    InsertAndDrop(T, usize), // insert 1 element, remove n elements, n >= 1
+}
+
+#[derive(Debug, PartialEq)]
+pub enum CombineError {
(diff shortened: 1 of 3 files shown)


bluss:
Still wrapping my head around this. It's sort of a parallel to .batching(). Is this a practical API?

oli-obk:
think of this PR as an RFC ;) I'm sure you could get a similar behavior for Iterators implementing `RandomAccessIterator`. But the ability to match on slices creates much more readable code. So maybe we need a `SliceIterator` that has `O(1)` operations `to_slice` and implement `SliceExt` for `SliceIterator`. Then you could use `batching` to step the iterator forward whenever you want, and otherwise match on slices.

bluss:
I think we should stick to just iterator related functions in itertools for now. Feel free to open new PRs with wild ideas if you have :)

PR #20, opened 2015-02-25
Add an iterator adapter that allows peeking of multiple values

flo-l:
I find this quite useful, as the ```.peekable()``` adapter from ```std``` is often not enough (it can only peek one element).

The name is far from perfect, I can rename the whole thing if you come up with something better! ;)

### DIFF
--- a/src/adaptors.rs
+++ b/src/adaptors.rs
@@ -578,6 +578,7 @@ impl<K, I: Iterator> Iterator for EnumerateFrom<I, K> where
     }
 }
 
+/// An Iterator adaptor that allows the user to peek at multiple *.next()* values without advancing itself.
 pub struct MultiPeek<I: Iterator> {
     iter: Fuse<I>,
     buf: Vec<I::Item>,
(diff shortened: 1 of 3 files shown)


bluss:
Thank you, I think it's a good idea -- is the stateful .peek() API useful in that shape? Seems a bit weird, what about something like an indexed peek: `.peek(&mut self, index: usize) -> Option<&I::Item>` ?

And to be correct with the iterator protocol, we need an exhaustion flag for the iterator, i.e. use Fuse.

flo-l:
I thought about that, but decided to build it the way I submitted it in the end because it resembles how ```.next()``` works. The first name I came up with was actually ```.fake_next()```.

I however have no strong opinion on that, so if you think it would be of more use with an index I'll edit my code, no problem.

I'm sorry but I don't understand the part about the exhaustion flag. The MultiPeek iterator yields None if the underlying iterator yields None. And afterwards the behaviour is unspecified anyway.

*The Iterator protocol states that an iterator yields a (potentially-empty, potentially-infinite) sequence of values, and returns None to signal that it's finished. The Iterator protocol does not define behavior after None is returned. A concrete Iterator implementation may choose to behave however it wishes, either by returning None infinitely, or by doing something else.*

[source](http://doc.rust-lang.org/std/iter/trait.Iterator.html)

flo-l:
Ok I got it, good catch, thanks!

bluss:
Ok *.peek()* is fine. I didn't think about fuse for that reason, just for the case when you peek a None first, then call *.next()*. I suppose it's a bit of a gray area given that you modify the iterator with other methods.

flo-l:
awesome! anything else you're not happy with?

hmm it's indeed a grey area. But with the fuse in place it's now at least really close to conformity!

bluss:
I'm happy either way, I promise. I'll merge this another night when I have time to clear up some other WIP in itertools.

If you want to, add docs to all public items here (the struct and methods) and impl Clone, but if you don't I'll do it later.

flo-l:
Nice!

I've added docs and the Clone impl (I hope I've done the latter in the right way).

Thanks!

bluss:
Thank you!

PR #22, opened 2015-02-28
allow putting back multiple values with PutBack

flo-l:
Another thing I would find useful from time to time. I modified the original PutBack, so this is a breaking change, if you want me to I could also create a new struct.

I'm not sure if you even want this in your library, it wasn't much work so I'm totally ok with abandoning the idea again ;)


bluss:
Needs to be a separate struct, putback is purposefully simple.

flo-l:
Closed in favour of #23.

PR #23, opened 2015-02-28
Add Feed, an iterator adapter that lets one set multiple elements to be yielded next by itself

flo-l:
Remake of #22, it's a new struct now!

I borrowed the name from ruby, suggestions are as always more than welcome!

### DIFF
--- a/src/adaptors.rs
+++ b/src/adaptors.rs
@@ -188,7 +188,7 @@ impl<I: Iterator> PutBackN<I>
     #[inline]
     pub fn new(it: I) -> Self
     {
-        Feed{top: vec![], iter: it}
+        PutBackN{top: vec![], iter: it}
     }
 
     /// Puts x in front of the iterator.
@@ -223,7 +223,7 @@ impl<I: Iterator> Clone for PutBackN<I> where
 {
     fn clone(&self) -> Self
     {
-        clone_fields!(Feed, self, top, iter)
+        clone_fields!(PutBackN, self, top, iter)
     }
 }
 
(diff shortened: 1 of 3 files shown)


bluss:
Looks fine. Where is the Ruby version? Is there precedent for the way multiple values are fed? I would have expected them to be pushed to the front.

flo-l:
The ruby version: http://ruby-doc.org/core-1.9.3/Enumerator.html#method-i-feed

It supports just one element.

Yeah I also wasn't sure about the ordering. But I decided to push it to the back of the feed-vec so that you can write the following:

```rust
use itertools::Itertools;
let v: Vec<u8> = vec![];
let mut it = v.feedable();
it.feed(2u8);
it.feed(3u8);
assert_eq!(it.next(), 2u8);
assert_eq!(it.next(), 3u8);
```

My mental model is the iterator being a pipe into which I throw items...the one I throw in first comes out first. But I can also change this if you prefer!

bluss:
Oh so the Ruby feed doesn't do this at all. It's like `.send()` on a generator in Python, it feeds a value back into a coroutine. That kind of thing doesn't exist in Rust.

So I think we should rename this thing.

Anyway about the ordering, the reverse order feels more natural.

flo-l:
Ordering reversed..

Any suggestions for the new name? PutBackMult sounds a bit clumsy :(

bluss:
What about PutBackN and method .put_back?

flo-l:
ping @bluss 

sorry it took me some time to react!

bluss:
Looks good. I'll merge it when I can (I'm away tomorrow probably).

Can I ask you to squash the fixes together (if you want) and actually remove the method in the Itertools trait? Some adaptors just aren't worth using a method on IMO. PutBack doesn't have a method, and I think it's logical given that you never method chain after you create it. It is not like map() where you want to create a pipeline.

flo-l:
reopened as #29

PR #24, opened 2015-03-02
Take advantage of new where clause feature.

Stebalien:
... `where I: Trait` .... `I::AssocType` now works. This is purely a style cleanup but touches everything. I won't be offended if you don't want it.


bluss:
It's great!

PR #27, opened 2015-04-16
Add merge_by

ehiggs:
Doesn't yet compile.

### DIFF
--- a/src/adaptors.rs
+++ b/src/adaptors.rs
@@ -9,6 +9,7 @@ use std::mem;
 use std::num::One;
 #[cfg(feature = "unstable")]
 use std::ops::Add;
+use std::cmp::Ordering;
 use std::usize;
 use std::iter::{Fuse, Peekable};
 use super::Itertools;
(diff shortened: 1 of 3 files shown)


src/adaptors.rs
```diff
@@ -465,49 +466,62 @@ impl<I> Iterator for Step<I>
...
     J: Iterator<Item=I::Item>,
+    F: Fn(&I::Item, &I::Item) -> Ordering
 {
     type Item = I::Item;
 
     fn next(&mut self) -> Option<I::Item> {
         if match (self.a.peek(), self.b.peek()) {
-            (Some(a), Some(b)) => a <= b,
+            (Some(a), Some(b)) => {
+                match (self.cmp)(&a, &b) {
```
bluss: I think these `&` on a and b are redundant.

tests/tests.rs
```diff
@@ -322,6 +322,35 @@ fn merge() {
 }
 
 #[test]
+fn merge_by() {
+    let odd : Vec<(u32, &str)> = vec![(1, "hello"), (3, "world"), (5, "!")];
+    let even = vec![(2, "foo"), (4, "bar"), (6, "baz")];
+    let expected = vec![(1, "hello"), (2, "foo"), (3, "world"), (4, "bar"), (5, "!")];
+    let results = odd.iter().merge_by(even.iter(), |a, b|{ a.0.cmp(&b.0)});
+    for (r, e) in results.zip(expected) {
```
bluss: there's assert_iters_equal for this if you want to. It actually catches something here -- result and expected are different length! :)

     assert_iters_equal(results, expected.iter());

tests/tests.rs
```diff
@@ -322,6 +322,35 @@ fn merge() {
...
+    use std::collections::BTreeMap;
+    let mut bt1 = BTreeMap::new();
+    bt1.insert("hello", 1);
+    bt1.insert("world", 3);
+    let mut bt2 = BTreeMap::new();
+    bt2.insert("foo", 2);
+    bt2.insert("bar", 4);
+    let results = bt1.into_iter().merge_by(bt2.into_iter(), |a, b|{a.0.cmp(&b.0)});
+    let expected = vec![("bar", 4), ("foo", 2), ("hello", 1), ("world", 3)];
+    for (r, e) in results.zip(expected) {
```
bluss: assert_iters_equal(results, expected.into_iter());

bluss:
Great, this is a useful addition to itertools. Small comments in the patch. Please update the PR title & message & commit message since that's the information that's associated with the change when it ends up in git.

tests/tests.rs
ehiggs: As I'm comparing tuples, `assert_iters_equal` doesn't work.

bluss: That's not true (I tried it).

PR #28, opened 2015-05-10
Iterator adaptor for iterating through Combinations

AngelOfSol:
I needed an iterator that went through combinations out of a vec so I made this.  I don't know if the name is all too good though.

### DIFF
--- a/src/adaptors.rs
+++ b/src/adaptors.rs
@@ -714,3 +714,59 @@ impl<'a, I, F> Iterator for TakeWhileRef<'a, I, F> where
         (0, hi)
     }
 }
+
+/// An iterator to iterate through all the combinations of pairs in a **Clone**-able iterator.
+pub struct CombinatePair<I: Iterator> {
+    iter: I,
+    next_iter: Option<I>,
+    val: Option<I::Item>,
+}
+impl<I> CombinatePair<I> where I: Iterator {
+    /// Create a new **CombinatePair** from a clonable iterator.
+    pub fn new(iter: I) -> CombinatePair<I> {
+        CombinatePair { 
+            iter: iter, 
+            next_iter: None, 
+            val: None,
+        }
+    }
+}
+
+impl<I> Iterator for CombinatePair<I> where I: Iterator + Clone, I::Item: Clone{
+    type Item = (I::Item, I::Item);
+    fn next(&mut self) -> Option<Self::Item> {
+        // not having a value means we iterate once more throug the first iterator
+        if self.val.is_none() {
+            self.val = self.iter.next();
+            self.next_iter = Some(self.iter.clone());
+        }
+        // if its still none, we're out of values
+        if self.val.is_none() {
+            return None;
+        }
+        
+
+        let ret_ele = self.val.clone().unwrap();
+        match self.next_iter.as_mut().unwrap().next() {
+            Some(ref x) => {
+                return Some((ret_ele, x.clone()));
+            },
+            None => {
+                self.val = None;
+            }
+        }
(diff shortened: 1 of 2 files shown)


bluss:
I like it, we should have this! Nice attention to examples & docs following the rest of itertools. I'll leave comments inline in the patch.

It's called [combinations here](https://docs.python.org/2/library/itertools.html#itertools.combinations) so I think that's a fine name, what do you think? How we generalize this to n-tuples I don't know yet (we needn't worry).

src/adaptors.rs
```diff
@@ -714,3 +714,59 @@ impl<'a, I, F> Iterator for TakeWhileRef<'a, I, F> where
         (0, hi)
     }
 }
+
+/// An iterator to iterate through all the combinations of pairs in a **Clone**-able iterator.
+pub struct CombinatePair<I: Iterator> {
```
bluss: derive Clone

src/adaptors.rs
```diff
@@ -714,3 +714,59 @@ impl<'a, I, F> Iterator for TakeWhileRef<'a, I, F> where
...
+            Some(ref x) => {
+                return Some((ret_ele, x.clone()));
+            },
+            None => {
+                self.val = None;
+            }
+        }
+        // try again if we ran out of values in the second iterator
+        self.next()
+    }
```
bluss: I'm sure we can write this in a way that .unwrap() isn't needed. For example `let elt = match self.val { Some(ref x) => x.clone(), None => return None };`.  Why not use the abbreviation `elt` like the other code too.

src/adaptors.rs
```diff
@@ -714,3 +714,59 @@ impl<'a, I, F> Iterator for TakeWhileRef<'a, I, F> where
...
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let (_, hi) = self.iter.size_hint();
+        match self.next_iter {
+            Some(ref i) => {
+                let (low, hi_next) = i.size_hint();
+                (low, hi.and_then(|x| hi_next.and_then(|y| Some(x * y))))
+            },
+            None => (0, hi),
+        }
```
bluss: These calculations always make my head hurt. Can you use some of the functions in itertools::size_hint? The exact number of elements from the start is *n* &times; (*n* - 1) / 2, but it's ok if we don't get it exact right now. (As a rule, we don't change the iterator representation to get better size_hint, so if we don't know, we don't know.)

src/lib.rs
```diff
@@ -655,6 +656,23 @@ pub trait Itertools : Iterator {
...
+    /// Iterator element type is **(Self::Item, Self::Item)**.
+    ///
+    /// ```
+    /// use itertools::Itertools;
+    ///
+    /// let it = (1..5).combinate_pairs();
+    /// assert!(itertools::equal(it, vec![(1, 2), (1, 3), (1, 4), (2, 3), (2, 4), (3, 4)]));
+    /// ```
+    fn combinate_pairs(self) -> CombinatePair<Self> where
+        Self: Sized + Clone
```
bluss: We should probably put `Self::Item: Clone` here just to give users better help messages.

AngelOfSol:
In response to generalizing it to n-tuples, I think variadics are needed before that becomes easy.  In the mean time though The second iterator type could be a different type variable, and then you could chain them as CombinatePair<CombinatePair<Iter>> and you'd get a tuple looking like so: (Item, (Item, Item)).  If you wrote out code for each one you could fix the tuple to look like (Item, Item, Item).

bluss:
Look good! What do you think about changing the name? Something like Combinations, CombinePairs, PairCombinations ?

PR #30, opened 2015-05-31
Add a .unique() method to Iterator.

Wilfred:
I've found it very useful to be able to discard duplicates from iterators. I think this would be a useful addition to itertools, what do you think?


bluss:
Hi, that sounds useful. What do you think about this:

- It should be in the itertools trait
- Use a single struct, not two
- Add it in the form of `.unique_by` with the closure giving the key
- You're welcome to put it in a new module, but not a public submodule.

bluss:
This was superseded by PR #37

PR #31, opened 2015-06-04
Add `pad_tail_using` and `interleave_shortest` adaptors.

DanielKeep:
`pad_tail_using` ensures an iterator has some minimum number of elements, with a closure used to fill in the missing ones.

`interleave_shortest` does exactly what it says on the tin.

Both are adapted from `grabbag`.

### DIFF
--- a/src/adaptors.rs
+++ b/src/adaptors.rs
@@ -70,6 +70,84 @@ impl<I, J> Iterator for Interleave<I, J> where
     }
 }
 
+/// An iterator adaptor that alternates elements from the two iterators until
+/// one of them runs out.
+///
+/// This iterator is *fused*.
+///
+/// See [*.interleave_shortest()*](trait.Itertools.html#method.interleave_shortest)
+/// for more information.
+#[derive(Clone)]
+pub struct InterleaveShortest<I, J> where
+    I: Iterator,
+    J: Iterator<Item=I::Item>,
+{
+    it0: Fuse<I>,
+    it1: Fuse<J>,
+    phase: bool, // false ==> it0, true ==> it1
+}
+
+impl<I, J> InterleaveShortest<I, J> where
+    I: Iterator,
+    J: Iterator<Item=I::Item>,
+{
+    /// Create a new **InterleaveShortest** iterator.
+    pub fn new(a: I, b: J) -> InterleaveShortest<I, J> {
+        InterleaveShortest {
+            it0: a.fuse(),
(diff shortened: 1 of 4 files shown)


src/adaptors.rs
```diff
@@ -70,6 +70,84 @@ impl<I, J> Iterator for Interleave<I, J> where
...
+
+impl<I, J> Iterator for InterleaveShortest<I, J> where
+    I: Iterator,
+    J: Iterator<Item=I::Item>,
+{
+    type Item = I::Item;
+
+    #[inline]
+    fn next(&mut self) -> Option<I::Item> {
+        match self.phase {
```
bluss: I think we can shorten this with phase = !phase here.

bluss:
Looking forward to these prs from grabbag! Don't you think a name like pad_to or pad_to_length would be better?

DanielKeep:
@bluss From the commit message:

> I've chosen the `using` suffix as I believe possibly useful variants of
this would better suit the alternatives:
> 
> * `to`: best for a variant that just repeats the final element of the base sequence.
> * `from`/`with`: sounds like it should take another iterator as input; *i.e.* a kind of `chain`+`take` combination.

I hadn't considered `length`, but I think that belongs more with "repeats final element".

src/adaptors.rs
DanielKeep: You can, but I really didn't want to do that because then you need to match on `!phase`, which is weird.  If you want to cut the repetition out, I suppose I can bind the result of the match to a local `r`, then flip `phase` *after* the match, then return `r`... but the code would be the same length *anyway*.

I dunno; might just be my penchant for state machines.  :P

bluss:
That's a good point. I didn't look at the commit msg, away from my computer today.. What about simply pad_using?

DanielKeep:
Ok, I think that did it.  Not sure if you get a notification for that or not.

bluss:
Leaving a message here is the best yeah, no notification on just pushes.

bluss:
Looks fine. I'm a stubborn mind, so I wanted to ask. Is it ok if I merge this? I'm going to change the name to what I prefer, is that ok?

DanielKeep:
Yeah, if you want to change the names, go ahead; it's your prerogative, after all.

*Disclaimer*: I'm not going to be held responsible for naming conflicts with future hypothetical `pad_to_*` variants.  :P

PR #36, opened 2015-06-13
Add sort_by iterator adapter.

hjr3:
The existing `sort_by()` function in core does the sorting in place. This
breaks apart iterator chaining. This adapter removes the boilerplate of
collecting the existing iterator, sorting it and then turning it back
into an iterator.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1058,6 +1058,37 @@ pub trait Itertools : Iterator {
             None
         }
     }
+
+    /// Sort iterator elements.
+    ///
+    /// **Note:** This consumes the entire iterator, uses the
+    /// **slice::sort_by()** function and returns the sorted vector.
+    ///
+    /// ```
+    /// use itertools::Itertools;
+    ///
+    /// // sort people in descending order by age
+    /// let people = vec![("Jane", 20), ("John", 18), ("Jill", 30), ("Jack", 27)];
+    ///
+    /// let oldest_people_first = people
+    ///     .into_iter()
+    ///     .sort_by(|&a, &b| {
+    ///         a.1.cmp(&b.1).reverse()
+    ///     })
+    ///     .into_iter()
+    ///     .map(|(person, _age)| person)
+    ///     .collect::<Vec<_>>();
+    ///
+    /// assert_eq!(oldest_people_first, vec!["Jill", "Jack", "Jane", "John"]);
+    /// ```
+    fn sort_by<F: FnMut(&Self::Item, &Self::Item) -> Ordering>(self, cmp: F) -> Vec<Self::Item> where
+        Self: Sized
+    {
+        let mut v: Vec<Self::Item> = self.collect();
+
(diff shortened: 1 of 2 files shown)


bluss:
I think it's a bit wasteful in API terms to return a Vec::IntoIter when we have a vec. If we return a Vec then the user has more options. Your default use case will be more verbose -- `it.sort_by(f).into_iter().map(..)`  the `into_iter` would be needed to keep it in a method chain. I think it's for the best anyway.

We can't really call it an iterator adaptor since it eagerly consumes the whole iterator. If we return a vec we can put it in the section for the other methods.

hjr3:
Thanks for the feedback. I went back and forth on whether to return a vec or iterator. I updated the code to return a vec.

I am not sure about `sorted_by` mainly because we are now in the past tense and I have not seen that anywhere else. The [#min_by](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.min_by) function does not a return an iterator and is named in the present tense. I was following that convention.

src/lib.rs
```diff
@@ -1058,6 +1058,27 @@ pub trait Itertools : Iterator {
...
+    /// Sort iterator elements.
+    ///
+    /// **Note:** This consumes the entire iterator, uses the
+    /// **slice::sort_by()** function and returns the sorted vector.
+    ///
+    /// ```
+    /// use itertools::Itertools;
+    ///
+    /// // sort in descending order
+    /// let v = (0..5).into_iter().sort_by(|&a, &b| a.cmp(&b).reverse());
```
bluss: Thank you. This first into_iter is redundant, maybe the example can demonstrate using sort_by().into_iter() so that users don't miss that possibility?

src/lib.rs
```diff
@@ -1058,6 +1058,37 @@ pub trait Itertools : Iterator {
...
+    /// let oldest_people_first = people
+    ///     .into_iter()
+    ///     .sort_by(|&a, &b| {
+    ///         a.1.cmp(&b.1).reverse()
+    ///     })
+    ///     .into_iter()
+    ///     .map(|(person, _age)| person)
+    ///     .collect::<Vec<_>>();
+    ///
+    /// assert_eq!(oldest_people_first, vec!["Jill", "Jack", "Jane", "John"]);
```
hjr3: @bluss how about thie example? It does not show off `sort_by(Ord::cmp)` but it is the pattern I find myself going to a lot.

bluss: Great, I like verbose examples

bluss:
Do you want to squash together the commits? Add all fixups together or so, if you want?

hjr3:
@bluss squashed and the commit message updated. I really appreciate your time helping me with this PR!

bluss:
Travis builds have intermittent failures at the moment

bluss:
Thank you!

bluss:
published in 0.3.15.

PR #38, opened 2015-06-23
Add .group_by_lazy()

bluss:
`GroupByLazy` is the storage for the lazy grouping operation.

If the groups are consumed in order, or if each group's iterator is
dropped without keeping it around, then `GroupByLazy` uses no
allocations.  It needs allocations only if several group iterators
are alive at the same time.

This type implements `IntoIterator` (it is **not** an iterator
itself), because the group iterators need to borrow from this
value. It should stored in a local variable or temporary and
iterated.

Iterator element type is `Group` (an iterator).

cc @DanielKeep @eddyb, what do you think?
The iterator value is really fat, but in most common cases it will
not buffer anything and then it uses no allocations.

### DIFF
--- a/src/groupbylazy.rs
+++ b/src/groupbylazy.rs
@@ -49,6 +49,8 @@ impl<K, I, F> GroupInner<K, I, F>
                 }
             }
             elt
+        } else if self.done {
+            return None;
         } else if self.top == client {
             self.step_current()
         } else {
@@ -101,9 +103,7 @@ impl<K, I, F> GroupInner<K, I, F>
 
     /// This is the immediate case, where we use no buffering
     fn step_current(&mut self) -> Option<I::Item> {
-        if self.done {
-            return None;
-        }
+        debug_assert!(!self.done);
         if let elt @ Some(..) = self.current_elt.take() {
             return elt;
         }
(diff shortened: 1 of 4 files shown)


eddyb:
The `IntoIterator` approach is kinda neat, I like it :+1:.

bluss:
It's neat: I think it's the best we can do unless some kind of boxing / Rc is used around the common storage data. But it's very cryptic to those that don't know rust well, and it's not as easy to use in iterator chains.

By the way, we can solve `Chunks` in the same way.

DanielKeep:
Having only just woken up recently, I'm not going to try working out the details on this.  's makin' mah head hurt.

I tried an approach like this, but got stuck on the borrow checker; I mentioned this to a certain someone who assured me it couldn't be done like this *cough*eddyb*cough*.

So if this works, fantastic!  As you said, the only other solution I could come up with involved using `Rc` for the shared state.

I think the "it doesn't compose well" problem is significantly mitigated by the existence of the non-lazy `group_by`.  If someone wants a quick, composable group, ok: pay for the allocation.  If they want to avoid allocation, ok: pull the group out of the chain.

So really, I suppose it comes down to this: is saving one allocation worth the loss of chaining?  If yes, do this.  If no, I have an `Rc`-based `group_by_lazy` pretty much ready to go... that I probably should have pushed a while ago, but got distracted.  *cough*

eddyb:
@DanielKeep I was trying to do 0 allocations, using borrowing to ensure only one sub-iterator existed at any given time, which is the part not possible without "streaming iterators".

PR #39, opened 2015-06-24
Lazy formatter helper for iterators

bluss:


bluss:
I'm using PR's now so that I can have the travis builds run on stable, beta, nightly before I merge.

PR #43, opened 2015-07-06
Improve benchmarks & Add ZipSlices

bluss:


bluss:
@dotdash: Just a ping to say that almost every improvement in this PR is thanks to your comments on benchmarking but also the codegen of match on tuples

dotdash:
:blush: :tada:

Thanks for letting me know!

PR #51, opened 2015-08-20
Add reservoir sampling

shepmaster:
Howdy! This PR is certainly not production ready, but it works and I wanted to get some guidance about how to continue. Some points I have questions about:

1. How should this be tested? Testing random stuff is not my specialty, but I've always done basic statistical analysis. In this case, I did "*do each of the items occur about the same amount of times, with some fuzziness*".
2. How do you feel about including a production dependency? Should it have a Cargo feature flag including it?
3. Does the code fit the libraries coding style?

Obviously, documentation will need to be added.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -36,6 +36,9 @@
 //!
 //!
 
+extern crate rand;
+use rand::distributions::{IndependentSample, Range};
+
 use std::iter::{self, IntoIterator};
 use std::fmt::Write;
 use std::cmp::Ordering;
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -1260,6 +1263,23 @@ pub trait Itertools : Iterator {
...
+        where Self: Sized
+    {
+        let rng = &mut rand::thread_rng();
+
+        let mut reservoir: Vec<_> = self.by_ref().take(sample_size).collect();
+        for (i, item) in self.enumerate() {
+            let range = Range::new(0, i + sample_size + 1);
+            let replacement_index = range.ind_sample(rng);
+
+            if let Some(spot) = reservoir.get_mut(replacement_index) {
```
bluss: ah I love this, when the bounds check has a purpose for the algorithm.

bluss:
Hey, reservoir sampling is a cool algorithm, but I think the dependency and so on signals that this is not for itertools. It may be better in another kind of crate. We should try to keep focus on general iterator adaptors, I guess the boundary is fuzzy. We may already have too many special case things included.

As for testing, I'd do it just like you I think. Don't see any other nice way to do it immediately now.

shepmaster:
> another kind of crate

That makes sense. Do you know of any "iterator terminator" crates? In my mind sampling is very similat to functions like `fold` or `min`.

bluss:
oh I meant general iterator adaptors and methods; there are iterator methods / terminators in itertools.

shepmaster:
Closing in favor of rust-lang/rand#72.

Thanks!

PR #58, opened 2015-09-12
Added TakeWhilePutBack and TakeWhileMapPutBack.

Popog:
These are kind of an alternative to TakeWhileRef for iterators that don't support Clone.

### DIFF
--- a/src/adaptors.rs
+++ b/src/adaptors.rs
@@ -239,6 +239,7 @@ impl<I> PutBack<I> where
     {
         self.top = Some(x)
     }
+
 }
 
 impl<I> Iterator for PutBack<I> where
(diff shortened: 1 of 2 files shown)


bluss:
This is interesting. Maybe these should be methods on PutBack instead? They seem useful but very special case.

bluss:
Or even, an extension that (somehow) applies both to peekable and put back iterators.

Popog:
Yeah, I kind of agree, but I just couldn't find a design for those that felt clean to use.

bluss:
This PR is very well prepared.

While calling take_while_put_back on any iterator is nice, doesn't this come up mostly when you want to do this repeatedly? Imagine several uses, and the layers of `PutBack`s stacking up. I think that can be a reason it's better as an adaptor on `PutBack` itself.

bluss:
Thanks, I would love to see this in the adaptor on peekable / put back form. I might try to do it myself too.

PR #62, opened 2015-09-20
Add merge_join and hash_join

milancio42:
Hi,

This pull request contains the extension of the rust-itertools with SQL-like join adaptors. It covers all the standard SQL joins (inner + outer) as well as non-standard (but very useful) exclusive joins, which are well explained in docs.
Since they are all iterator adaptors, I thought  it would be better to add them into rust-itertools rather than creating a standalone library/project.

The adaptors are designed in a generic way, therefore they are very flexible.

There are two join strategies provided: **hash_join** and **merge_join**.

**hash_join**: allows join of non-sorted iterators, but requires the smaller one to be collectable into hash map (the other one can be arbitrarily long). This condition is usually satisfied, therefore this is the preferred strategy for its efficiency (no need to sort anything).

**merge_join**: allows join of sorted, arbitrarily large iterators. The user is responsible for providing a function comparing the left and the right iterator elements. As distinct from `merge_by` adaptor, the function must return `Ordering`, because we are interested also in the `Greater` variant.

All the objects are (hopefully) well documented and there are also tests for each adaptor.

The last point - this is my first pull request, therefore I apologize in advance if I missed something in the etiquette :)

Thank you for your consideration.
Regards,
milancio42

### DIFF
--- a/src/hash_join.rs
+++ b/src/hash_join.rs
@@ -25,27 +25,32 @@
 
 use std::collections::hash_map::{HashMap, IntoIter,};
 use std::mem;
-use std::rc::Rc;
 use std::hash::Hash;
 use super::EitherOrBoth::{self, Right, Left, Both};
 
-/// An iterator adaptor that [inner
-/// joins](https://en.wikipedia.org/wiki/Join_%28SQL%29#Inner_join) the two base iterators. The
-/// resulting iterator is the intersection of the two base iterators.
+/// An iterator adaptor that [inner joins](https://en.wikipedia.org/wiki/Join_%28SQL%29#Inner_join)
+/// the two base iterators in ascending order. The resulting iterator is the intersection of the
+/// two base iterators.
 ///
-/// The base iterators do *not* need to be sorted. The right base iterator is loaded into HashMap
-/// and thus must be unique on the join key (e.g. by
+/// The base iterators do *not* need to be sorted. The right base iterator is loaded into
+/// `HashMap` and thus must be unique on the join key (e.g. by
 /// [grouping](trait.Itertools.html#method.group_by), if necessary) to produce the correct
 /// results. The left base iterator do not need to be unique on the key.
 ///
 /// The left base iterator element type must be `(K, LV)`, where `K: Hash + Eq`. 
-/// The right base iterator element type must be `(K, RV)`, where `K: Hash + Eq`.
+/// The right base iterator element type must be `(K, RV)`, where `K: Hash + Eq` and `RV:
+/// Clone`.
 ///
-/// Iterator element type is `(LV, std::rc::Rc<RV>)`.
+/// When the join adaptor is created, the right iterator is **consumed** into `HashMap`.
+///
+/// Iterator element type is `(LV, RV)`. 
+/// The `RV` is cloned from `HashMap` for each joined value. It is expected a single `RV` will
+/// be joined (and cloned) multiple times to `LV`. To increase performance, consider wrapping
+/// `RV` into `std::rc::Rc` pointer to avoid unnecessary allocations.
 #[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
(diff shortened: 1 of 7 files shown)


bluss:
Hi, sorry if this is going to take some time — I'm not so familiar with sql. I'll do some high level comments first.

We will have to work towards removing explicit `Rc` in the interface. I think you can actually make this more general by requiring `Clone` from the value instead. The user mapping their values `T → Rc<T>` returns you back to the current code, as a special case of the new code.

In general, less is more :-) So maybe we don't need to expose all these variants?

Also, this work is significant enough that you can easily publish it as a separate crate if you want.

milancio42:
I think you're right. I had some doubts if it's not an overkill for rust-itertools, and now I see it probably is. To use joins properly and efficiently, one needs to have a good understanding of how they work, what trade-offs they have (simply one needs to know what (s)he is doing) and to expect such knowledge from clients of rust-itertools is probably not appropriate. My mistake.
Therefore, as you've said, perhaps it's better if I publish it as a standalone crate.

If you agree, we can close this PR.

PS: thanks for the suggestion regarding `Rc`, I'll apply it.

bluss:
I'm not saying it has to go into its own crate. But if you want to do that, go for it. I'll add some inline review notes now.

src/hash_join.rs
```diff
@@ -0,0 +1,474 @@
...
+impl<L, K, RV> HashJoinInner<L, K, RV> 
+    where K: Hash + Eq,
+{
+    /// Create a `HashJoinInner` iterator.
+    pub fn new<LI, RI>(left: LI, right: RI) -> Self
+        where L: Iterator<Item=LI::Item>,
+              LI: IntoIterator<IntoIter=L>,
+              RI: IntoIterator<Item=(K, RV)>
+    {
+        let mut map: HashMap<K, Rc<RV>> = HashMap::new();
```
bluss: This type doesn't need to be written out explicitly.

src/hash_join.rs
```diff
@@ -0,0 +1,474 @@
...
+/// The base iterators do *not* need to be sorted. The right base iterator is loaded into HashMap
+/// and thus must be unique on the join key (e.g. by
+/// [grouping](trait.Itertools.html#method.group_by), if necessary) to produce the correct
+/// results. The left base iterator do not need to be unique on the key.
+///
+/// The left base iterator element type must be `(K, LV)`, where `K: Hash + Eq`. 
+/// The right base iterator element type must be `(K, RV)`, where `K: Hash + Eq`.
+///
+/// Iterator element type is `(LV, std::rc::Rc<RV>)`.
+#[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
```
bluss: To be helpful, stress that the whole  `right` iterator is consumed when this JoinInner value is created.

src/lib.rs
```diff
@@ -550,6 +555,329 @@ pub trait Itertools : Iterator {
...
+    /// use itertools::Itertools;
+    ///
+    ///
+    /// let a = (0..).zip("AB".chars());
+    /// let b = (1..).zip("XY".chars());
+    /// let mut it = a.merge_join_inner_by(b, |x, y| Ord::cmp(&x.0, &y.0));
+    /// assert_eq!(it.next(), Some(((1, 'B'), (1, 'X'))));
+    /// assert_eq!(it.next(), None);
+    /// ```
+    fn merge_join_inner_by<J, F, R>(self, other: J, cmp: F) -> MergeJoinInner<Self, J::IntoIter, F>
```
bluss: type parameter `R` appears to be unneeded.

regexident:
FYI: This is equivalent to the [ordered_iter](https://crates.io/crates/ordered_iter) crate (~3000 downloads) by @contain-rs.

bluss:
Closing for inactivity. Again, this algorithm could just as well be published in its own crate.

Issue #54, opened 2015-08-30
'sorted()' method


zummenix:
It would be useful to have `sorted()` method for regular sorting that calls `sorted_by(Ordering::cmp)` internally. What do you think?

bluss:
Yes in the long run. Right now due to #49 it feels like we can't really use functions like this profitably, and supporting another variant is some more code. See how I'm trying to solve it in #53.

bluss:
A sorted() method is still a good idea, it was the unrelated breaking changes that were declined in #71.

PR #71, opened 2015-10-14
Change closure in the 'sorted_by' method to return bool.

zummenix:
Now signature of the closure similar to closure in the 'merge_by' method.
Also added 'sorted' method requested in #54 and removed deprecated 'sort_by' method.
Breaking change.


bluss:
I'm not so keen on breaking changes right now, we just had them bundled up in release 0.4.

zummenix:
Ok, I understand. Close this for now.

PR #72, opened 2015-10-21
Write iproduct!() recursively

bluss:


FranklinChen:
Cool, one step toward https://github.com/bluss/rust-itertools/issues/10#issuecomment-102572242 which does require more machinery.

bluss:
@FranklinChen Ah it's not really a step, this just reorganizes the code that's already present. It's mostly a technicality, in this new shape, iproduct permits using temporary values in its arguments better.

Issue #64, opened 2015-10-01
Add a .windows(n) iterator adapter


flo-l:
Similar to [```.windows()```](https://doc.rust-lang.org/std/primitive.slice.html#method.windows) in std, just not only for slices but for arbitrary iterators. I saw that ```.chunks()``` is already implemented as ```chunks_lazy()```. Would you find eg. ```windows_lazy()``` implemented in a similar fashion useful?

I've read the code of ```chunks_lazy()```, it's quite complicated. But with a few hints from somebody who knows the code I could probably implement ```windows_lazy()``` myself.

bluss:
I haven't thought about this too deeply, but it would require a lot of copying of iterator elements? Since each element appears in several windows. It might not be bad at all, when applied to usual by-reference iterators like the slice iterator.

bluss:
I'll be assigned just to keep the issue on my radar. (Feel free to implement this if you want). Not sure how easy it's to translate GroupByLazy's model to windows..

flo-l:
ok I'll give it a try as soon as I find time!

bluss:
A link to https://crates.io/crates/iterslide came up when browsing /r/rust

flo-l:
sorry for the long response time, university is tough sometimes!

This crate looks like the exact thing I want in itertools! Should the author be asked for their permission to include the code or should I reimplement it?

bluss:
It's very wasteful by cloning so much, not sure if I like its approach.

flo-l:
ok you're right it clones the whole vec on each iteration :disappointed:

I'll give it a try and code something in the weekend and open a pr here! (hopefully without being as wasteful)

flo-l:
I gave it some thought and found a problem.

I wanted to implement it like so: Have the adapter struct store the current window in a vec and let next() return a slice of that vec.

This is impossible because the iterator protocol allows this:

```rust
let it = other_it.windows_adapter(3);
let slice_a = it.next();
let slice_b = it.next();
```

Clearly the second call to next() invalidates slice_a.

I think it could be implemented utilizing a RefCell or a custom type providing similar functionality. The user then had to make sure slice_a isn't used after the second call to next().

Any thoughts?

bluss:
A solution parametrized by arrays would be awesome, but I guess it would dictate compile time window sizes, which does not fit all use cases.

Then how to solve the borrowing problem.. it's certainly expected, and iterslide uses cloning to not run into it.

[...]

PR #73, opened 2015-11-09
[WIP] implement sliding_windows()

flo-l:
This is the code for #64.

@bluss How do you feel about the design?

I'll write docs as soon as the design is approved.

The code as written now disallows the following by runtime checks:

```rust
let mut storage: SlidingWindowStorage<u32> = SlidingWindowStorage::new(3)
let iter = (0..100).sliding_windows(storage);

// a must go out of scope before the second call to iter.next()
let a = iter.next();
let b = iter.next();

// storage may not be dropped before the last Window, in this case a
let mut storage: SlidingWindowStorage<u32> = SlidingWindowStorage::new(3);
let a = {
    let iter = (0..100).sliding_windows(storage);
    iter.next()
};
std::mem::drop(storage);
// a points to garbage now
```

In either case the code ```panic!()```s with an error message.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 
 name = "itertools"
-version = "0.4.2"
+version = "0.4.3"
 
 license = "MIT/Apache-2.0"
 repository = "https://github.com/bluss/rust-itertools"
--- a/README.rst
+++ b/README.rst
@@ -33,6 +33,11 @@ How to use in your crate:
 Recent Changes
 --------------
 
+- 0.4.3
+
+  - Write iproduct!() as a single expression; this allows temporary values
+    in its arguments.
+
 - 0.4.2
 
   - Add .fold_options()


flo-l:
I think I hit some [breaking-change] related behaviour, but it seems to compile on stable at least :p

bluss:
Thanks. I'll poke a bit at the code. I hope it can be done a lot safer with using Cell and references.

bluss:
Ok I sent some edits to your branch as a PR.

The clean way to implement custom interior mutability is via UnsafeCell; also splitting storage and iterator enables the much safer pattern of borrowing the storage in each iterator element. This is all quite neat.

However, I remain sceptical of an iterator that panics if multiple elements live at the same thing; it will break when put into some iterator adaptors. So it's certainly interesting and cool. All solutions have downsides.

bluss:
Maybe it would be maximally conservative, but we could consider using a Result type, returning Error if there is a borrowing error. It's a bit ridiculous, but it puts the error handling over in the user's court :/

flo-l:
I like the Result idea. I found it scary how invisible the whole unsafe borrowing thing was after I had cleaned up the api a bit!

I'll go ahead and merge your PR (thanks btw) and continue my work atop your changes. The Cell types make it a lot less scary ;)

bluss:
Yes it's a lot better :) Cell is always completely safe to use, while UnsafeCell must be used with care.

bluss:
Note that something that is mutated behind a shared reference (&-pointer) *must* be inside an `UnsafeCell`

flo-l:
Is it desired that the adapter throws away the last value of the underlying iterator and returns ```None``` forever as soon as it encounters a borrow error? (Actually it returns ```Some(Err(()))``` first and then ```None```)

flo-l:
Wouldn't it be cool to give the user a chance to recover from a borrow error?

bluss:
Right, that doesn't sound good. Better to check up front. I don't think recovering from borrow error is something that will happen in actual code; encountering it is evidence for a bug. With that reasoning, it really is borderline on whether it should panic or not.

flo-l:
I can't think of a realistic scenario where a borrow error is expected. Maybe there is some crazy code possible which does all the borrowing dynamically etc. etc. but that's a niche use case. Especially as window is neither send nor sync, so scary thread constructions are impossible. In that case a cloning implementation of windows wouldn't be a huge performance loss anyway I guess.

But I like how result makes it explicit that such an error can happen. It's quite likely that some other adapter messes things up.

However I'm leaning towards the panic solution because it makes user code more compact and result doesn't give the user a real alternative to panicking. Also if the user messes up a panic probably occurs early in the development process anyway as it's unusual to call next() just in rare cases twice.

flo-l:
I'll reopen it soonish, I just don't want to develop on master

bluss:
great

PR #74, opened 2015-11-12
Sliding windows Take 2 [WIP]

flo-l:
reopened version of #73, but this time without ```Option<Result>``` as return type, for rationale see my comment in #73.

### DIFF
--- a/src/sliding_windows.rs
+++ b/src/sliding_windows.rs
@@ -39,6 +39,7 @@ impl<T> SlidingWindowStorage<T> {
         Window { drop_flag: &self.uniquely_owned, data: &self.data }
     }
 
+    // push value onto self
     fn push(&self, elt: T) -> bool {
         assert!(self.uniquely_owned.get(), "next() called before previous Window went out of scope");
         let data = unsafe { &mut *self.data.get() };
@@ -49,6 +50,13 @@ impl<T> SlidingWindowStorage<T> {
             data.push(elt);
         }
         data.len() == self.window_size
+    }
+
+    // clear backing storage
+    fn clear(&self) {
+        assert!(self.uniquely_owned.get(), "next() called before previous Window went out of scope");
+        let data = unsafe { &mut *self.data.get() };
+        data.clear();
     }
 }
 
@@ -140,6 +148,9 @@ impl<'a, I: Iterator> SlidingWindowAdapter<'a, I> {
     ///
     /// See [`.sliding_windows()`](trait.Itertools.html#method.sliding_windows) for more information on the adapter.
     pub fn new(iter: I, storage: &'a SlidingWindowStorage<I::Item>) -> SlidingWindowAdapter<'a, I> {
+        // in case the storage was reused
+        storage.clear();
+
         SlidingWindowAdapter {
             iter: iter,
             done: false,
--- a/src/sliding_windows.rs
+++ b/src/sliding_windows.rs
@@ -1,7 +1,5 @@
 use std::cell::{Cell, UnsafeCell};
-use std::mem;
 use std::ops::{Deref, DerefMut};
-
 use std::fmt;
 
 // This stores the current window
--- a/src/sliding_windows.rs
+++ b/src/sliding_windows.rs
@@ -2,7 +2,7 @@ use std::cell::{Cell, UnsafeCell};
 use std::ops::{Deref, DerefMut};
 use std::fmt;
 
-/// This holds the backing allocation for the `Window` of a `SlidingWindowAdapter`.
+/// This holds the backing allocation for the `Window` of a `SlidingWindowAdaptor`.
 /// 
 /// See [`.sliding_windows()`](trait.Itertools.html#method.sliding_windows) for more information.
 pub struct SlidingWindowStorage<T> {
@@ -60,7 +60,7 @@ impl<T> SlidingWindowStorage<T> {
     }
 }
 
-/// This is the `Item` type of the `SlidingWindowAdapter` iterator.
+/// This is the `Item` type of the `SlidingWindowAdaptor` iterator.
 ///
 /// # Usage:
 ///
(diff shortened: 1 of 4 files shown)


flo-l:
@bluss ping

this is ready for review

bluss:
I'm mulling over the idea

src/sliding_windows.rs
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    // push value onto self
+    fn push(&self, elt: T) -> bool {
+        assert!(self.uniquely_owned.get(), "next() called before previous Window went out of scope");
+        let data = unsafe { &mut *self.data.get() };
+        if data.len() != self.window_size {
+            data.push(elt);
+        } else {
+            data.remove(0);
```
mitchmindtree: It seems like the `data` field might be more efficiently implemented as a [`VecDeque`](http://doc.rust-lang.org/std/collections/struct.VecDeque.html)? `VecDeque::pop_front` should be more efficient than `Vec::remove(0)` for this case.

bluss: Asymptotically yes. Vec will be better when windows are small. I guess we'll have to take a deque to not be catastrophic on larger sizes yeah. It can be fixed later.

flo-l: If we do that Window can't deref to a slice anymore, so we'd either have to implement Iterator for Window (a proxy to VecDeque's iterator) or let Window deref to VecDeque directly (which would be a bit strange but also quite practical). Both solutions make mutating Items of Window less straightforward.

But you're right, for big Windows a Vec will be painfully slow.

bluss: Oh good point, I did already look at it that way. I don't think it's an issue.

flo-l:
pushed a commit to show how I imagine reusing an allocation could work!

src/sliding_windows.rs
```diff
@@ -0,0 +1,201 @@
...
+    fn clear(&self) {
+        assert!(self.uniquely_owned.get(), "next() called before previous Window went out of scope");
+        let data = unsafe { &mut *self.data.get() };
+        data.clear();
+    }
+}
+
+impl<T> Into<Vec<T>> for SlidingWindowStorage<T> {
+    fn into(self) -> Vec<T> {
+        assert!(self.uniquely_owned.get(), "Storage dereferenced before previous Window went out of scope");
```
flo-l: I guess this could be a ```debug_assert!``` as the rust type system forbids this.

bluss:
It's a hard decision, I'm sorry for responding so late.

We are not doing so well in trying to force this nice algorithm to be an iterator. This is really a [“streaming iterator”](http://burntsushi.net/rustdoc/fst/trait.Streamer.html) kind of API, the one-at-a-time requirement is common to that whole class of traversals.

In itertools we already have `GroupByLazy`, which favours the one at a time case, but it degrades gracefully in the general case. We also have `Tee`, that can have borrow errors, but only if you tie some kind of iterator knot where it calls back into itself in .next(). I can't even remember how to do it right now.

I think that using multiple elements at a time is too common for us to want to introduce an iterator with this gotcha — an example of such use is the `dedup()` adaptor. Putting the unwrapping over in the user's court would help, but it doesn't really make a real iterator out of this unfortunately.

I think this algorithm would do fine on its own, in its own crate; it could provide both “streamer” and iterator interfaces.

src/sliding_windows.rs
ehiggs: Maybe add a benchmark in the benchmark dir. Something like the following (which doesn't compile):

```rust
#[bench]                                                                        
fn slice_windows(b: &mut test::Bencher) {                                       
    let data = vec![0; 1024];                                                   
                                                                                
    let data = test::black_box(data);                                           
    let sz = test::black_box(10);                                               
                                                                                
    b.iter(|| {                                                                 
        for group in data.windows(sz) {                                         
            for elt in group {                                                  
                test::black_box(elt);                                           
            }                                                                   
        }                                                                       
    })                                                                          
}                                                                               
#[bench]                                                                        
fn sliding_windows(b: &mut test::Bencher) {                                     
    let data = vec![0; 1024];                                                   
                                                                                
    let data = test::black_box(data);                                           
    let sz = test::black_box(10);                                               
                                                                                
    let mut storage: SlidingWindowStorage<u32> = SlidingWindowStorage::new(10);
    b.iter(|| {                                                                 
        for win in &data.iter().sliding_windows(&mut storage) {             
            for elt in win.iter() {                                             
                test::black_box(elt);                                           
            }                                                                   
        }                                                                       
    })                                                                          
}
```

ehiggs:
I didn't see a test for the `sliding_window` function that is actually exposed by the module.

flo-l:
Hmm I understand your concerns. I'll transfer the code into it's own crate. If you change your mind please let me know, I would like to see this in itertools as I've missed the functionality several times.

ehiggs:
>We are not doing so well in trying to force this nice algorithm to be an iterator. This is really a “streaming iterator” kind of API, the one-at-a-time requirement is common to that whole class of traversals.

This is kinda weird since `iter-tools` has chunking which mirrors the slice method. But now windows which is also a slice method. For users looking to use slice functionality on iterators without having to call `collect` beforehand, it's weird to traipse around cargo.io to find all the crates.

Would it be useful to have windowing functions in a submodule of iter-tools. Or maybe pull @burntsushi's StreamingIterator in as a submodule.

bluss:
I'm not sure that itertools should start collecting streaming iterators too. They are a developing concept, best do that on their own.

Including a `windows()` in itertools is totally dependent on being able to find a way to make it a nice and simple to use iterator. I don't think we've found that nice solution yet. The mechanics of chunks and windows are clearly different, as you can see the libstd can provide `chunks_mut` but no `windows_mut` for example — `windows_mut` is only possible as a streaming iterator.

flo-l:
for reference,  this code lives now here:

https://github.com/flo-l/rust-sliding_windows/

ehiggs:
>Including a windows() in itertools is totally dependent on being able to find a way to make it a nice and simple to use iterator. I don't think we've found that nice solution yet.

Agree, but this can be mitigated w/ `#[unstable...]`.

On the matter of iterators vs. streaming iterators and single item iterators vs. multi-item iterators, how could you go about organizing them if you were the curator of all iterator type types. e.g.

* Rust [std::Iter](http://doc.rust-lang.org/src/core/iter.rs.html#11-2977)
  * ...
* itertools - single item iterators
* [sliding_windows](https://github.com/flo-l/rust-sliding_windows/) - multi item iterators
* [fst::Streamer](https://github.com/BurntSushi/fst/blob/master/src/stream.rs) - @burntsushi's streaming iterator
* [timely-dataflow::dataflow::Stream](https://github.com/frankmcsherry/timely-dataflow/blob/master/src/dataflow/stream.rs) - @frankmcsherry's dataflow::Stream is not strictly an Iterator type (I don't see a `next` fn), but has a lot of the same intent.
* Infinite stream of random samples Iterator
* Infinite stream of random window samples Iterator

bluss:
I would like to be involved in generic streaming iterator adaptors or combinators, but from most of my experiments, there unfortunately isn't much that is easily expressible in Rust.

I'm certainly not the curator of all iterators :smile:. All kinds of interesting data structure work  involves iterators and streaming iterators, and I've published specific kinds of iterators in [odds](https://crates.io/crates/odds), [permutohedron](https://crates.io/crates/permutohedron), [petgraph](https://crates.io/crates/petgraph), and other places.

PR #76, opened 2015-11-15
Expand the MultiPeek API

arcnmx:


bluss:
My idea for expansion has been that it could have a “peeking” iterator, which would subsume this.

bluss:
Superseded by issue #93

PR #78, opened 2015-11-24
Add `run_length` adaptor

1uks:
Hi,

this commit implements an iterator adaptor that yields the run length of an element and its value


bluss:
Hi, thanks for the PR!

I'm unsure if this should be included, it's relatively straightforward to do this with .coalesce as it is (although not so short to write).

bluss:
Thank you, I think this is not general enough that we want to include it at this point.

Issue #79, opened 2015-12-14
Add minmax/minmax_by_key


birkenfeld:
Sometimes I'd like to keep track of minimum and maximum without iterating twice (if that's even possible). Itertools might be a good place for such adapters?

bluss:
yes, that sounds good! We just need to pick a name that doesn't conflict with std. I guess they don't have minmax right now?

birkenfeld:
I haven't seen anything that is called `minmax` there. I don't know if there is a "mathematically nicer" name though :)

I'll have a go at a pull request.

bluss:
These are in now.

PR #80, opened 2015-12-14
Add minmax() and minmax_by_key() adapters (#79)

birkenfeld:
The Clone bound on Self::Item is pretty obvious in hindsight, but
I didn't think of it when starting the implementation.  It is also
the reason for the slightly complicated fold() closure.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1236,6 +1236,58 @@ pub trait Itertools : Iterator {
     {
         self.sorted_by(cmp)
     }
+
+    /// Return the minimum and maximum element of an iterator.
+    ///
+    /// For the minimum, the first minimal element is returned.  For the maximum,
+    /// the last maximal element wins.  This matches the behavior of the standard
+    /// `Iterator::min()` and `Iterator::max()` methods.
+    ///
+    /// This requires `clone()`-able elements since the returned minimum and
+    /// maximum might actually be the same item.
+    fn minmax(mut self) -> Option<(Self::Item, Self::Item)>
+        where Self: Sized, Self::Item: Ord, Self::Item: Clone
+    {
+        self.next().map(|first| {
+            self.fold((first.clone(), first), |minmax, el| {
+                match (el < minmax.0, el >= minmax.1) {
+                    (true, false) => (el, minmax.1),
+                    (false, true) => (minmax.0, el),
+                    (true, true)  => (el.clone(), el),
+                    _ => minmax,
+                }
+            })
+        })
+    }
+
+    /// Return the minimum and maximum element of an iterator, as determined by
+    /// the specified function.
+    ///
+    /// For the minimum, the first minimal element is returned.  For the maximum,
+    /// the last maximal element wins.  This matches the behavior of the standard
+    /// `Iterator::min()` and `Iterator::max()` methods.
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -1236,6 +1236,58 @@ pub trait Itertools : Iterator {
...
     }
+
+    /// Return the minimum and maximum element of an iterator.
+    ///
+    /// For the minimum, the first minimal element is returned.  For the maximum,
+    /// the last maximal element wins.  This matches the behavior of the standard
+    /// `Iterator::min()` and `Iterator::max()` methods.
+    ///
+    /// This requires `clone()`-able elements since the returned minimum and
+    /// maximum might actually be the same item.
```
bluss: The old min max in libstd handled this by using a None / One / Two kind of enum. It's a tradeoff, but I think the best is to support non-Clone elements.

A good ergonomic solution may be a method on the result enum that converts the result to a pair of values, defined only if the result is clonable?

Old code: https://github.com/rust-lang/rust/blob/d07d465cf60033e35eba16b9e431471d54c712f4/src/libcore/iter.rs#L809-L891

src/lib.rs
```diff
@@ -1236,6 +1236,58 @@ pub trait Itertools : Iterator {
...
+    fn minmax(mut self) -> Option<(Self::Item, Self::Item)>
+        where Self: Sized, Self::Item: Ord, Self::Item: Clone
```
bluss: I think using just `PartialOrd` is fine; user should decide which kind of guarantees about ordering they want. Alternatively, maybe there's a generalization of  let min = min(a, b) that works for both Ord and for floats? That could be interesting to explore.

src/lib.rs
birkenfeld: Oh wow, didn't even know this existed in std. Well then, let's copy it shamelessly :)

src/lib.rs
birkenfeld: So far I followed std min and max, which require `Ord`. This might be a more useful addition with `PartialOrd`, but I'm not sure what the behavior for NaN should be - three not too unreasonable ones come to mind:

a) skip all NaNs, return None for only-NaN iterator
b) skip all NaNs, but return NaN for only-NaN iterator
c) if there is a NaN, return it as min and max

I also guess that if we want this, we have to create a custom trait `OrdOrFloat`?

bluss: `f32::min/max` return the non-NaN number if there is one.

I'm not saying this is possible, but I have the ambition :smile:

bluss: I think std is too conservative, and that this is a typical case where floats will want to be used as well. Ord vs PartialOrd is the user's choice: they choose what type to input. Unfortunately there may not be a nice way to put all Ord types and floats under the same trait, since trait coherence will maintain that f32 may impl Ord in the future, ruling it conflicting.

If it's not possible, then we'll skip the float concerns for now.

bluss:
Closing since it's inactive.

birkenfeld:
I can update it with the `Ord` bound. Not sure how to progress with floats and `PartialOrd`.

bluss:
maybe itertools can't solve the float problems. float-itertools are needed :D

PR #83, opened 2015-12-24
Add Flatten adaptor

panicbit:


panicbit:
The build fails because Rust  <= 1.1.0 had no `Clone` impl for `option::IntoIter`.

bluss:
Everything looks good! Will be a new release when I'm home.

bluss:
Glorious automation

Issue #59, opened 2015-09-18
Add iterate() function


gyscos:
To replace the deprecated one from the std:
https://doc.rust-lang.org/std/iter/fn.iterate.html

bluss:
PR welcome.

We want to reorganize all "source" / "generator" iterators (iterators without other iterators as input). The should all be created by functions, Right now we have `RepeatCall` but also `linspace`.

PR #85, opened 2015-12-29
Add itertools::func.

Stebalien:
This is effectively a shortcut for `iter::repeat(()).map(|_| something)`
but less ugly.

### DIFF
--- a/src/fn_iter.rs
+++ b/src/fn_iter.rs
@@ -4,6 +4,27 @@ pub struct FnIter<F>(F);
 /// Iterate over a function.
 ///
 /// The returned Iterator calls the passed function on each call to next.
+///
+/// Example:
+///
+/// ```
+/// // Generate the fibonacci sequence (no recursion).
+/// let fib = itertools::func({
+///     let mut a = 0;
+///     let mut b = 1;
+///     move || {
+///         let ret = a;
+///         let next = a + b;
+///         a = b;
+///         b = next;
+///         Some(ret)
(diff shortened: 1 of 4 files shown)


Stebalien:
Fixes  #59 (but is much simpler than the original stdlib version).

bluss:
It's similar to the existing `RepeatCall`. RepeatCall probably needs a free function constructor. `func` also needs a better name.

src/fn_iter.rs
```diff
@@ -0,0 +1,41 @@
+/// An iterator that yields values returned by a function.
+pub struct FnIter<F>(F);
```
bluss: For future reference, iterator structs should be named after the method that creates them as far as possible. So with the current name this would be `Func`.

Stebalien:
Sorry, I didn't notice `RepeatCall` and `Unfold` due to the lack of free functions. Unfortunately, `RepeatCall` doesn't allow one to end the `Iterator` and `Unfold` is unnecessarily complicated given that rust has mutable closures (there's a reason this was dropped from std). However, I'm happy enough with `RepeatCall` (one can always `repeat_call(...).filter_map(|x| x)`).

bluss:
or `repeat_call().while_some()` to allow terminating it.

bluss:
I've realized that itertools needs some reforming, iterators should have free function constructors (as long as the iteration is the important part, and not the struct itself).

PR #88, opened 2016-01-03
Added `diff.rs` module with `diff` and `copy_on_diff` - for efficiently "diff"ing and caching non-`Clone` iterators.

mitchmindtree:
I originally wrote this module for a couple of cases in conrod where I wanted to cache a non-`Clone` iterator without having to collect the iterator into a `Vec` for comparison in case the cache needed updating.

I've just come across another similar case in a personal project, so before copying code across I thought I'd approach the itertools team in case you deem it worthy of adding to your luscious collection.

Either way, any feedback appreciated!

### DIFF
--- a/src/adaptors.rs
+++ b/src/adaptors.rs
@@ -176,6 +176,13 @@ impl<I> PutBack<I> where
         PutBack{top: None, iter: it}
     }
 
+    /// Create a `PutBack` along with the `value` to put back.
+    #[inline]
+    pub fn value(value: I::Item, it: I) -> Self
+    {
+        PutBack{top: Some(value), iter: it}
+    }
+
     /// Put back a single value to the front of the iterator.
     ///
     /// If a value is already in the put back slot, it is overwritten.
--- a/src/adaptors.rs
+++ b/src/adaptors.rs
@@ -178,9 +178,17 @@ impl<I> PutBack<I> where
 
     /// Create a `PutBack` along with the `value` to put back.
     #[inline]
-    pub fn value(value: I::Item, it: I) -> Self
+    pub fn with_value(value: I::Item, it: I) -> Self
     {
         PutBack{top: Some(value), iter: it}
+    }
+
+    /// Split the `PutBack` into its parts.
+    #[inline]
+    pub fn into_parts(self) -> (Option<I::Item>, I)
+    {
+        let PutBack{top, iter} = self;
+        (top, iter)
     }
 
     /// Put back a single value to the front of the iterator.
--- a/src/diff.rs
+++ b/src/diff.rs
@@ -8,8 +8,9 @@
 //! iterators `a` and `b`, borrowing the source of `a` if they are the same or creating a new owned
 //! collection with `b`'s elements if they are different.
 
+use adaptors::PutBack;
 use std::borrow::{Cow, ToOwned};
-use std::iter;
+use std::iter::FromIterator;
 
 /// A type returned by the [`diff`](./fn.diff.html) function.
 ///
@@ -22,9 +23,9 @@ pub enum Diff<I>
 {
     /// The index of the first non-matching element along with the iterator's remaining elements
     /// starting with the first mis-matched element.
-    FirstMismatch(usize, iter::Chain<iter::Once<I::Item>, I>),
+    FirstMismatch(usize, PutBack<I>),
     /// The remaining elements of the iterator.
-    Longer(iter::Chain<iter::Once<I::Item>, I>),
+    Longer(PutBack<I>),
     /// The total number of elements that were in the iterator.
     Shorter(usize),
 }
@@ -59,11 +60,11 @@ pub fn diff<'a, I, J>(i: I, j: J) -> Option<Diff<J::IntoIter>>
         match j.next() {
             None => return Some(Diff::Shorter(idx)),
             Some(j_elem) => if *i_elem != j_elem {
-                return Some(Diff::FirstMismatch(idx, iter::once(j_elem).chain(j)));
+                return Some(Diff::FirstMismatch(idx, PutBack::value(j_elem, j)));
             },
         }
     }
-    j.next().map(|elem| Diff::Longer(iter::once(elem).chain(j)))
+    j.next().map(|elem| Diff::Longer(PutBack::value(elem, j)))
 }
 
 /// Returns `Cow::Borrowed` `collection` if `collection` contains the same elements as yielded by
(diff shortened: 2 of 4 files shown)


mitchmindtree:
Travis seems to build successfully on all versions of rustc other than stable `1.1.0`.

bluss:
Looks interesting. I wish we had an itertools team!

- Don't you want both a and b iterators for the first mismatch case, and the a iterator in the shorter case?
- `IterDiff` can have just one type param since the `E` depends on the iterator type. Would be two params if it's changed to have two kinds of iterators, of course
- Equality could be a variant of the enum
- We can use PutBack instead of chain with once, or alternatively, anything that works in Rust 1.1 inside a wrapper newtype that hides the type.

Style stuff

- I like to use `I` and `J` for type parameters that are iterators (or iterables). It would be consistent to do so. `A` and `B` are good for element types.
- We can drop the iter prefix on the function and type. Everything in itertools is iter- something :smile:.

mitchmindtree:
> Don't you want both a and b iterators for the first mismatch case, and the a iterator in the shorter case?

Hmmm i'm unsure what you mean here - the idea of `iter_diff` is that `a` is the currently cached state and `b` is the new state which we want to diff and adapt too. For `FirstMismatch` we only need the first index at-which `b` is different, along with all of its elements from that index onwards. For the `Shorter` case, we only need the index at which `b` finished, as this is enough for us to `truncate` or `take` on `a`'s collection after `iter_diff` is called. There are examples of these in the `copy_on_diff` function and tests.

> Equality could be a variant of the enum

Indeed, though in practise i've found it to be a little more ergonomic using `Option`, as normally I want to know if there is `Some` difference or `None` before deciding to match, i.e.

```rust
if let Some(diff) = iter_diff(&state.view().elems, new_elems) {
    state.update(|inner| match diff {
        Diff::Shorter(len) => inner.elems.truncate(len),
        Diff::FirstMismatch( , ) => ...
        Diff::Longer ...
    })
}
```

Cheers for the suggestions! They all sound good, I'll do them and commit 'em tomorrow :+1:

mitchmindtree:
@bluss I've made the recommended changes - looking much nicer, thanks.

I also added a function to `PutBack` that allows constructing it with the value to put back so that it's a little more expressive. Let me know if this is ok or if you'd prefer I left it as is :+1:

bluss:
Nice. I think I understand your usecase, but wouldn't it make sense to expand it to do a symmetrical diff? So that both sides are treated the same way. It should be simple to still use.

bluss:
Something like 

```rust
FirstMismatch(usize, PutBack<I>, PutBack<J>),
Longer(usize, PutBack<I>),
Shorter(usize, PutBack<J>),
```

mitchmindtree:
Ahh yes I see what you mean! I'm not sure if this will work or be useful in the case of `iter_diff` though, as `I` yields references to its elements (for checking PartialEq without moving the values out of the collection) while `J`s elements are by value. Maybe another type like `SymmetricDiff` could be made for this?

bluss:
Oh, I didn't notice that detail.. if T == &T was defined everywhere, it'd be simple.

bluss:
I'm sorry for making it difficult for you. I'm mainly interested in the symmetric diff.

mitchmindtree:
Hey sorry @bluss, been busy with other stuff over the past couple weeks but will get around to this asap :+1:

mitchmindtree:
Okydoke!

I've rebased and changed the `Diff` and `diff` to be symmetric as you described above. I also added a `diff_by_ref` function which allows for the original `diff` behaviour and is a little friendlier to iterator caching.

src/adaptors.rs
```diff
@@ -176,6 +176,13 @@ impl<I> PutBack<I> where
         PutBack{top: None, iter: it}
     }
 
+    /// Create a `PutBack` along with the `value` to put back.
+    #[inline]
+    pub fn value(value: I::Item, it: I) -> Self
```
bluss: Can we change this to for example `with_value`? It's more clear it's a constructor then.

bluss:
I like the new diff! At this point we want some kind of method to split PutBack into its parts too I think? `-> (Option<I::Item>, I)`

mitchmindtree:
Both good suggestions, Done!

bluss:
Great. What pains me is how much more efficient it is to write copy_on_diff specifically for slices and vecs rather than iterators. I haven't profiled it, but it's my general understanding. Slice version below. It could be even a bit better with more detailed preallocation of the vector.

```rust
pub fn copy_on_diff<'a, U, T: 'a>(collection: &'a [T], update: U) -> Cow<'a, [T]>
    where U: IntoIterator<Item=T>,
          T: Clone + PartialEq,
{
    let c_iter = collection.into_iter();
    match diff_by_ref(collection.into_iter(), update.into_iter()) {
        Some(diff) => match diff {
            Diff::FirstMismatch(idx, _, mismatch) => {
                let mut v = collection[..idx].to_vec();
                let (top, rest) = mismatch.into_parts();
                if let Some(elt) = top {
                    v.push(elt);
                }
                v.extend(rest);
                Cow::Owned(v)
            }
            Diff::Longer(_, remaining) => {
                let mut v = collection.to_vec();
                let (top, rest) = remaining.into_parts();
                if let Some(elt) = top {
                    v.push(elt);
                }
                v.extend(rest);
                Cow::Owned(v)
            }
            Diff::Shorter(num_update, _) => {
                Cow::Owned(collection[..num_update].to_vec())
            }
        },
        None => Cow::Borrowed(collection),
    }
}
```

mitchmindtree:
@bluss totally, this crossed my mind also. Perhaps the original is being a bit too generic, and is sacrificing too much performance-wise as a result. I'd be fine going with your slice-specific behaviour, perhaps changing `copy_on_diff` to something like `copy_slice_on_diff`?

Another option might be to abstract this functionality into a trait, which can be optimally implemented for each collection. We could start by just providing an impl for `Vec`, and adding others if they are requested. This way, `copy_on_diff` could instead take some collection that implements this trait, retaining the original genericness with greater flexibility?

mitchmindtree:
I'll write up a draft for the trait in a minute

bluss:
Hm it's really borderline if it should be in itertools

mitchmindtree:
Yeah I can see where you're coming from. Might as well drop the `copy_on_diff` stuff for now due to the uncertainty related to its place in itertools as well as the possibility of going with a more specialisable approach in the future.

Alrighty, just removed it!

PR #89, opened 2016-01-05
Add 'sorted()' method.

### LINKED ISSUE #54: 'sorted()' method

zummenix:
Closes #54


bluss:
Thanks!

PR #90, opened 2016-01-06
Add combinations_n

salty-horse:
itertools already provides ```combinations``` which returns all choices of 2 items. It Seemed odd that it doesn't provide a more general form.

```combinations_n``` returns all possible choices of n items from a given iterator.

Perhaps this implementation isn't a good fit for the library. Up to you to decide.

A few issues:

1. I've implemented lazy buffering. combinations_n only consumes data from the iterator when it has to. However, I don't know if the implementation is good enough, or if I placed it in the right file.
2. I couldn't manage to set the Clone trait correctly on the new structs (due to being new to Rust), so I removed it.

### DIFF
--- a/src/adaptors.rs
+++ b/src/adaptors.rs
@@ -1126,14 +1126,16 @@ struct LazyBuffer<I: Iterator> {
     buffer: Vec<I::Item>,
 }
 
-impl<I> LazyBuffer<I> where I: Iterator + Clone {
+impl<I> LazyBuffer<I> where I: Iterator {
     pub fn new(it: I) -> LazyBuffer<I> {
-        let mut it = it.clone();
-        let first = it.next();
-        let done = first.is_none();
+        let mut it = it;
         let mut buffer = Vec::new();
-        if !done {
-            buffer.push(first.unwrap());
+        let done;
+        if let Some(first) = it.next() {
+            buffer.push(first);
+            done = false;
+        } else {
+            done = true;
         }
         LazyBuffer {
             it: it,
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by bluss** [comment]
Thanks

**Comment by bluss** [comment]
The whole lazy buffer thing should mean we don't need to clone the iterator at all? I think Clone on the iterator can go away everywhere.

File: src/lib.rs [hunk 1]
```diff
@@ -794,6 +795,28 @@ pub trait Itertools : Iterator {
         Combinations::new(self)
     }
 
+    /// Return an iterator adaptor that iterates over the `n`-length combinations of
+    /// the elements from an iterator.
+    ///
+    /// Iterator element type is `Vec<Self::Item, Self::Item>`.
```
bluss: Just `Vec<Self::Item>`. We should stress that the iterator produces a vec per iteration, and clones the iterator elements.

src/adaptors.rs
```diff
@@ -1119,6 +1120,145 @@ impl<I> Iterator for Combinations<I> where I: Iterator + Clone, I::Item: Clone{
...
+}
+
+impl<I> LazyBuffer<I> where I: Iterator + Clone {
+    pub fn new(it: I) -> LazyBuffer<I> {
+        let mut it = it.clone();
+        let first = it.next();
+        let done = first.is_none();
+        let mut buffer = Vec::new();
+        if !done {
+            buffer.push(first.unwrap());
```
bluss: Don't use unwrap, `if let` is better.

salty-horse:
I can squish the commits into one once it's ready to merge (or you can do it).

PR #102, opened 2016-02-27
Add iterator method .fill_with(element)

bluss:


bluss:
Doesn't seem to be worth it, will wait for champion.

PR #103, opened 2016-02-29
Add .get(usize) to Stride

lud4ik:
https://github.com/bluss/rust-itertools/issues/61

### DIFF
--- a/src/stride.rs
+++ b/src/stride.rs
@@ -167,6 +167,19 @@ macro_rules! stride_impl {
             pub fn len(&self) -> usize {
                 ((self.end - self.offset) / self.stride) as usize
             }
+
+            /// Returns the element of a stride at the given index, or None if the index is out of bounds.
+            #[inline]
+            pub fn get(&self, i: usize) -> Option<$elem> {
+                if i >= self.len() {
+                    None
+                } else {
+                    unsafe {
+                        let ptr = self.begin.offset(self.offset + self.stride * (i as isize));
+                        Some(mem::transmute(ptr))
+                    }
+                }
+            }
         }
 
         impl<'a, A> Iterator for $name<'a, A>
--- a/src/stride.rs
+++ b/src/stride.rs
@@ -168,9 +168,10 @@ macro_rules! stride_impl {
                 ((self.end - self.offset) / self.stride) as usize
             }
 
-            /// Returns the element of a stride at the given index, or None if the index is out of bounds.
+            /// Return a reference to the element of a stride at the
+            /// given index, or None if the index is out of bounds.
             #[inline]
-            pub fn get(&self, i: usize) -> Option<$elem> {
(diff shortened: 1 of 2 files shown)


bluss:
Thanks. Do you want to do .get_mut too? I will leave some comments inline.

src/stride.rs
```diff
@@ -167,6 +167,19 @@ macro_rules! stride_impl {
             pub fn len(&self) -> usize {
                 ((self.end - self.offset) / self.stride) as usize
             }
+
+            /// Returns the element of a stride at the given index, or None if the index is out of bounds.
+            #[inline]
+            pub fn get(&self, i: usize) -> Option<$elem> {
```
bluss: Since the input is `&self` here the output must be a shared reference too `&A`. `$elem` is either `&A` or `&mut A`, so it's not correct for StrideMut.

bluss:
get_mut for StrideMut can be handled like `IndexMut` -- separately below the macro.

lud4ik:
Hope I've understood everything correctly, 'cos I'm novice in rust.

bluss:
This is good. I'll merge it.

src/stride.rs
```diff
@@ -266,6 +280,21 @@ impl<'a, A> Clone for Stride<'a, A> {
...
+impl<'a, A> StrideMut<'a, A> {
+    /// Return a mutable reference to the element of a stride at the
+    /// given index, or None if the index is out of bounds.
+    pub fn get_mut<'b>(&'b mut self, i: usize) -> Option<&'b mut A> {
+        if i >= self.len() {
+            None
+        } else {
+            unsafe {
+                let ptr = self.begin.offset(self.offset + self.stride * (i as isize));
+                Some(mem::transmute(ptr))
```
bluss: The macro uses transmute as a lazy way to work with both the mutable and immutable version. This being outside the macro, it can use a regular dereference here instead.

`&mut *ptr` makes a mutable reference from a raw pointer in an `unsafe` block.

I will fix this up when merging.

src/stride.rs
```diff
@@ -167,6 +167,20 @@ macro_rules! stride_impl {
             pub fn len(&self) -> usize {
                 ((self.end - self.offset) / self.stride) as usize
             }
+
+            /// Return a reference to the element of a stride at the
+            /// given index, or None if the index is out of bounds.
+            #[inline]
+            pub fn get<'b>(&'b self, i: usize) -> Option<&'b A> {
```
bluss: The lifetime parameter ('b) here is not needed, it matches what the default would be without it. Some of the methods around also have explicit lifetimes, I guess they are a bit old school.

I will fix this up while merging.

bluss:
Thank you!

lud4ik:
Thanks for explanation!

PR #108, opened 2016-03-28
Add partition_map

shepmaster:


src/lib.rs
```diff
@@ -1331,6 +1340,46 @@ pub trait Itertools : Iterator {
...
+    /// assert_eq!(failures, [false, true]);
+    /// ```
+    fn partition_map<A, B, F, L, R>(self, predicate: F) -> (A, B)
+        where Self: Sized,
+              F: Fn(Self::Item) -> PartitionTo<L, R>,
+              A: Default + Extend<L>,
+              B: Default + Extend<R>,
+    {
+        let mut left: A = Default::default();
+        let mut right: B = Default::default();
```
bluss: style: This is neater as `let mut right = B::default();` but I don't hold up PRs for small things that I can fix.

bluss:
This is simple and neat, we should have this. Thank you!

bluss:
I'll think over the type names before merge.

src/lib.rs
shepmaster: > This is neater as

Yes it is. I will fix it unless you don't want me to :-)

shepmaster:
> I'll think over the type names before merge.

I'm more than happy to discuss. I saw you had used `Left` and `Right` before (in addition to `Both`). Not sure if similar names is an upside or a downside.

src/lib.rs
shepmaster: Too late, I'm super fast at fixing my misteakez.

bluss:
I think it's fine, we could use a generic Either here, but libstd does not have one (and depending on crate either seems pointless). PartitionTo could be named just Partition? As a verb it makes sense.

shepmaster:
> PartitionTo could be named just Partition?

I can change this, I just didn't want to preclude any future `Partition` iterator adapter that could exist (which I don't know how it would...).

bluss:
You know the SizeHint hack that partition uses? (Or is it unzip?) Either way, I think it's good to *not* use it.

shepmaster:
> (Or is it unzip?) Either way, I think it's good to not use it.

Yeah, it's `unzip`. I think I had seen that while skimming, but hadn't really understood it (and I won't say that I do now, either). I'm glad my laziness / lack of knowledge led me down the right path :surfer:

shepmaster:
I've renamed `PartitionTo` to `Partition`.

bluss:
Great!

PR #109, opened 2016-03-31
Add fold_while (fold with short-circuit).

regexident:
I've been in need for a pure loop with short circuiting and return value.
So I wrote a μ-crate `fold_while` which enables just that.

And before cluttering crates.io with yet another one-trick-pony I figured I'd first try to get it added to the mother of all iter crates.

`fold_while` turns this:

```rust
let mut result = 0;
for i in (0..10) {
    if i > 5 { break; }
    result = result + i;
}
```

into this:

```rust
use itertools::Itertools;
use itertools::FoldWhile::{Continue, Exit};

let result = (0..10).fold_while(0, |acc, i| {
    if i > 5 { Exit(acc) } else { Continue(acc + i) }
});
```

I'd love to get this added to itertools.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -63,6 +63,7 @@ pub use adaptors::{
     Unique,
     UniqueBy,
     Flatten,
+    FoldWhile,
 };
 #[cfg(feature = "unstable")]
 pub use adaptors::EnumerateFrom;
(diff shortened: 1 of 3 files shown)


bluss:
Thanks, this makes sense. I'm not sure there's another good way to do this. Though I'm not afraid to break out the for loop myself when it's needed.

What do you think about renaming Exit? Maybe to Done or Return?

Can you also remove the Examples heading, such things just add noise in the docs. (This is how the docs render http://bluss.github.io/rust-itertools/doc/itertools/trait.Itertools.html#method.fold1 ) And put the method right below fold1.

regexident:
> Thanks, this makes sense. I'm not sure there's another good way to do this. Though I'm not afraid to break out the for loop myself when it's needed.

In most situations having a mutable value and simply `break`ing from a `loop` is the most pragmatic way to go, no doubt. But if you want to write pure and functional Rust code that also is efficient (as in not iterating over thousands of elements when the third already should have triggered `break`), then `fold_while` is pretty much the only way to go (in some situations [`scan()`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.scan) can help, but not when the next value depends n the current one).

> What do you think about renaming Exit? Maybe to Done or Return?

Sure. Pick one and I'll make the change. :wink:

Now that I think about it, most enums in Rust have Nouns as cases:
`Some`, `None`, `Ok`, `Err` (assuming `Ok` to be somewhat of a noun).
- So `Done` would probably be the idiomatic choice, while requiring `Continue` to be changed, too.
- While `Continue` + `Return` would deviate from that norm it makes the intentions very explicit and clear.

(My first version actually just used Result but I didn't like returning Ok/Err for loop control at all.)

> Can you also remove the Examples heading, such things just add noise in the docs. (This is how the docs render http://bluss.github.io/rust-itertools/doc/itertools/trait.Itertools.html#method.fold1 ) And put the method right below fold1.

Sure, will do.

---

If there's anything else you'd like me to adjust, …

src/lib.rs
```diff
@@ -86,6 +86,7 @@ pub use ziptuple::Zip;
 #[cfg(feature = "unstable")]
 pub use ziptrusted::{ZipTrusted, TrustedIterator};
 pub use zipslices::ZipSlices;
+pub use misc::FoldWhile as FoldWhile;
```
bluss: By the way, if this is exported in the top level it should not be in mod misc.

bluss:
Done seems good. Return is too close to the keyword I think.

Does pure and functional rust exist :wink:? I :heart: the [raw-pointer-fuckery](https://github.com/bluss/matrixmultiply/pull/2) part of rust, which is basically the opposite end..

src/lib.rs
regexident: Makes sense, any particular place you'd like me to put it?

bluss: it can be in lib.rs. It would be nice to split that file at some point though.

regexident:
> Done seems good. Return is too close to the keyword I think.

So `Done` it is then. :+1: 

> Does pure and functional rust exist :wink:? I :heart: the raw-pointer-fuckery part of rust, which is basically the opposite end..

And so do I!
I used to do C and C++ for the sake of performance and control.
Now it's Rust and I'm not looking back. :heart_eyes: 

If `fold_while` allows me to much more easily reason about my own code I'm happy already.
Doesn't have to be truly 100% pure to be useful, imho.

The machine learning project (SVM) I'm currently working on weights ~2500loc and a quick search revealed a total of 3 `let mut`. :sunglasses: 

What I use `fold_while` for is basically turning stuff like this:

```rust
let mut epochs = 0;
let mut weights = zero_weights;
for epoch in 0..max_epochs {
    if calculate_error(weights) < threshold { break }
    let (error, new_weights) = gradient_descent_step(state);
    weights = new_weights;
    epochs = epoch + 1;
});
```

into this:

```rust
let (epochs, weights) = (0..max_epochs).fold_while((0, zero_weights), |state, epoch| {
    if calculate_error(state.1) < threshold { Done(state) }
    else { Continue((epoch + 1, gradient_descent_step(state))) }
});
```

regexident:
I moved `FoldWhile` into `adaptors`. Hope that's fine with you. Should now match the rest of `itertools`.

bluss:
It's not an adaptor, so I'll have to move it sometime down the line. Where it's actually defined is just an implementation detail, so it doesn't matter much.

bluss:
Thank you

PR #113, opened 2016-04-02
Add minmax() and minmax_by_key() adapters (#79)

birkenfeld:
This time, stolen + adapted the old stdlib code.

I had to duplicate most of the code; the `minmax -> minmax_by_key(|x| x)` definition produced lifetime errors that I didn't find a fix for. Is there one?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -74,6 +74,7 @@ pub use intersperse::Intersperse;
 pub use islice::ISlice;
 pub use kmerge::KMerge;
 pub use linspace::{linspace, Linspace};
+pub use minmax::MinMaxResult;
 pub use pad_tail::PadUsing;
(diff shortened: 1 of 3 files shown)


bluss:
Cool. The code duplication isn't super fun, I wonder if we can fix that? Maybe with some refactoring, but it's a bit of work for deduplicating relatively little code.

birkenfeld:
OK, this should do it. It's not especially pretty, but no duplication anymore.

bluss:
I did not think of that solution, it's nice!

bluss:
Future possibilities here:

- Add quickcheck tests. Automatic testing is more exhaustive than single case tests
- Consider relaxing from Ord to PartialOrd. Then the user has the choice & flexibility to use the algorithm on floats if they want to.

PR #117, opened 2016-04-30
Add a note that the iterator must be fused for usage with `dropping`

tbu-:


bluss:
must is too strong

tbu-:
It will be incorrect if the iterator isn't fused and we hit the end. Should I phrase it like this?

bluss:
That is much better.

tbu-:
Changed.

src/lib.rs
```diff
@@ -985,6 +985,9 @@ pub trait Itertools : Iterator {
     /// It works similarly to *.skip(* `n` *)* except it is eager and
     /// preserves the iterator type.
     ///
+    /// This will produce incorrect results if the iterator is not *fused* and
+    /// the number of elements in the iterator is less than `n`.
+    ///
```
bluss: Ok, sorry, can we add another correction? Dropping will work fine if n is larger than the number of elements in the iterator. The error would be to try to get more items from the returned iterator if n is larger than the number of elements.

tbu-:
Changed. I'm not sure it's useful to provide this very long note instead of saying "This function assumes that the iterator is fused." and letting the user dive into the code if they want the exact behavior.

bluss:
I can write a shorter note if you want.

tbu-:
I'm not disappointed if you write your own note. :)

PR #122, opened 2016-05-10
Update adaptors.rs

### LINKED ISSUE #118: bound function not correct
See here: https://github.com/bluss/rust-itertools/blob/master/src/adaptors.rs#L142

Isn't it missing the case `self.phase && a < b`?

kindlychung:
fixes lower bound error

Fixes #118


bluss:
Thanks for reading & fixing this code that I thought nobody would care about. This can indeed be better. Quickcheck tests should use an exact test for interleave_shortest. Can you please add the testcase from your bug report though?

kindlychung:
Sure, can I just use `#[test]`?

bluss:
Yes

bluss:
Thanks, merged with test.

PR #123, opened 2016-05-10
avoid calculating b

kindlychung:
I figure this is more efficient, to return as early as possible, since when `a_cur` is None, it's not necessary to get the next item of b.


bluss:
This is not an IDE. Please do this locally.

kindlychung:
Yeah, I thought it was a minor change, turns out the borrow checker is not happy with me.

PR #138, opened 2016-08-21
Implemented minmax_by

phimuemue:
I implemented a basic version of `minmax_by` (similar to `sort_by`).

Note that it is (at least, currently) implemented to take an `Fn` (instead of an `FnMut`) as a comparison function. Do you think a `FnMut` comparison function would offer sensible advantages?

Just in case this matters: I also suggested to add `max_by` and `min_by` to the stdlib, so `minmax_by` would be in line with that (https://github.com/rust-lang/rust/pull/35856).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1530,6 +1530,25 @@ pub trait Itertools : Iterator {
     {
         minmax::minmax_impl(self, f, |_, _, xk, yk| xk < yk, |_, _, xk, yk| xk > yk)
     }
+
+    /// Return the minimum and maximum element of an iterator, as determined by
+    /// the specified comparison function.
+    ///
+    /// The return value is a variant of `MinMaxResult` like for `minmax()`.
+    ///
+    /// For the minimum, the first minimal element is returned.  For the maximum,
+    /// the last maximal element wins.  This matches the behavior of the standard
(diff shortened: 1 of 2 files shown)


bluss:
I use FnMut by "default", i.e everywhere it sensibly works. This should be an FnMut, that just permits more closures to be used (for example those modifying an upvar for some reason). Could be any reason, including modifying something for debugging/logging purposes.

The usual reason to restrict to Fn is if you need shared ownership/shared access or other concurrency, is how I understand it.

bluss:
It would be even cooler if there was a quickcheck test for minmax.

bluss:
I'd like to merge this. Using Fn over FnMut *is* required right, since we're sharing the comparison function between two locations (without some refactoring or internal mutability)?

bluss:
Using this, the iris.rs example can be updated to use minmax_by too.

phimuemue:
@bluss You're right, we are sharing the comparison function as of now.

Just my two cents: I suggest merging this pull request (after all, having `minmax_by` with an `Fn` is probably better than not having it at all), and opening a ticket regarding the `Fn/FnMut` issue (which I could take as soon as I have time).

bluss:
At the same time, path of least resistance usually leads us to simple, non-hacky solutions. It's not worth it to use interior mutability I think.

PR #140, opened 2016-08-26
Initial .starmap() implementation (WIP)

bluss:
.starmap() takes an iterator of n-tuples and allows applying an n-ary closure to it.

For example: `multizip((0..5, 1..5, 2..5)).starmap(|x, y, z| x + y + z)`


bluss:
The best would be if regular map was simply generic enough to support this transparently. That is probably not going to happen. This seems straightforward, but is it actually useful? I'm a bit hesitant.

PR #144, opened 2016-09-08
minmax_by takes FnMut

phimuemue:
`minmax_impl` took two comparison functions: less-than and greater-than. This is not necessary as far as I can see, since we can argue `less_than(x,y)=greater_than(y,x)`. If we changed `minmax_impl` to take only one comparison function, we could circumvent the `FnMut` borrowing issue.

### DIFF
--- a/src/minmax.rs
+++ b/src/minmax.rs
@@ -46,11 +46,10 @@ impl<T: Clone> MinMaxResult<T> {
 }
 
 /// Implementation guts for `minmax` and `minmax_by_key`.
-pub fn minmax_impl<I: Itertools, K, F, L, G>(mut it: I, mut key_for: F,
-                                             lt: L, gt: G) -> MinMaxResult<I::Item>
+pub fn minmax_impl<I: Itertools, K, F, L>(mut it: I, mut key_for: F,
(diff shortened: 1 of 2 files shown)


bluss:
This changes what happens at equality, and since rules for equality are already documented, I'm not sure we can do this.

bluss:
Maybe that comment was too quick. Time to do some closer reading

src/minmax.rs
```diff
@@ -46,11 +46,10 @@ impl<T: Clone> MinMaxResult<T> {
 }
 
 /// Implementation guts for `minmax` and `minmax_by_key`.
-pub fn minmax_impl<I: Itertools, K, F, L, G>(mut it: I, mut key_for: F,
-                                             lt: L, gt: G) -> MinMaxResult<I::Item>
+pub fn minmax_impl<I: Itertools, K, F, L>(mut it: I, mut key_for: F,
+                                             mut lt: L) -> MinMaxResult<I::Item>
```
bluss: Great PR. Indent/align mismatch here because of the G being removed above. The `I` type param should really be bounded by Iterator, not itertools. (And I: Sized is redundant).

Issue #87, opened 2015-12-30
Breaking changes in 0.5


bluss:
- [x] #86 
- [x] Possibly depend on use crate either  #129
- [x] Remove .map_fn(), MapFn
- [x] Remove .mend_slices
- [x] ~~Remove reexports `itertools::{enumerate, rev}` (exist in the free module)~~
- [x] Remove .into_rc()
- [x] Remove .sort_by (deprecated)
- [x] Remove ZipSlices (doesn't live up to its perf promise anymore due to optimizer changes)
- [x] Remove ZipTrusted (will have new life with specialization)
- [x] Rename Group by lazy and chunks lazy?
- [x] Remove Stride/StrideMut? Are separate crate material
- [x] Remove Linspace (doesn't work properly without using crate num-traits)
- [x] ~~Remove itertools::equal (Iterator.eq is the same).~~
- [x] Remove dropn (can't use .nth() like dropping does).

bluss:
Updated with proposed removals

bluss:
- [ ] Migrate from Struct::new to free function constructors

This is a tricky case of consistency. The libstd sets a precedence by functions like `std::iter::once`.

The argument for a free function is that the iterator's functionality is much more important than the identity of the type/struct involved.

The argument for Struct::new (for example `RcIter::new` and so on) is that it is consistent with other rust structs.

Candidates:

- `Zip::new` -> `zip` or `multizip`
- `RcIter::new` -> `rciter`
- probably more

Robbepop:
In my opinion, Zip::new(it1, it2, ...) is a superior syntax to it1.multizip(it2, ...). Of course this is a subjective matter but for a generalized zip with multiple parameters (and not just one as in the std lib) it just fits better.
However, for chaining of iterators the std-lib version fits better.
So in the end, maybe supporting both versions may be the best idea?

bluss:
the mentioned `multizip` would be a free function, not a method. Either way, both of them need to use a tuple, so the actual syntax is `Zip::new((it1, it2, ..))` or `multizip((it1, it2, ..))`.

Robbepop:
I like the idea very much in general. (free all the functions !! =D)

However, it isn't a Rusty-API style as you have mentioned above.

Replacing the old-syntax with the new would be a breaking change, I guess, so if you really implement this feature you would choose between breaking stuff or supporting both APIs for a while.

bluss:
Will be a breaking change, that's why it's planned in advance in this issue.

[...]

Issue #142, opened 2016-09-07
Add a generic Chain iterator


Robbepop:
It would be nice to have a ```Chain``` iterator in this library that could chain several iterators together instead of just too as it is supported in the std lib at the moment.
This should behave similar to the currently implemented ```Zip::new((..))```.

__Simple-silly example code:__

```rust
let v = vec![4,5,6];
let words: Vec<String> = vec!["Hello", "World", "!"];
for x in Chain::new((&[1,2,3], v.iter(), words.iter().map(|word| word.len()))) {
    println!("{}", x);
}
```

Important would be a possibility to also create a way to chain a bunch of iterators over **mutable** elements to obtain yet another iterator over mutable data.

__A (possibly buggy) example code:__

```rust
let (ax, bx, cx) = (&[1,2,3], &[4,5,6], &[7,8,9]);
for mut x in Chain::new((ax.iter_mut(), bx.iter_mut(), cx.iter_mut())) {
    x = x*x + x;
}
```

bluss:
If it works for general iterators, it will work for the mutable-reference carrying ones as well.

bluss:
I'm unsure if this is needed. It's not such a big thing as .zip(), since chain doesn't need to flatten anything. Chaining .chain() calls should work fine.

Robbepop:
To me it would be a superior syntax in cases where I need to chain more than two iterators together. But you are right that a Chain iterator isn't as needed as a Zip iterator was.

bluss:
I will decline this up front for now. Thanks anyway.

Supporting "n-ary" constructions in Rust just requires hacks, traits over several tuple types and so on. In this case it is not worth it (does not buy us enough), and the regular chaining style is truly n-ary for any *n*, not just a finite trait implementation list.

PR #148, opened 2016-09-19
Preparation for next release series (0.5.x)

### LINKED ISSUE #86: RepeatCall shouldn't implement DoubleEndedIterator
For example, the following iterator is definitely not reservable:

```rust
RepeatCall::new({
    let mut counter = 0;
    move || {
        counter += 1;
        counter
    }
});
```

bluss:
We will remove a lot of deprecated items, and rename some.

This is on the road towards an eventual 1.0 version of the library.

## Notes

- Numerical things will move because we don't want num-traits in itertools. `linspace` will move to a crate that uses num-traits
- Use free functions to construct iterators that are sources. For example `Unfold::new(x, y)` → `unfold(x, y)`. Follows libstd convention (`std::iter::once`).
- Unstable/nightly parts will need to live in another crate
- Rename and move stuff around to increase consistency. This is the best for the long run.
- See the commits for an overview

See also #87 

Fixes #86


bluss:
In general: Unstable stuff should go in another crate, not here.

Turn on --feature=unstable in itertools 0.4 to see deprecation warnings.

bluss:
Research shows StrideMut does have provable users in the wild, so deprecation will be reconsidered.

Stebalien:
LGTM so far.

However, personally I'd:

1. Either move  (now) or re-export (some day) the free functions in the root so one can call, e.g.,  `itertools::zip(a, b)` directly. It would also kind of be nice to get rid of that long list of adapter structs in the root of the documentation but I don't know of a good way to do this.
2. Provide a prelude module (export `Itertools` and the free functions)? This can obviously be done later.
3. Move (now) some of the helper traits out of `itertools::misc` into separate crates. I'd be happy to donate `tool`'s `Cons` trait if you want to maintain a crate for dealing with fixed-sized lists of things (splitting, joining, etc.) :wink:. To me, the name `misc` indicates that it doesn't really belong.

bluss:
All the structs could go into a new module like `itertools::types::`. The hard decision is whether to leave them reexported in the main module (deprecated one version) or not.

bluss:
Oh god, there is so much to do.

Any renaming scheme for these?

- group_by: groups as vectors
- group_by_lazy: groups as iterators,
- chunks_lazy: chunks as iterators

If `group_by` was removed and the others lost the lazy part of the name, it would be consistent but one less useful adaptor.

And these:
- combinations_n: combinations as vectors
- combinations: combinations as pairs

→ Some kind of neat scheme that can be expanded for using 3-tuples too. And is consistent if we add equivalent "windows" variants.

bluss:
@Stebalien The first order goal is to not have any dependencies in itertools

Stebalien:
> The first order goal is to not have any dependencies in itertools.

That's a bit unfortunate but understandable (although I don't think a few small `no_std` dependencies are really a problem).

> Any renaming scheme for these?

Is the non-lazy `GroupBy` adapter really that useful? Can't one just write: `iter.group_by_lazy(...).map(|(k, v)| (k, v.collect())`? Is there some performance penalty? Also, given some useful functions (that may not really be useful enough to include):

1. A free-floating `collect` function.
2. `Itertools::map_first` and `Itertools::map_second` .

One could write `iter.group_by(...).map_second(collect)` (renamed `group_by_lazy`).

bluss:
I think there's probably a performance penalty on the pretty large state machine thing in the group_by_lazy iterator. It's also got the distinction of being an IntoIterator adaptor confusingly enough, i.e. it doesn't return an iterator directly.

Ok. so that's what I *think*. Then I tried the benchmark. Take the benchmark that exist for group_by_lazy and try them with group_by.

The lazy iterators get the benefit of not making vectors here:

```
test group_by_1                               ... bench:      26,125 ns/iter (+/- 1,301)
test group_by_2                               ... bench:      41,493 ns/iter (+/- 1,653)
test group_by_lazy_1                          ... bench:       5,762 ns/iter (+/- 329)
test group_by_lazy_2                          ... bench:       8,282 ns/iter (+/- 197)
```

If we change so that we make vectors in the lazy benches too:

```
test group_by_1                               ... bench:      26,134 ns/iter (+/- 2,836)
test group_by_2                               ... bench:      40,151 ns/iter (+/- 1,547)
test group_by_lazy_1                          ... bench:      33,682 ns/iter (+/- 1,089)
test group_by_lazy_2                          ... bench:      52,549 ns/iter (+/- 2,432)
```

bluss:
Now the misc module is dead. The various helper traits can be replaced with just massive multidispatch impls of existing traits like Iterator and From.

Stebalien:
Hmm... this feels like something that could be fixed by specialization but I don't know the right abstraction.

malbarbo:
`multizip` is a great addition. Can a `multichain` be added? Should `multizip` and `multichain` be in `free` module?

How about merging `multizip` and `zip` (`multichain` and `chain`)? So one always use the same name without worrying about the number of arguments: `zip((a, b))` or `zip((a, b, c))`.

bluss:
We discussed multichain in  #142 and I put my thoughts there. What I mean is that `multizip(tuple)` is an ugly chack, and multichain does not motivate an ugly hack at all as much. `a.chain(b).chain(c)` is already truly n-ary for arbitrary n, and does not need the tuple flattening cleanup that multizip does.

I'm happy to have feedback and discussion about this. I think it would be a step back to require zip to take a tuple. .zip() specialization in libstd also has me thinking, either we need to reuse that in multizip somehow, or we need to replicate it. As it is in current nightly, `a.zip(b).zip(c)` can lead to very efficient autovectorized code if a, b, c are slices and the moon shines the right way on the optimizer and `multizip((a, b, c))` can't do this.

bluss:
Also note, multizip is not an addition but a new name for what was already there as `izip!()` and `Zip::new`.

bluss:
Regarding `free`, I'd like to move everything in there to the top level, that seems great to me.

malbarbo:
Thanks for the clarification.

Some other things.

How about adding `collect_` for all collection types in std?

- `collect_vec()` // already exist
- `collect_vec_deque()`
- `collect_linked_list()`
- `collect_string()`
- `collect_binary_heap()`
- `collect_hash_set()`
- `collect_hash_map()` for `Item = (_, _)`
- `collect_btree_set()`
- `collect_btree_map()` for `Item = (_, _)`

In my private code I have used `collect_hash_set` a lot in tests.

How about a way to collect to tuples?

```rust
// the problem here is that the produced items are lost if there are not enough items
let (a, b) = iter.next_tuple().unwrap(); // returns Option<(_, _)>
let (a, b, c) = iter.next_tuple().unwrap();

// or panic if there are not enough items
let (a, b) = iter.collect_tuple();
let (a, b, c) = iter.collect_tuple();

// and
for (a, b) in iter.tuples() { // this is similar to [T]::chuncks
}
```

bluss:
What I want to do now is to add `collect` as a free function, and not `collect_vec` but just `vec`. Free functions like `vec(Intoiterator) -> Vec<T>` and `hashset(IntoIterator) -> HashSet<T>` are very useful.

The static size 2 chunks is definitely interesting, how to handle uneven sizes though?

Also .next_tuple() is an interesting idea.

bluss:
Then again the 0.5 release bump is mostly focused on renaming and removing things! We can add new features whenever we want, but only break with a bump.

bluss:
@Stebalien Warning, you might not want to learn this. When I need a free function from a trait method I sometimes use this syntax: `<_>::collect`. Only works for trait methods.

Stebalien:
@bluss that's genius!

malbarbo:
The `collect` (and `vec`, etc) as a free function idea is great.

`tuples` could return a iterator backed by `MultiPeak`, so the last items are not consumed from the iterator. Note that I think `tuples` could be used for any tuple size (max 8?):

```rust
for (a, b, c) in iter.tuples() {}
for (a, b, c, d) in iter.tuples() {}
```

Please, keep the focus on 0.5 release. Maybe a new issue should be open so we can talk about collecting to tuples.

bluss:
I was wondering if this code using `<_>::fmt` was fugly or genious. https://github.com/bluss/rust-ndarray/blob/master/src/arrayformat.rs#L133-L151  Each one picks up the `fmt` method that is "primed" by the trait bound.  (I recently learned that the compiler works by kind of lifting the methods from an explicit trait bound into a priority class similar to that of inherent methods. That means that it will pick the particular method that corresponds to a trait bound, even if others would be applicable.)

bluss:
Ok -- this will merge and be a prerelease itertools 0.5.0-alpha.0. Feedback, breaking changes, takebacks are welcome until the final 0.5.0

bluss:
docs: https://docs.rs/itertools/0.5.0-alpha.0/itertools/

PR #151, opened 2016-09-23
Add Itertools::tuples_combination

malbarbo:


malbarbo:
This is an initial attempt to generalize `pair_combinations`. Is this viable?

I also have a working implementation for [`windows`](https://github.com/bluss/rust-itertools/issues/141) and `chunks` (called [`tuples`](https://github.com/bluss/rust-itertools/pull/148#issuecomment-248892454)) that produces tuples.

bluss:
Nice. Great API.

We won't depend on crate unreachable. I'd also try to not have any helper traits at all, having just killed off the `misc` module and the traits it had.

malbarbo:
The unreachable crate is not necessary, but makes the code faster:

Using `unwrap`

```
test comb_c1    ... bench:      41,952 ns/iter (+/- 5,528)
test comb_c2    ... bench:     128,934 ns/iter (+/- 213)
test comb_c3    ... bench:     200,215 ns/iter (+/- 386)
test comb_c4    ... bench:     317,023 ns/iter (+/- 432)
```

Using `unsafe_unwrap` from unreachable:

```
test comb_c1    ... bench:      43,054 ns/iter (+/- 4,764)
test comb_c2    ... bench:     128,971 ns/iter (+/- 227)
test comb_c3    ... bench:     158,748 ns/iter (+/- 430)
test comb_c4    ... bench:     249,699 ns/iter (+/- 853)
```

The only helper trait is `HasCombination`. I do not know how can this be removed. The problem is that each tuple size needs a iterator type that generates the combination, the away it is specified it by implementing `HasCombination` (that as `Combination` associated type). Any suggestion? Considering that it is intended to be implemented by tuples (not user types), is it enough to hide the trait in the docs?

What do think of the type inference limitation that appears in the `Itertools::tuple_combinations` doc?

bluss:
The type inference issue seems unproblematic. Only that the type of the iterator is going to be very verbose I think.

I was wondering, I'm not sure you can trust the iterator to return something that's not None at that point. This might be too lawyery, but [FusedIterator](https://doc.rust-lang.org/nightly/std/iter/trait.FusedIterator.html) is not an unsafe, so we can't trust `.fuse()` for unsafe code.

bluss:
I think this is perfect:

```
    /// // this requires a type hint
    /// let it = (1..5).tuple_combinations::<(_, _, _)>();
    /// itertools::assert_equal(it, vec![(1, 2, 3), (1, 2, 4), (1, 3, 4), (2, 3, 4)]);
```

bluss:
completely reasonable

bluss:
Some changes on top of that https://github.com/malbarbo/rust-itertools/pull/1

The implementation is great, I like it. It's not your problem, but Rust's problem, that we have to make a lot of helper traits and structs `pub`, and I don't really like it.

I'll remove pair_combinations in 0.5 to make space for adding this in, when it is ready.

bluss:
Ok, after much experimentation (also doing something with tuple_windows), I think the trait + associated type is necessary to get some measure of good type inference. Like your for loop example.

So even a tuple_windows iterator would use the same kind of scheme, I think.

malbarbo:
Ok, I read the discussion about using `FusedIterator` in unsafe code and I agree with you, we can't trust .fuse() for unsafe code.

I update the docs to include the type hint an also how to specify a `TupleCombinations` type.

I tried to rebase to solve a conflict in quick.rs and got a confusing log. To fix it I rebase and squash all commits. You modifications are in but do not show in the logs... Is it a problem? (Sorry, I do not have much experience with git...)

I also did a lot of experimentation, and the only away to get type inference in the for loop was using this code. For tuple_windows, I got it without using an associated type. When we finalize the tuple_combinations issues I will make another PR.

bluss:
Do you mind removing pair_combinations? Then we can have only this in 0.5. I don't want to rename it only to deprecate it, so it should just go away.

bluss:
I think this is going to be great. Same with tuple windows and so on.

malbarbo:
Done. size_hint is missing. I think it will require a bit more of thinking to write.

bluss:
That's fine. Merging this since we want to replace pair_combinations now before 0.5. Thanks!

PR #155, opened 2016-09-29
`size_hint` for `Interleave`

pthariensflame:
See (the accidentally created) pthariensflame/rust-itertools#1 for discussion.


bluss:
Thank you. We even have a quickcheck setup to test size hints (tests/quick.rs), I'll push such a test for interleave now.

PR #156, opened 2016-09-29
Add functionality similar to chunks and windows that produces tuples

### LINKED ISSUE #141: Add Windows-like Iterator that returns static-sized tuple elements instead of slices.
The [Windows](https://doc.rust-lang.org/std/slice/struct.Windows.html) iterator in the std lib returns its values as slices of elements.
This makes it unpleasant to work with since pattern matching and borrowing of slices is more restricted than (for example) for tuples.

Besides that a tuple-based Windows iterator should not involve runtime overhead and should make it possible to make compile time assumptions about the correctness of the implementation. For example, it can be checked at compiletime if accesses to iterator elements are "out-of-bounds".

E.g.:

```rust
    let primes = &[2, 3, 5, 7, 11, 13, 17, ...];
    for (fst, snd) in primes.iter().windows(2) {
        println!("({}, {})", fst, snd);
    }
```

malbarbo:
Fixes #141

### DIFF
--- a/src/tuples.rs
+++ b/src/tuples.rs
@@ -1,14 +1,11 @@
 //! Some iterator that produces tuples
 
-use std::marker::PhantomData;
 use std::iter::Fuse;
 
 /// An iterator over a incomplete tuple.
 ///
-/// See [`.next_tuple()`](../trait.Itertools.html#method.tuples),
-/// [`Tuples::into_buffer()`](struct.Tuples.html#method.into_buffer) and
-/// [`TupleWindows::into_parts()`](struct.TupleWindows.html#method.into_parts) for more
-/// information.
+/// See [`.next_tuple()`](../trait.Itertools.html#method.tuples) and
+/// [`Tuples::into_buffer()`](struct.Tuples.html#method.into_buffer).
 pub struct TupleBuffer<T>
     where T: TupleCollect
 {
@@ -106,20 +103,27 @@ pub struct TupleWindows<I, T>
 {
     iter: I,
     last: Option<T>,
-    buf: T::Buffer,
-    _marker: PhantomData<T>,
 }
 
 /// Create a new tuple windows iterator.
-pub fn tuple_windows<I, T>(iter: I) -> TupleWindows<I, T>
+pub fn tuple_windows<I, T>(mut iter: I) -> TupleWindows<I, T>
     where I: Iterator<Item = T::Item>,
-          T: TupleCollect
+          T: TupleCollect,
+          T::Item: Clone
 {
+    use std::iter::once;
+
+    let mut last = None;
+    if T::num_items() != 1 {
+        if let Some(item) = iter.next() {
+            let iter = once(item.clone()).chain(once(item)).chain(&mut iter);
+            last = T::collect_from_iter_no_buf(iter);
+        }
+    }
+
     TupleWindows {
+        last: last,
         iter: iter,
-        last: None,
-        buf: Default::default(),
-        _marker: PhantomData,
     }
 }
 
@@ -131,50 +135,16 @@ impl<I, T> Iterator for TupleWindows<I, T>
     type Item = T;
 
     fn next(&mut self) -> Option<T> {
+        if T::num_items() == 1 {
+            return T::collect_from_iter_no_buf(&mut self.iter)
+        }
         if let Some(ref mut last) = self.last {
             if let Some(new) = self.iter.next() {
                 last.left_shift_push(new);
-                Some(last.clone())
-            } else {
-                None
+                return Some(last.clone());
             }
-        } else {
-            self.last = T::collect_from_iter(&mut self.iter, &mut self.buf);
-            self.last.clone()
         }
-    }
-}
-
-impl<I, T> TupleWindows<I, T>
-    where I: Iterator<Item = T::Item>,
-          T: TupleCollect
-{
-    /// Return a pair with a buffer containing the items that was produced but not consumed and the
-    /// inner iterator.
-    ///
-    /// ```
-    /// use itertools::Itertools;
-    ///
-    /// let mut w = (0..10).tuple_windows();
-    /// assert_eq!(Some((0, 1, 2)), w.next());
-    ///
-    /// let (mut buffer, mut iter) = w.into_parts();
-    /// // Every produced item was consumed
-    /// assert_eq!(None, buffer.next());
-    /// // The next item is 6
-    /// assert_eq!(Some(3), iter.next());
-    ///
-    /// let mut w = (0..2).tuple_windows::<(_, _, _)>();
-    /// assert_eq!(None, w.next());
-    ///
-    /// let (buffer, mut iter) = w.into_parts();
-    /// // The items 0 and 1 was produced but not consumed
-    /// itertools::assert_equal(vec![0, 1], buffer);
-    /// // The is no more items
-    /// assert_eq!(None, iter.next());
-    /// ```
-    pub fn into_parts(self) -> (TupleBuffer<T>, I) {
-        (TupleBuffer::new(self.buf), self.iter)
+        None
     }
 }
 
@@ -185,9 +155,7 @@ pub trait TupleCollect: Sized {
     fn collect_from_iter<I>(iter: I, buf: &mut Self::Buffer) -> Option<Self>
         where I: IntoIterator<Item = Self::Item>;
 
-    // used on benchs
-    fn collect_from_iter_<I>(iter: I) -> Self
-        where I: IntoIterator<Item = Self::Item>;
+    fn collect_from_iter_no_buf<I>(iter: I) -> Option<Self> where I: IntoIterator<Item = Self::Item>;
 
     fn try_collect_from_iter<I>(iter: I) -> Result<Self, TupleBuffer<Self>>
         where I: IntoIterator<Item = Self::Item>
@@ -199,6 +167,8 @@ pub trait TupleCollect: Sized {
             Err(TupleBuffer::new(buf))
         }
     }
+
+    fn num_items() -> usize;
 
     fn left_shift_push(&mut self, item: Self::Item);
 }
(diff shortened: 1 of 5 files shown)


bluss:
Great! I'm on board with the APIs and the names I think, but I'd like to implement tuple windows and tuples without Vec.

This was my try when I was experimenting with the tuple combinations PR.

```rust
pub trait HasWindow<I>  where Self: Sized {
    type Window: From<I> + Iterator<Item = Self>;
}

/// See [`.tuple_windows()`](../trait.Itertools.html#method.tuple_windows) for more
/// information.
pub struct TupleWindows<I, T>
    where I: Iterator,
          T: HasWindow<I>,
{
    iter: T::Window,
    ty: PhantomData<I>,
}

pub struct TupleWindowsInner<I, T> {
    last: Option<T>,
    iter: I,
}


/// Create a new `TupleWindows` from an iterator with clonable items
pub fn tuple_windows<T, I>(iter: I) -> TupleWindows<I, T>
    where I: Iterator,
          I::Item: Clone,
          T: HasWindow<I>,
{
    TupleWindows {
        iter: <_>::from(iter),
        ty: PhantomData,
    }
}

impl<I, T> Iterator for TupleWindows<I, T>
    where I: Iterator,
          I::Item: Clone,
          T: HasWindow<I>,
{
    type Item = T;

        self.iter.next()
    }
}

macro_rules! impl_tuple_windows {
    ($A:ident; $($I:ident),* ; $($X:ident),*; $($X_rev:ident)*) => (
        impl<I: Iterator<Item=$A>, $A: Clone> HasWindow<I> for ($($I,)*) {
            type Window = TupleWindowsInner<I, Self>;
        }

        impl<I: Iterator<Item=$A>, $A> From<I> for TupleWindowsInner<I, ($($I,)*)> {
            fn from(mut iter: I) -> Self {
                'out: loop {
                $(
                    let $X = match iter.next() {
                        Some(v) => v,
                        None => break 'out,
                    };
                )*
                    return TupleWindowsInner {
                        last: Some(($($X,)*)),
                        iter: iter,
                    }
                }
                TupleWindowsInner {
                    last: None,
                    iter: iter,
                }
            }
        }

        impl<I, $A> Iterator for TupleWindowsInner<I, ($($I,)*)>
            where I: Iterator<Item = $A>,
                  I::Item: Clone,
                  ($($I,)*): HasWindow<I>,
        {
            type Item = ($($I),*);

                use ::std::mem::replace;
                let elt = self.last.clone();
                if let Some(x) = self.iter.next() {
                    if let Some(($(ref mut $X,)*)) = self.last {
                        let tmp = x;
                        $(
                            let tmp = replace($X_rev, tmp);
                        )*
                        drop(tmp);
                    }
                } else {
                    self.last = None;
                }
                elt
            }
        }
    );
}

impl_tuple_windows!(A; A, A ; a, b; b a);
impl_tuple_windows!(A; A, A, A ; a, b, c; c b a);
impl_tuple_windows!(A; A, A, A, A ; a, b, c, d; d c b a);
```

malbarbo:
I replaced the use of a vector in TupleWIndows with a tuple, like in your attempt. This makes more sense, there is less move to do (the benchmarks got improved).

The remaining uses of Vec is to return the produced but unconsumed items. We can remove this functionality or replace Vec with a stack allocated buffer (arrayvec?). What to you think?

[`[_]::windows`](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.windows) return no value if the windows size is greater than the slice. In [`[_]::chunks`](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.chunks) the last chunk can be smaller.

Adding fuse to the iterator cause a big regression in the benchmarks (inline issues). I will wait the decision on the use of Vec to investigate better.

bluss:
I am thinking that we don't need to save the overflow elements in either.

Especially tuple windows: The iterator element should be cheap, because we are copying it *n - 1* times for a *n* window. Typically a good use of tuple_windows will use an iterator where the element is a reference or an integer. (But we still have to use Clone, not Copy, otherwise it would restrict useful uses).

For this reason I think the elements are not "precious" in that sense.

Also: I don't think fuse is needed for tuple windows. Doesn't look like it in my PoC.

For `.tuples()` I understand that we could want to make the uneven tail end available. I would reach for a tuple of options before anything else (I haven't looked if that is feasible at all), but I would also accept just skipping / dropping them; I think most use cases will use this with iterators of length that divides *n*.

Having an efficient version of `.tuples()` would be awesome.

malbarbo:
I agree with you about the use of `tuple_windows`. Anyway, I replaced the use of vector with a new type: `TupleBuffer` (maybe it should be named `IncompleteTuple`).

I left `TupleWindows::into_parts` because it doesn't degrade the benchmarks, but it can be removed if you prefer. fuse as removed.

`tuples` is now faster than `slice::chunks(1, 2, 3)` in the benchmarks!

bluss:
Nice. Using `black_box` in the loop body is quite an artificial benchmark. I usually try something different now (some benchmarks in itertools are super old though, and still use that style).

I tried using a sum over `s[0]` or `s.0` of each window. And then I removed some stuff from tuple windows, and it seems to perform spectacularly in the n=2 and n=3 cases..

I push that quickly just to show.

bluss:
Comparison: https://github.com/malbarbo/rust-itertools/compare/tuples...bluss:tuple-windows-2?expand=1

That's an autovectorized sum in the `[T]::slice` case. That's an autovectorized sum in the tuple_windows case (for n=2 and n=3 only?)


```
test windows_slice_1       ... bench:       3,901 ns/iter (+/- 172)
test windows_slice_2       ... bench:       3,915 ns/iter (+/- 103)
test windows_slice_3       ... bench:       3,907 ns/iter (+/- 103)
test windows_slice_4       ... bench:       3,910 ns/iter (+/- 130)
test windows_tuple_1       ... bench:      71,290 ns/iter (+/- 676)
test windows_tuple_2       ... bench:       3,908 ns/iter (+/- 152)
test windows_tuple_3       ... bench:       3,907 ns/iter (+/- 109)
test windows_tuple_4       ... bench:      43,740 ns/iter (+/- 1,319)
```

malbarbo:
I changed the benchmarks intending making them more real, that is, making all the values of a chunks/windows to be read. This avoided the autovectorized sum.

Considering that iterators are supposed to be lazy, what are the issues of creating a item eargly in your `TupleWindows` constructor?

I tried you version but in the new benchmarks there are not much difference. 

```
test for_chunks_1    ... bench:       3,352 ns/iter (+/- 9)
test for_chunks_2    ... bench:       3,352 ns/iter (+/- 8)
test for_chunks_3    ... bench:       3,771 ns/iter (+/- 17)
test for_chunks_4    ... bench:       3,354 ns/iter (+/- 8)
test slice_chunks_1  ... bench:       2,521 ns/iter (+/- 11)
test slice_chunks_2  ... bench:       2,874 ns/iter (+/- 6)
test slice_chunks_3  ... bench:       3,407 ns/iter (+/- 16)
test slice_chunks_4  ... bench:       3,261 ns/iter (+/- 29)
test tuple_chunks_1  ... bench:       2,502 ns/iter (+/- 3)
test tuple_chunks_2  ... bench:       2,668 ns/iter (+/- 16)
test tuple_chunks_3  ... bench:       2,765 ns/iter (+/- 5)
test tuple_chunks_4  ... bench:       2,934 ns/iter (+/- 12)

test for_windows_1   ... bench:       2,502 ns/iter (+/- 5)
test for_windows_2   ... bench:       2,317 ns/iter (+/- 11)
test for_windows_3   ... bench:       2,644 ns/iter (+/- 3)
test for_windows_4   ... bench:       2,739 ns/iter (+/- 9)
test slice_windows_1 ... bench:       2,499 ns/iter (+/- 3)
test slice_windows_2 ... bench:       2,562 ns/iter (+/- 5)
test slice_windows_3 ... bench:       2,624 ns/iter (+/- 7)
test slice_windows_4 ... bench:       2,732 ns/iter (+/- 5)
test tuple_windows_1 ... bench:       2,517 ns/iter (+/- 4)
test tuple_windows_2 ... bench:       2,520 ns/iter (+/- 7)
test tuple_windows_3 ... bench:       2,667 ns/iter (+/- 5)
test tuple_windows_4 ... bench:       2,955 ns/iter (+/- 3)
```

bluss:
I really don't like or trust black_box inside the loop body. That inhibits any of the more interesting loop optimizations, and it's important to know if / when an iterator allow it, for example loop unrolling and autovectorization.

malbarbo:
I got what you sad. I removed the black_box uses.

The only away I could make tuple_windows faster was using `iter.next` in the constructor (this makes the TupleWindows::next method simpler, which I think allows the loop optimizations).

```
test for_windows_1   ... bench:         712 ns/iter (+/- 1)
test for_windows_2   ... bench:         965 ns/iter (+/- 21)
test for_windows_3   ... bench:       1,492 ns/iter (+/- 2)
test for_windows_4   ... bench:       2,061 ns/iter (+/- 4)
test slice_windows_1 ... bench:         213 ns/iter (+/- 0)
test slice_windows_2 ... bench:         917 ns/iter (+/- 7)
test slice_windows_3 ... bench:       1,088 ns/iter (+/- 2)
test slice_windows_4 ... bench:       1,484 ns/iter (+/- 18)
test tuple_windows_1 ... bench:         213 ns/iter (+/- 0)
test tuple_windows_2 ... bench:         920 ns/iter (+/- 11)
test tuple_windows_3 ... bench:       1,423 ns/iter (+/- 1)
test tuple_windows_4 ... bench:       2,750 ns/iter (+/- 2)
```

src/tuples.rs
```diff
@@ -0,0 +1,253 @@
...
+          T: TupleCollect,
+          T::Item: Clone
+{
+    use std::iter::once;
+
+    let mut last = None;
+    if T::num_items() != 1 {
+        if let Some(item) = iter.next() {
+            let iter = once(item.clone()).chain(once(item)).chain(&mut iter);
+            last = T::collect_from_iter_no_buf(iter);
```
bluss: Ah, this was strange but it makes sense. It's smart to do work upfront to have a simpler loop. I'd put a comment here to explain what it does, though.

bluss:
Thanks for all your hard work. I don't know what to say about calling next() or not in the constructor, I think it is ok (it certainly happens). A more principled approach could be possible, but I haven't thought that much about it.

bluss:
I think this looks good. I'm inclined to merge just so that we get on with it, it's too boring for us both to have it not be finished. Then I will do some minor adjustments myself. Thanks for writing such nice benchmarks and performance looks very good, doesn't it? I think these are going to be super useful as iterator adaptors.

malbarbo:
Yes, the performance looks very good! It took more time than what I expected, but was a good experience. Thanks for your observations (the black_box issues are more clear to me now!).

bluss:
Ok. I'm going to look closer at next_tuple now. It's principled to return a Result, but will the method have much use that way? I'd be inclined to just use `-> Option<T>` there (the tuple). If the user wants more control, they can use the tuples iterator.

malbarbo:
`next_tuple -> Option` is a good simplification. You are right, if the user wants more control, they can use the tuples iterator.

bluss:
Thanks a lot for this. Small edits on top in #158.

Just to ask the question: Why do we implement this for the unary tuples?

bluss:
Now part of itertools 0.5.2. I wonder if we should write a blog post about it.

malbarbo:
I think that we (should) support unary tuple for the same reason that (I think) unary tuple exist: uniformity.

For example, you can write a recursive macro (eg tuple combinations) that has a unary tuple as base case.

PR #160, opened 2016-10-14
Add iterate() function

### LINKED ISSUE #59: Add iterate() function

ArtemBan:
Fixes #59

### DIFF
--- a/src/sources.rs
+++ b/src/sources.rs
@@ -122,7 +122,7 @@ impl<A, St, F> Iterator for Unfold<St, F>
     }
 }
 
-/// An iterator that infinitely applies function to state and yields results.
+/// An iterator that infinitely applies function to value and yields results.
 ///
 /// This `struct` is created by the [`iterate()`] function. See its documentation for more.
 ///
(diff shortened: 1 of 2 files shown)


src/sources.rs
```diff
@@ -119,3 +121,47 @@ impl<A, St, F> Iterator for Unfold<St, F>
...
+        Some(mem::replace(&mut self.state, next_state))
+    }
+
+    #[inline]
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        (usize::max_value(), None)
+    }
+}
+
+/// Creates a new iterator that infinitely applies function to state and yields results.
```
bluss: The doc is actually missing the detail here that the "state" is updated with the new result each time. I think maybe it's best not to talk of it as "state" at all, but as applying a function on a value.

bluss:
Thank you. Can you if possible rebase your branch or repick the commits so that it's without the merge?

I see that as usual in Rust we have to make some choices about references or how to pass the arguments. The current version seems to be the most general, but I guess most users would expect a `FnMut(St) -> St` closure if they don't think about it. Do you think the current version is the best?

ArtemBan:
@bluss I think passing as reference is better because it's more general and it's more consistent with other iterator functions.

And things like `iterate(1, |&i| i *= 3).take_while(|&i| i < 100)` look prettier with it XD

bluss:
Not sure about the prettiness, more like uniformly ugly.

great, thank you.

bluss:
Released in 0.5.4, thank you.

I created an interesting mess by releasing 0.5.3 while not realizing I was not up to date with upstream master (and this merged PR). Fortunately releasing a new version is ridiculously easy.

PR #163, opened 2016-11-02
Add repeat_call_n() to generate n instances of a function

nabijaczleweli:
Tried to make this as close in looks as possible to `repeat_n()` and `repeat_call()`, but not sure if it's too great.


bluss:
Hi, I'm curious, what's the motivation for this addition? Does it have a cool use case? I'd think that using `repeat_call(f).take(n)` would be enough.

nabijaczleweli:
Welp, I'm an idiot, of course it's just like `repeat_call(f).take(n)`

bluss:
Don't worry. I kind of regret `repeat_n` already, it's too niche. But.. it had some uses in github code search, so I decided to keep it around.

PR #169, opened 2016-11-15
Add itertools::note_position.

luser:
`note_position` wraps each returned iterator element in a `Position<T>`
element, which indicates the position of the element in the iterator
results. This is useful for special-case handling of the first or
last elements of an iterator.

I wrote an implementation of this on a whim last week and someone suggested that itertools would be a good home for it. I've written lots of code in other languages where I've needed to handle the first or last
element of an iteration specially, and I was curious as to whether Rust would make writing a general-purpose implementation of that easy. Of course it did. :)


src/note_position.rs
```diff
@@ -0,0 +1,92 @@
+use std::fmt;
+use std::iter::Peekable;
+
+/// An iterator adaptor that wraps each element in an [`Position`](../enum.Position.html).
+///
+/// Iterator element type is `Position<I::Item>`.
+/// This iterator is *fused*.
```
bluss: The iterator is not fused. It doesn't need to be, just needs correct docs.

src/note_position.rs
```diff
@@ -0,0 +1,92 @@
...
+    First(T),
+    /// This is neither the first nor the last element.
+    Middle(T),
+    /// This is the last element.
+    Last(T),
+    /// This is the only element.
+    Only(T),
+}
+
+impl<T: PartialEq> PartialEq for Position<T> {
```
bluss: This is equivalent to derive(PartialEq), prefer using it.

src/note_position.rs
```diff
@@ -0,0 +1,92 @@
...
+            (&Position::First(ref a), &Position::First(ref b)) => a == b,
+            (&Position::Middle(ref a), &Position::Middle(ref b)) => a == b,
+            (&Position::Last(ref a), &Position::Last(ref b)) => a == b,
+            (&Position::Only(ref a), &Position::Only(ref b)) => a == b,
+            _ => false,
+        }
+    }
+}
+
+impl<T: fmt::Debug> fmt::Debug for Position<T> {
```
bluss: This is equivalent to derive(Debug) except for pretty printing, why not use deriving.

src/note_position.rs
```diff
@@ -0,0 +1,92 @@
...
+                // Peek to see if this is the last item.
+                match self.peekable.peek() {
+                    Some(_) => Some(Position::Middle(item)),
+                    None => Some(Position::Last(item)),
+                }
+            }
+            // Iterator is finished.
+            (_, None) => None,
+        }
+    }
```
bluss: Should forward size_hint from the underlying iterator.

src/lib.rs
```diff
@@ -894,6 +897,27 @@ pub trait Itertools : Iterator {
...
+    /// ease special-case handling of the first or last elements.
+    ///
+    /// Iterator element type is
+    /// [`Position<Self::Item>`](enum.Position.html)
+    ///
+    /// ```
+    /// use itertools::{Itertools, Position};
+    ///
+    /// let it = (0..4).note_position();
+    /// itertools::assert_equal(it, vec![Position::First(0), Position::Middle(1), Position::Middle(2), Position::Last(3)]);
```
bluss: This line is too long, keep it inside 100 columns please.

src/note_position.rs
```diff
@@ -0,0 +1,92 @@
...
+            &Position::Only(ref a) => write!(f, "Only({:?})", a),
+        }
+    }
+}
+
+impl<I: Iterator> Iterator for NotePosition<I> {
+    type Item = Position<I::Item>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        match (self.handled_first, self.peekable.next()) {
```
bluss: I would prefer if this method was written in a style where there's a branch `if self.handled_first` inside the match/conditional on .next()

bluss:
This looks interesting. Why do you need to separate first from only? Just wondering.

Also, is the name `.with_position()` better?

src/note_position.rs
luser: It is, although it doesn't use `Fuse` directly. `Peekable` is fused.

src/note_position.rs
```diff
@@ -0,0 +1,92 @@
...
+        handled_first: false,
+        peekable: iter.peekable(),
+    }
+}
+
+/// A value yielded by `NotePosition`.
+/// Indicates the position of this element in the iterator results.
+///
+/// See [`.note_position()`](trait.Itertools.html#method.note_position) for more information.
+pub enum Position<T> {
```
bluss: derive(Copy, Clone) are appropriate.

src/lib.rs
```diff
@@ -70,6 +70,7 @@ pub mod structs {
     pub use intersperse::Intersperse;
     pub use kmerge_impl::KMerge;
     pub use pad_tail::PadUsing;
+    pub use note_position::NotePosition;
```
bluss: this line is off by one from being in alpha order.

src/note_position.rs
bluss: I don't agree

luser:
I split `Only` out because if you are expecting to special-case the last element of an iterator and there's only one element, you'd miss it. If there's only one element it's the first and the last, but I don't think I'd want it to be returned twice, that seems more confusing.

I am open to whatever naming you want. I don't know of prior art in this area so I just sort of plucked that name out of thin air.

Thanks for the review! I tried to copy things from the surrounding context but I guess I missed a few things. :)

src/note_position.rs
bluss: Now you know what the deriving recipes do: literally what you typed :)

bluss:
The rest that I didnt' comment on was very good. That's always what the reviewer forgets to say.

src/note_position.rs
luser: Okay, I will remove the comment.

bluss: A fused iterator would pass the test called `exact_position` in this diff on top of your PR (as it was), and it doesn't. Peekable calls `.next()` on the underlying iterator, even if it has seen `None` already.

```rust
diff --git a/src/note_position.rs b/src/note_position.rs
index 58f8dc8..8bac7b1 100644
--- a/src/note_position.rs
+++ b/src/note_position.rs
@@ -89,4 +89,8 @@ impl<I: Iterator> Iterator for NotePosition<I> {
             (_, None) => None,
         }
     }
+
+        self.peekable.size_hint()
+    }
 }
diff --git a/tests/quick.rs b/tests/quick.rs
index a0176c8..9aa8681 100644
--- a/tests/quick.rs
+++ b/tests/quick.rs
@@ -346,6 +346,12 @@ quickcheck! {
     fn exact_interleave(a: Iter<i16>, b: Iter<i16>) -> bool {
         exact_size_for_this(a.interleave(b))
     }
+    fn exact_position(a: Iter<i16>) -> bool {
+        exact_size_for_this(a.note_position())
+    }
+    fn exact_position_2(a: Vec<i16>) -> bool {
+        exact_size_for_this(a.iter().note_position())
+    }
     fn size_interleave_shortest(a: Iter<i16>, b: Iter<i16>) -> bool {
         correct_size_hint(a.interleave_shortest(b))
     }
```

(`Iter` is a non-fused iterator that the quickcheck test suite uses).

bluss:
You know what, if you give the decision to me I'd prefer `.with_position()`

But now I have also contemplated `Position<T>` versus `(Position, T)` and I actually think the latter is better. The former is good if we want to force the user to handle each case exhaustively, but that's not necessarily the case. It seems harmless to allow `(Position, T)` instead?

bluss:
Maybe embedding T in position has a point, that it makes people not forget about the Only case.

src/note_position.rs
bluss: hm ok, after some thinking about peekable, it's actually pretty bad that it's not fused.

Calling `.peek()` and observing `None` is the only signal the user gets that the iterator is exhausted -- calling `.next()` after that leads to iterator-specific behaviour.

bluss:
I've reported rust-lang/rust/issues/37784 since Peekable's current behaviour makes it very hard to use, indeed this PR shows why (the implementation needs to use .fuse()).

src/note_position.rs
luser: Yuck. FWIW, I think I misread the std docs, which say `impl<I> FusedIterator for Peekable<I> where I: FusedIterator`, which makes sense.

Should I change this to `Peekable<Fuse<I>>` to work around that problem for now?

luser:
> But now I have also contemplated Position<T> versus (Position, T) and I actually think the latter is better. The former is good if we want to force the user to handle each case exhaustively, but that's not necessarily the case. It seems harmless to allow (Position, T) instead?

This argument is pretty compelling. If you only want to handle the first item specially, right now you'd have to write:
```rust
match item {
  Position::First(i) | Position::Only(i) => {}
  Position::Middle(i) | Position::Last(i) => {}
}
```

If the element type was a tuple you could instead write:
```rust
match item {
  (Position::First, i) | (Position::Only, i) => {}
  (_, i) => {}
}
```
with the downside that yeah, you could easily forget `Position::Only`. I guess an alternative here would be to instead do something like:
```rust
struct Position {
  pub is_first: bool,
  pub is_last: bool,
}
```

and then have the iterator produce `(Position, T)` tuples. Then you'd instead use it like:
```rust
for (pos, i) in iter.with_position() {
  if pos.is_first {
    //...
  }
  if pos.is_last {
    // ...
  }
}
```

What do you think?

bluss:
Enum is better. Rust enforces exhaustive matching or explicit opt out, so that's good enough?

luser:
This just needs to be merged, then.

bluss:
I'm sorry for keeping this waiting. This is going to be useful I think.

bluss:
I still disagree about documenting it as fused, but I can make small edits myself.

I'm adding .into_inner() -> T, but further methods could be useful too, just don't know what to name them.

For example `is_first() -> bool` (First, Only) and `is_strictly_first() -> bool` (First), maybe?

luser:
I think `is_first()`, `is_last()` make sense, because they each cover two enum variants. I don't know about `is_strictly_first()`, it seems like at that point you could just `match` on the enum.

PR #172, opened 2016-12-01
Fix performance of iteration over an multipeek adaptor

weiznich:
In case of calling peek often, the buffering vector will become large.
In those case each iteration step will copy all but the first element of
the vector one element to the right. Using a VecDeque fixes this
behaviour.

Code to reproduce this issue on master:
```rust
extern crate itertools;
use std::time::Instant;

const MAX: usize = 1_000_000;


fn main() {
    let data = (0..MAX).collect::<Vec<_>>();

    let mut mp = itertools::multipeek(data.into_iter());
    let mut counter_1 = 0;
    let start1 = Instant::now();
    while let Some(p) = mp.peek() {
        counter_1 += *p;
    }
    println!("Iterating using peek: {:?} -> {}",
             start1.elapsed(),
             counter_1);
    let mut counter_2 = 0;
    let start2 = Instant::now();
    for d in mp {
        counter_2 += d;
    }
    println!("Iterating using iterator: {:?} -> {}",
             start2.elapsed(),
             counter_2);
    assert_eq!(counter_1, counter_2);
}
```

### DIFF
--- a/src/adaptors.rs
+++ b/src/adaptors.rs
@@ -9,7 +9,7 @@ use std::fmt;
 use std::mem::replace;
 use std::ops::Index;
 use std::iter::{Fuse, Peekable};
-use std::collections::HashSet;
+use std::collections::{HashSet, VecDeque};
 use std::hash::Hash;
 use std::marker::PhantomData;
 use size_hint;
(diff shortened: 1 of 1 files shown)


bluss:
Thanks. It's a trade off, but this is the winner when there are many peeks.

By the way, #93 is still open.

PR #179, opened 2016-12-21
Add quickcheck test for minmax method.

### LINKED ISSUE #145: Add quickcheck tests for minmax methods

theemathas:
Fixes #145.

I decided to duplicate the `Val` type from the `minmax` function in the `test/tests.rs` file into the `test/quick.rs` file. I could not think of an alternative that does not require duplication.

P.S. This is my first pull request I made on someone else's repository. If I did something wrong, please tell me.


bluss:
Thanks!

PR #183, opened 2016-12-23
Add .reset_peek() to multipeek

### LINKED ISSUE #180: Ability to reset peek in MultiPeek
While working on a parser, I encountered a case where I'm inside a function trying to determine if I should bail out or not depending on the next two characters.

MultiPeek allows me to do just that, but when I get back into the main function, I can't reason about the state of the iterator.

The next peek may be the iterator position +1, or +3 depending on how I got to that place.

What I think I'd like to see is the ability to "reset" my peek back to 0 so that the next peek will look up just one item ahead of the position.

Is that doable?

bluss:
Fixes #180


EfficiencySimplicity:
This is quite a helpful feature! Thanks from 10 years later.

PR #187, opened 2017-01-20
Replace kmerge binary heap with std library binary heap

windwardly:
Hi all, I'm Proposing to switch kmerge back to the original author's use of BinaryHeap, but make use of features to give the same (theoretical) efficiency of the recent versions: don't pop and push, just sort_down after reading an element out of the top iterator using PeekMut.

A new [feature](https://github.com/rust-lang/rust/commit/f2cb47adf95171e48a9f918aae00193a7d36c7c0) in nightly adds the ability to pop the top instead of sorting down, for when the iterator is empty.

This PR includes two versions of `KMerge::next()` : one using the `PeekMut::pop()` feature and one without.
Build with `cargo build --features nightly` to use the new feature.

The good new and bad news:
```
bash-3.2$ cargo benchcmp perf.local perf.stdlib | grep kmerge
 kmerge_default                6,108               17,788                     11,680  191.22% 
 kmerge_tenway                 268,831             272,354                     3,523    1.31% 
```

The bad news is some situations get much worse.  The good news is that this puts pressure on the std libcollections/binary_heap.rs to possibly improve as:
1. putting the local version of sort_down into a locally built std lib got most of the perf back.
2. In PeekMut::drop(), putting a precondition to only call sort_down if the tip would actually move got half of the performance back (with no change to sort_down).
The test case has long vectors that sometime have sequences. The local version of sort_down does no writes (swap) until something moves. The std library version has some writes even if nothing moves.

Any more analysis is beyond the scope of this fix, as it relates to altering the std library.
Thanks.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ default-features = false
 version = "0.2"
 
 [features]
+nightly = []
 
 [profile]
 bench = { debug = true }
--- a/src/kmerge_impl.rs
+++ b/src/kmerge_impl.rs
@@ -4,6 +4,9 @@ use Itertools;
 
 use std::cmp::Ordering;
 use std::mem::replace;
+use std::collections::BinaryHeap;
+#[cfg(feature = "nightly")]
+use std::collections::binary_heap::PeekMut;
 
 macro_rules! clone_fields {
     ($name:ident, $base:expr, $($field:ident),+) => (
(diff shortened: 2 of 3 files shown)


bluss:
Hi thank you, but itertools will not depend on unstable features at this point.

windwardly:
I hadn't realize 1.11 was considered the stable baseline and had been testing on current stable 1.14 .
Okay. I'll withdraw.

Thank you for itertools. It's the kmerge code has been a fabulous way to dig into Rust's std libraries to understand shortcomings in the std libraries.

bluss:
That's interesting. Itertools doesn't use the `unsafe`-code optimization that stdlib uses to replace the element swap in sift down with just a move of the element.

But you are right, the stop condition in sift_down is probably the major difference. I didn't remember about that one.

PR #189, opened 2017-01-23
Add iter! macro

malbarbo:
A temporary solution for the missing `IntoIterator` implementation for `[T; N]` (https://github.com/rust-lang/rust/issues/25725). While I was finishing this PR another issue was opened https://github.com/rust-lang/rust/issues/39257.

When `IntoIterator` get implemented for `[T; N]` an user can port their code removing the `iter!` macro invocations.

My main use case for this is to use the free functions in `itertools`, like `max`, with an arbitrary number of elements. The benchmarks show no difference for `maxN(a, b, ...)` and `max(iter!(a, b, ...))`.

This introduces a dependency on `nodrop`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ test = false
 
 [dependencies]
 either = { version = "1.0", default-features = false }
+nodrop = "0.1.8"
 
 [dev-dependencies.quickcheck]
 version = "0.4"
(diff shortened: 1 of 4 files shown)


bluss:
Hi, this is of course useful but it's not necessarily for itertools, it's better as its own crate, especially since it involves a series of hacks (It's unfortunate that Rust does not yet support it better, simply!). Previous existing ones solutions were [literator](https://github.com/kmcallister/literator) and arrayvec, I guess.

bluss:
Closing with the motivation in preceding comment.

PR #192, opened 2017-03-23
ENH: Relax minmax to PartialOrd

bluss:
The rationale is that it's the user's choice to input floating point elements or not. The restriction to `Ord` is too strict for regular use.


bluss:
cc @birkenfeld as the original author.

It breaks with libstd conventions; I think they are too strict.

birkenfeld:
Sure!

bluss:
Good to hear from you.

birkenfeld:
Thanks for your great work.

PR #194, opened 2017-04-01
Add .kmerge_by()

### LINKED ISSUE #191: Add kmerge_by
Hi there!

I have a scenario where I have multiple iterators and would like to join them via a kmerge; however, currently I would need to implement `Ord` for my type. This becomes hard as the sorting is only relevant to the specific algorithm generating this iterator. I don't want to have `Ord` implemented for the type as it will behave differently based on the scenario.

Would it be possible to be able to pass a closure to something like `kmerge_by` to determine the ordering?

bluss:
The heap implementation is bespoke. (We could possibly have used
libstd's BinaryHeap but not before PeekMut's pop goes stable).

Change the heap implementation to use an ordering closure. Also change
its default formulation to be a min-heap (ordering lesser objects
first). Also relax constraints to use PartialOrd.

The usual thinking is that it should be the user's choice how strict
they want to be about Ord, not our place to restrict; kmerge only needs
“is a before b?”.

The chosen formulation for .kmerge_by() is to take a closure
`first: FnMut(&T, &T) -> bool`. By default this is the less than operator,
if that makes it clear.

Fixes #191

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -68,7 +68,7 @@ pub mod structs {
     pub use format::{Format, FormatWith};
     pub use groupbylazy::{IntoChunks, Chunk, Chunks, GroupBy, Group, Groups};
     pub use intersperse::Intersperse;
-    pub use kmerge_impl::KMerge;
+    pub use kmerge_impl::{KMerge, KMergeBy};
     pub use pad_tail::PadUsing;
     pub use peeking_take_while::PeekingTakeWhile;
     pub use rciter_impl::RcIter;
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -623,11 +623,41 @@ pub trait Itertools : Iterator {
...
+    /// let mut it = vec![a, b].into_iter().kmerge_by(|a, b| a.abs() < b.abs());
+    /// assert_eq!(it.next(), Some(0.));
+    /// assert_eq!(it.last(), Some(-7.));
+    /// ```
+    fn kmerge_by<F>(self, first: F)
+        -> KMergeBy<<<Self as Iterator>::Item as IntoIterator>::IntoIter, F>
+        where Self: Sized,
+              Self::Item: IntoIterator,
+              F: FnMut(&<<Self as Iterator>::Item as IntoIterator>::Item,
+                       &<<Self as Iterator>::Item as IntoIterator>::Item) -> bool
```
bluss: (Shorter ways or private type aliases are not available in Rust 1.11).

PR #196, opened 2017-04-11
Add all_equal

phimuemue:
This refers to https://github.com/bluss/rust-itertools/issues/175. I think going with `dedup` is a convenient, simple solution.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -997,6 +997,28 @@ pub trait Itertools : Iterator {
         None
     }
 
+    /// Check whether all elements compare equal.
+    ///
+    /// Empty iterators are considered to have equal elements:
+    ///
+    /// ```
+    /// use itertools::Itertools;
+    ///
+    /// let data = vec![1, 1, 1, 2, 2, 3, 3, 3, 4, 5, 5];
+    /// assert!(!data.iter().all_equal());
+    /// assert!(data[0..3].iter().all_equal());
+    /// assert!(data[3..5].iter().all_equal());
+    /// assert!(data[5..8].iter().all_equal());
+    ///
+    /// let data : Option<usize> = None;
+    /// assert!(data.into_iter().all_equal());
+    /// ```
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -732,6 +732,29 @@ pub trait Itertools : Iterator {
...
+    /// let data = vec![1, 1, 1, 2, 2, 3, 3, 3, 4, 5, 5];
+    /// assert!(!data.iter().all_equal());
+    /// assert!(data[0..3].iter().all_equal());
+    /// assert!(data[3..5].iter().all_equal());
+    /// assert!(data[5..8].iter().all_equal());
+    ///
+    /// let data : Option<usize> = None;
+    /// assert!(data.into_iter().all_equal());
+    /// ```
+    fn all_equal(self) -> bool
```
bluss: Ok, it's not always easy to decide, but I think this method should use `&mut self` and no `Self: Sized`, just like `all`. What do you think?

bluss: By the way, the method should be sorted among the methods, not the adaptors, so it needs to be moved down. next_tuple is the first non-adaptor in the trait.

phimuemue:
I think we could use `&mut self`, but when I reconsidered the problem I found another issue that I think is possibly even more important: Do you think it would be wise to return `Option<Item>` instead of `bool` so that the users have the possibility to get a value if all are considered equal?

bluss:
That's a good extension. But we need to find a new name of it then!

Hehe, I think of a good feature as being a combination of a natural name and a wanted functionality. We need both for it to be used.

phimuemue:
I re-checked with `cargo +nightly test` on my machine and obtained `test result: ok. 74 passed; 0 failed; 5 ignored; 0 measured` as output.

I think the travis failure is related to https://github.com/rust-lang-nursery/rustup.rs/issues/1062.

bluss:
Great, thank you!

PR #197, opened 2017-04-21
New iterator adapter allowing feedback loops

jameysharp:
I want to be able to express cyclic graphs of iterator adapters (with a one-cycle delay at every back-edge, to preserve causality) so I wrote some Rust to do it, and `itertools` seems like a natural home for it. My use-case is for digital signal processing, but I think this captures a general dataflow pattern that isn't covered by other iterator tools.

My simple example from the doctest:

```rust
use itertools::feedback;
let input = [1, -2, 3, -4, 5];
let result: Vec<i32> = feedback(0, |feedback|
        feedback.zip(&input)
                .map(|(a, b)| a + b)
    ).collect();
assert_eq!(result, &[0, 1, -1, 2, -2, 3]);
```

I have some concerns about my implementation so I'll understand if you don't want to just merge it. :smile:

1. `feedback` takes a callback whose job is to construct the inner pipeline of iterator adapters. At the head of that pipeline is an iterator provided by the `feedback` function itself. The type of that iterator should not be exposed to users of the library, but it has a concrete type and I want the callers to get monomorphized using the actual type so everything can be inlined appropriately. I'm currently using the approach documented in https://github.com/rust-lang/rust/issues/34537#issue-162798003 to use a private type in the public API without compiler warnings or errors, but that feels like a hack. Is there a better option?

2. I _think_ we can prove statically that the lifetime of the internal `Rc` is bounded by the lifetime of the `Feedback`, so this shouldn't need an `Rc` or even any heap allocation. But I'm not sure how to convince the compiler. This at least seems unlikely to change public API, so I think it doesn't matter whether it's fixed before merging.

3. There might well be another way to do this, though I couldn't find it. I considered both `Iterator::scan` and `itertools::unfold`, but neither one lets me construct the inner dataflow graph by composing other iterator adapters.

4. This could probably stand to have more tests than just the doctest, but I'm not sure what they should cover.

Thoughts? Thanks!

### DIFF
--- a/src/feedback.rs
+++ b/src/feedback.rs
@@ -1,0 +1,74 @@
+use std::cell::Cell;
+use std::rc::Rc;
+
+mod inner {
+    use std::cell::Cell;
+    use std::rc::Rc;
+
+    pub struct Inner<T>(pub Rc<Cell<Option<T>>>);
+
+    impl<T: Copy> Iterator for Inner<T> {
+        type Item = T;
+
+        fn next(&mut self) -> Option<T> {
+            self.0.get()
+        }
+    }
+}
+
+/// See [`feedback`](../fn.feedback.html) for more information.
+pub struct Feedback<I, T> {
+    iter: I,
+    inner: Rc<Cell<Option<T>>>,
+}
+
+/**
+ * Feed the output of an iterator pipeline back to the input. Feedback
+ * is delayed by one timestep to preserve causality, so the first input
+ * is provided by `initial`, and the output of that pass is used as the
+ * input for the second pass, and so on.
+ *
+ * Every time the input is requested, it yields the last result returned
+ * by the pipeline. The pipeline can request the feedback value any
+ * number of times per cycle, including ignoring it entirely. If the
+ * pipeline doesn't request a particular input, that input is discarded,
+ * not saved for the next cycle. If the pipeline requests an input
+ * multiple times in the process of producing an output, the same input
+ * will be returned each time.
+ *
+ * ```rust
+ * use itertools::feedback;
+ * let input = [1, -2, 3, -4, 5];
+ * let result: Vec<i32> = feedback(0, |feedback|
+ *         feedback.zip(&input)
+ *                 .map(|(a, b)| a + b)
+ *     ).collect();
+ * assert_eq!(result, &[0, 1, -1, 2, -2, 3]);
+ * ```
+ */
+pub fn feedback<F, I, T>(initial: T, inner: F) -> Feedback<I, T>
+    where F: FnOnce(inner::Inner<T>) -> I,
+          T: Copy
+{
+    let initial = Rc::new(Cell::new(Some(initial)));
+    Feedback {
+        iter: inner(inner::Inner(initial.clone())),
+        inner: initial,
+    }
+}
+
+impl<I, T> Iterator for Feedback<I, T>
+    where I: Iterator<Item = T>,
+          T: Copy
+{
+    type Item = T;
+
+    fn next(&mut self) -> Option<T> {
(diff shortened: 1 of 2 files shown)


jameysharp:
Any questions or comments I can address on this pull request?

bluss:
I'm sorry that I haven't had time to look at this before. It looks very cool, it takes some work to wrap one's head around.

We must in fact use Rc since the iterator is given away to the closure, and we can't protect against them storing it away somewhere else (variable of outer scope or otherwise). The approach I had to try to convert it to use references was to use a closure like this: `where F: FnOnce(inner::Inner<T>) -> I` but then the an iterator adapting `Inner` can't be returned from the closure either, so it defeats itself too much.

I think it requires an API change, and the typical Rust way of resolving it is to put the value that's borrowed into a separate struct, so it is possible to make it Rc free by making it less neat like this:

```rust
use std::cell::Cell;

mod inner {
    use std::cell::Cell;

    pub struct Inner<'a, T: 'a>(pub &'a Cell<Option<T>>);

    impl<'a, T: Copy> Iterator for Inner<'a, T> {
        type Item = T;

        fn next(&mut self) -> Option<T> {
            self.0.get()
        }
    }
}

pub struct Feedback<T> {
    t: Cell<Option<T>>,
}

pub struct FeedbackIter<'a, I, T: 'a> {
    iter: I,
    inner: &'a Cell<Option<T>>,
}

/**
 * Feed the output of an iterator pipeline back to the input. Feedback
 * is delayed by one timestep to preserve causality, so the first input
 * is provided by `initial`, and the output of that pass is used as the
 * input for the second pass, and so on.
 *
 * Every time the input is requested, it yields the last result returned
 * by the pipeline. The pipeline can request the feedback value any
 * number of times per cycle, including ignoring it entirely. If the
 * pipeline doesn't request a particular input, that input is discarded,
 * not saved for the next cycle. If the pipeline requests an input
 * multiple times in the process of producing an output, the same input
 * will be returned each time.
 *
 * ```rust
 * use itertools::feedback;
 *
 * let input = [1, -2, 3, -4, 5];
 * let result: Vec<i32> = feedback(0).feed(|feedback|
 *         feedback.zip(&input)
 *                 .map(|(a, b)| a + b)
 *     ).collect();
 * assert_eq!(result, &[0, 1, -1, 2, -2, 3]);
 * ```
 */
pub fn feedback<T>(initial: T) -> Feedback<T>
    where T: Copy
{
    Feedback {
        t: Cell::new(Some(initial)),
    }
}


impl<T> Feedback<T> where T: Copy
{
    pub fn feed<'a, F, I>(&'a mut self, inner: F) -> FeedbackIter<'a, I, T>
        where F: FnOnce(inner::Inner<'a, T>) -> I,
              T: Copy
    {
        FeedbackIter {
            iter: inner(inner::Inner(&self.t)),
            inner: &self.t,
        }
    }
}

impl<'a, I, T> Iterator for FeedbackIter<'a, I, T>
    where I: Iterator<Item = T>,
          T: Copy
{
    type Item = T;

    fn next(&mut self) -> Option<T> {
        if let Some(v) = self.inner.get() {
            self.inner.set(self.iter.next());
            Some(v)
        } else {
            None
        }
    }
}
```

bluss:
Closing for inactivity. Sorry about that. 

For some strange reason, I published https://docs.rs/itertools-wild/ and it includes your feedback iterator and some of my changes. Please tell me if/how I can credit you better there, like email or full name.

We can also look at it from the docs perspective there:

https://docs.rs/itertools-wild/0.1.1/itertools_wild/special/fn.feedback.html

When I read the docs for that now, I still can't recall exactly how it works :smile:

PR #199, opened 2017-05-16
Replace RefCell with Cell in Format and FormatWith

qthree:
I've been playing around with iterators, using itertools as reference, and found unnecessary RefCells in Format and FormatWith structs. There isn't that much overhead, but why not throw it off if possible? Also, alternative Clone implementation and test for that.
My first pull request for Rust project, please take care of me.

### DIFF
--- a/src/format.rs
+++ b/src/format.rs
@@ -1,5 +1,5 @@
 use std::fmt;
-use std::cell::RefCell;
+use std::cell::Cell;
 
 /// Format all iterator elements lazily, separated by `sep`.
 ///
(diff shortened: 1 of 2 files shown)


qthree:
Oh, i see. Non-copy Cell appears only after 1.17.0, never mind then.

bluss:
Requiring the iterator to be clonable is much too restrictive -- anything with a regular closure then is excluded. For this reason I think we don't want to do this.

bluss:
Oh I see, Clone is not actually needed for .take(). That's brilliant. Then this is a good idea when the version requirement allows.

PR #200, opened 2017-05-19
add maxn and minn methods of finding n largest/smallest elements

Fiedzia:


bluss:
Hi, thank you.

This seems reasonable, but we must first answer some questions around inclusion into itertools. I think it fits the scope of the crate.

- It is expected that the operation is efficient. Do we need to use quickselect or similar to reach this?
- Is it still in scope of the crate if we use a bigger algorithm like that?

Implementation specific review waits for now

Fiedzia:
This implementation would be efficient for small n for elements of small size, which I think is enough for it to be useful. I am not sure if there is more efficient way of selecting large n of elements of small size.
To be efficient for elements of large size, slightly different approach would be required.

> It is expected that the operation is efficient.

That's not very precise question. Perhaps better one would be "Shall we provide an implementation that we know could be improved in some aspects". For that I'd said yes, because it give a chance to involve more people who may come up with improvement.

bluss:
Thanks for posting this — at the end of the day, itertools is not going to contain all possible iterator related functionality and we will decline to carry this feature.

1. The n-largest/n-smallest problem deserves attention with a good algorithm for the problem, and it's not itertools role to contain significant bigger algorithms, they are fine in their own crate
2. We prioritise iterator and iterator-to-iterator API, so an adaptor would be more interesting than returning a vector here.

The following crate seems to offer an iteresting iterator adaptor interface to partially sorted sequences. https://github.com/benashford/rust-lazysort

PR #201, opened 2017-05-21
Inexact size hint testing

rkarp:
This extends the existing test iterator with a mode that returns inexact size hints randomly. When generated by `Arbitrary` with the `Inexact` type parameter, `Iter` will now under / overestimate its lower / upper bounds, so we can test the size hints more thoroughly.

The new tests failed in 4 cases, caused by 2 `size_hint` implementation bugs (wrong upper bound in `Product`, overflow in `InterleaveShortest`) and a bad trait constraint for `PadUsing`. I fixed them all, but the fix for `PadUsing` is a breaking change.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 /target/
 Cargo.lock
+/.idea/
+/*.iml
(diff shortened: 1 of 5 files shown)


bluss:
Thanks a lot for this, looks like some great work and more precision where it was needed.

bluss:
The only part I will not accept is the change to .gitignore, which the maintainer can easily adjust and it does not need to hold up the PR. Thanks!

Issue #166, opened 2016-11-06
Maybe add first_error() shortcircuit on first error


bluss:
Funky adaptor that lazily strips Results and stops on error.

```rust

pub struct FirstError<'a, I, E: 'a> {
    iter: I,
    error: &'a mut Result<(), E>,
}

pub fn first_error<T, E, I>(slot: &mut Result<(), E>, iter: I) -> FirstError<I::IntoIter, E>
    where I: IntoIterator<Item=Result<T, E>>,
{
    FirstError {
        iter: iter.into_iter(),
        error: slot,
    }
}

impl<'a, I, T, E> Iterator for FirstError<'a, I, E>
    where I: Iterator<Item=Result<T, E>>,
{
    type Item = T;
    fn next(&mut self) -> Option<Self::Item> {
        match self.iter.next() {
            Some(Ok(x)) => Some(x),
            Some(Err(e)) => {
                *self.error = Err(e);
                None
            }
            None => None,
        }
    }
    
    fn size_hint(&self) -> (usize, Option<usize>) {
        let (_, hi) = self.iter.size_hint();
        (0, hi)
    }
}

fn main() {
    let mut err = Ok(());
    
    let data = "1, 2, abc, 4";
    
    for line in first_error(&mut err, data.split(", ").map(str::parse::<u32>)) {
        println!("{}", line);
    }
    println!("Got error: {:?}", err);
}

```

bluss:
Fixed in #203

PR #203, opened 2017-06-14
Add ProcessResults

shepmaster:
This has been living in the standard library for a while and seems
like it could be more widely useful.

TODO

- [x] Missing `IntoIterator` conversions
- [x] Missing `size_hint`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1703,3 +1703,76 @@ impl<T> FoldWhile<T> {
     }
 }
 
+pub struct ProcessResults<'a, I, E: 'a> {
+    error: &'a mut Result<(), E>,
+    iter: I,
+}
+
+impl<'a, I, T, E> Iterator for ProcessResults<'a, I, E>
+    where I: Iterator<Item = Result<T, E>>
+{
+    type Item = T;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        match self.iter.next() {
+            Some(Ok(x)) => Some(x),
+            Some(Err(e)) => {
+                *self.error = Err(e);
+                None
+            }
+            None => None,
+        }
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let (_, hi) = self.iter.size_hint();
+        (0, hi)
+    }
+}
+
+/// “Lift” a function of the values of an iterator so that it can process
+/// an iterator of `Result` values instead.
+///
+/// `iterable` is an iterator or iterable with `Result<T, E>` elements, where
+/// `T` is the value type and `E` the error type.
+///
+/// `processor` is a closure that receives an adapted version of the iterable
+/// as the only argument — the adapted iterator produces elements of type `T`,
+/// as long as the original iterator produces `Ok` values.
+///
+/// If the original iterable produces an error at any point, the adapted
+/// iterator ends and the `process_results` function will return the
+/// error iself.
+///
+/// Otherwise, the return value from the closure is returned wrapped
+/// inside `Ok`.
+///
+/// # Example
+///
+/// ```
+/// use itertools::process_results;
+///
+/// type R = Result<i32, &'static str>;
+///
+/// let first_values: Vec<R> = vec![Ok(1), Ok(0), Ok(3)];
+/// let second_values: Vec<R> = vec![Ok(2), Ok(1), Err("overflow")];
+///
+/// // “Lift” the iterator .max() method to work on the values in Results using process_results
+///
+/// let first_max = process_results(first_values, |iter| iter.max().unwrap_or(0));
+/// let second_max = process_results(second_values, |iter| iter.max().unwrap_or(0));
+///
+/// assert_eq!(first_max, Ok(3));
+/// assert!(second_max.is_err());
+/// ```
+pub fn process_results<I, F, T, E, R>(iterable: I, processor: F) -> Result<R, E>
+    where I: IntoIterator<Item = Result<T, E>>,
+          F: FnOnce(ProcessResults<I::IntoIter, E>) -> R
+{
+    let iter = iterable.into_iter();
+    let mut error = Ok(());
+
+    let result = processor(ProcessResults { error: &mut error, iter: iter });
+
+    error.map(|_| result)
+}


bluss:
I've read it a couple of times and I recognized it as the same as https://github.com/bluss/rust-itertools/issues/166 just a different way to go about it.

Question first: Is this an iterator adaptor? It's got a public iterator impl, but the user can't use that for anything useful. We must either expose `new` or hide the iterator impl.

- If this is not an iterator adaptor then like Raymond Hettinger said at some point, a class with only one method begs to be a function.
- If this is an iterator adaptor, then it should have a free function constructor and the struct hidden away in the attic -- iterator::structs.
  - The motivation is roughly, see `std::iter::once`; iterators don't have an "I'm a value of a <type>"-identity, it's just a function or method that returns something that's an iterator and we don't care what. See also: iterator structs are named after the function that makes them.
- Found some missing IntoIterator conversions. Any time we see `Foo::process(x.into_iter())` we know that process should learn to take iterables instead.
- (If this is an iterator adaptor) Missing size_hint


I'm tempted to propose the first_error approach instead; maybe two functions -- `first_error` that makes an iterator and `process_to_first_error` that does what `ResultShunt::process` does

Sorry for the size of the reply..

shepmaster:
> the same as #166 just a different way to go about it.

Indeed! That one's nice in that it can be used directly like `for line in first_error(...)`. However, it doesn't help the user out if they forget to check the error — passing `&mut` counts as a use even though it's never read from.

>  but the user can't use that for anything useful

I'm afraid I don't fully understand what useful means in this case. 

> We must either expose `new` or hide the iterator impl.

Exposing `new` means that it's possible for the user to forget to check the error. This was brought up during code review for inclusion in the standard library, which is why `process` exists.

I don't think I know how you'd "hide" the iterator impl... is there an example of that I could read to better understand?

> Found some missing `IntoIterator` conversions
>
> Missing `size_hint`

Absolutely; I'll address those as we figure out what the code should look like and where it should live 😸 

> Sorry for the size of the reply..

Not at all!

bluss:
> I don't think I know how you'd "hide" the iterator impl... is there an example of that I could read to better understand?

You're right, it's my confusion. The iterator impl is user visible and user useful when the closure is called by process.

shepmaster:
I'm not any closer to deciding between implementations. It seems like the usage of `first_error` is more permissive — you can handle the `Ok` values *and* you can get the final error. As far as I can tell, that's isomorphic to calling `ResultShunt::reconstruct` with `()`. This API is also more dangerous for the reason mentioned earlier — forgetting to check the error.

In general, it feels like Rust's path is to make this kind of thing the default, which is why I'd want `process` to be the star of the show.

For reference, [here's the usage in the standard library](https://github.com/rust-lang/rust/blob/5c71e4ef90ef79c1ac79c4132333cbc80f5b85b9/src/libcore/iter/traits.rs#L804).

---

I'd guess I'd say that this is *both* an adapter and a regular method, coupled together. There's a function that terminates the iterator but by yielding another iterator (which is an adapter) to the callback. That doesn't really help make it any easier! :-)

bluss:
I'm pondering how to best expose this as an API. I want to use something that matches the crate: it uses free functions and iterator methods.

shepmaster:
What about a terminal function that yields the closure:

```rust
use itertools::Itertools;

let successes_and_failures: Vec<Result<(u8, u8), bool>> =
    vec![Ok((1, 2)), Err(false), Err(true), Ok((3, 4))];

successes_and_failures.into_iter().shunt(|iter| {
    let unzipped: (Vec<_>, Vec<_>) = iter.unzip();
    unzipped
})
```

> it uses free functions and iterator methods.

Unsolicited opinion: I never use the free functions, and have honestly never understood why they exist ;-)

bluss:
There are free functions that have no corresponding method too, like `repeat_n`, `partition`, `unfold`, `multipeek` etc.

The others say [on the tin](https://docs.rs/itertools/0.6.1/itertools/fn.zip.html) :smile: why they exist: because they are `IntoIterator`-enabled. It's a lot neater to *start* an iterator chain with one of them.

bluss:
“thinking” about this apparently turned into implementing, it's the same thing just with a slightly different angle. We can't hide the iterator impl, but we can make it not escape.

Most names are fine, maybe the function can even be named `result_shunt` or similar?

```rust
pub struct ProcessResults<'a, I: 'a, E: 'a> {
    error: &'a mut Result<(), E>,
    iter: &'a mut I,
}

impl<'a, I, T, E> Iterator for ProcessResults<'a, I, E>
    where I: Iterator<Item=Result<T, E>>
{
    type Item = T;

        match self.iter.next() {
            Some(Ok(x)) => Some(x),
            Some(Err(e)) => {
                *self.error = Err(e);
                None
            }
            None => None,
        }
    }

        let (_, hi) = self.iter.size_hint();
        (0, hi)
    }
}

/// “Lift” a function of the values of an iterator so that it can process
/// and iterator of `Result` values instead.
///
/// `iterable` is an iterator or iterable with `Result<T, E>` elements, where
/// `T` is the value type and `E` the error type.
///
/// `processor` is a closure that receives an adapted version of the iterable
/// as the only argument — the adapted iterator produces elements of type `T`,
/// as long as the original iterator produces `Ok` values.
///
/// If the original iterable produces an error at any point, the adapted
/// iterator ends and the `process_results` function will return the
/// error iself.
///
/// Otherwise, the return value from the closure is returned wrapped
/// inside `Ok`.
///
/// # Example
///
/// ```
/// use itertools::process_results;
///
/// type R = Result<i32, &'static str>;
///
/// let first_values: Vec<R> = vec![Ok(1), Ok(0), Ok(3)];
/// let second_values: Vec<R> = vec![Ok(2), Ok(1), Err("overflow")];
///
/// // “Lift” the iterator .max() method to work on the values in Results using process_results
///
/// let first_max = process_results(first_values, |iter| iter.max().unwrap_or(0));
/// let second_max = process_results(second_values, |iter| iter.max().unwrap_or(0));
///
/// assert_eq!(first_max, Ok(3));
/// assert!(second_max.is_err());
/// ```
pub fn process_results<I, F, T, E, R>(iterable: I, processor: F) -> Result<R, E>
    where I: IntoIterator<Item=Result<T, E>>,
          F: FnOnce(ProcessResults<I::IntoIter, E>) -> R
{
    let mut iter = iterable.into_iter();
    let mut error = Ok(());

    let result = processor(ProcessResults { error: &mut error, iter: &mut iter });
    error?;
    Ok(result)
}
```

bluss:
Are you happy if we merge this and then I put some changes on top? I can decide the name if there are no strong opinions.

shepmaster:
I'm very sorry for taking so long to respond! Your proposed change looks good — I updated the PR with that, fixing a small typo along the way. 

I also changed the return value to use `.map` and took ownership of the iterator instead of passing in the mutable reference. Let me know if either of those were for reasons I can't comprehend.

There's a warning left about documentation for `ProcessResults` — did you want that to be documented?

shepmaster:
> the function can even be named result_shunt

I'm not particularly tied to the name, I left it in the commit message to tie it back to the o.g. implementation in rustc.

bluss:
Good changes. No sweating the response time, it might as well be me that's slow next time.

bluss:
Let's merge so that this is finished. I'll put the struct away in the attic we reserve for structs (itertools::structs) and add the doc to it.

bluss:
Cool addition to itertools -- I think this is the right shape for it.

PR #207, opened 2017-07-17
Implement concat method

srijs:
Implements a `concat` method to mirror the [`Stream::concat2`](https://docs.rs/futures/0.1.14/futures/stream/trait.Stream.html#method.concat2) method from the [futures](https://github.com/alexcrichton/futures-rs) library.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -58,6 +58,7 @@ pub mod structs {
         UniqueBy,
         Flatten,
     };
+    pub use concat_impl::concat;
     pub use cons_tuples_impl::ConsTuples;
     pub use format::{Format, FormatWith};
     pub use groupbylazy::{IntoChunks, Chunk, Chunks, GroupBy, Group, Groups};
@@ -92,6 +93,7 @@ mod adaptors;
 pub mod free;
 #[doc(inline)]
 pub use free::*;
+mod concat_impl;
 mod cons_tuples_impl;
 mod diff;
 mod format;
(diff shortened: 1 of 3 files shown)


bluss:
Oh, I expected this to be like .flatten().collect() at first, but I see that it is not.

Puzzling around a bit, it's like this:

```rust
iter.fold1(|mut a, b| { a.extend(b); a }).unwrap_or_else(|| <_>::default())
```

The idea with the puzzle is to find out if it's already covered by itertools and a hint for the best name.

It's a fold with the operation `|mut a, b| { a.extend(b); a }` which is some kind of concatenation operator.

I think it's fine to add this using the concat name, but I would clarify the first line of the method doc to be even more specific.

Minor points:

- This is not an iterator adaptor so it should be listed in the other section; [explanation](https://github.com/srijs/rust-itertools/blob/4031feee38ed345cc1364bd47341eb5a1984f49f/src/lib.rs#L188-L197)
- The free function with the same name should be public, I think it's a good fit. I would probably rename the module so that it doesn't clash with the module
- Rewriting in terms of `fold1` or `fold` has a sense since it reuses `fold`, which is kind of the canonical internal iteration implementation.

srijs:
Cool, agreed on all points! I've updated the PR with your suggestions.

However, I failed to come up with a better first line for the doc, maybe you have an idea?

shepmaster:
I'm also a fan of having the example doctest so you can quickly see what the function can do.

bluss:
Thanks a lot! I'll adjust on top of this. I will have to take responsibility for that first line and if it sucks, your PR :wink:

bluss:
Yes, every method should have an example. I'll do it.

src/lib.rs
```diff
@@ -58,6 +58,7 @@ pub mod structs {
         UniqueBy,
         Flatten,
     };
+    pub use concat_impl::concat;
```
bluss: I was too fast, this should not have been in structs.

PR #212, opened 2017-08-13
Implement izip! using .zip()s and single tupple-flattening .map()

krdln:
This should *probably* enable vectorization-friendly specializations from `std`.

In theory, it's a breaking change, since it no longer returns `multizip::Zip`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -155,8 +155,8 @@ macro_rules! iproduct {
 /// returns `None`.
 ///
 /// Iterator element type is like `(A, B, ..., E)` if formed
-/// from iterators `(I, J, ..., M)` implementing `I: Iterator<A>`,
-/// `J: Iterator<B>`, ..., `M: Iterator<E>`
+/// from iterators `(I, J, ..., M)` implementing `I: IntoIterator<Item=A>`,
+/// `J: IntoIterator<Item=B>`, ..., `M: IntoIterator<Item=E>`
 ///
 /// ```
 /// #[macro_use] extern crate itertools;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -174,14 +174,27 @@ macro_rules! iproduct {
 /// # }
 /// ```
 macro_rules! izip {
-    ($I:expr) => (
-        (::std::iter::IntoIterator::into_iter($I))
-    );
-    ($($I:expr),*) => (
-        {
-            $crate::multizip(($(izip!($I)),*))
-        }
-    );
+    // @closure creates a tuple-flattening closure for .map() call. usage:
+    // @closure partial_pattern => partial_tuple , rest , of , iterators
+    // eg. izip!( @closure ((a, b), c) => (a, b, c) , dd , ee )
+    ( @closure $p:pat => $tup:expr ) => {
+        |$p| $tup
+    };
+
+    // The "b" identifier is a different identifier on each recursion level thanks to hygiene.
+    ( @closure $p:pat => ( $($tup:tt)* ) , $_iter:expr $( , $tail:expr )* ) => {
+        izip!(@closure ($p, b) => ( $($tup)*, b ) $( , $tail )*)
+    };
+
+    ( $first:expr $( , $rest:expr )* $(,)* ) => {
+        ::std::iter::IntoIterator::into_iter($first)
+            $(
+                .zip($rest)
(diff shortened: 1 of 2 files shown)


bluss:
Thanks!

src/lib.rs
```diff
@@ -174,14 +174,31 @@ macro_rules! iproduct {
...
+        |$p| $tup
+    };
+
+    // The "b" identifier is a different identifier on each recursion level thanks to hygiene.
+    ( @closure $p:pat => ( $($tup:tt)* ) , $_iter:expr $( , $tail:expr )* ) => {
+        izip!(@closure ($p, b) => ( $($tup)*, b ) $( , $tail )*)
+    };
+
+    ( $first:expr $( , $rest:expr )* $(,)* ) => {
+        IntoIterator::into_iter($first)
```
bluss: Needs to use absolute path for `IntoIterator`. In fact the original `::std` path is not perfect, but we can at least use that.

src/lib.rs
```diff
@@ -174,14 +174,31 @@ macro_rules! iproduct {
...
-        (::std::iter::IntoIterator::into_iter($I))
-    );
-    ($($I:expr),*) => (
-        {
-            $crate::multizip(($(izip!($I)),*))
-        }
-    );
+    () => {
+        ::std::iter::empty::<()>()
+    };
```
bluss: I'd prefer to not add this as a new feature. This iterator is better if we syntactically ensure there is at least one iterator.

krdln: You're right – there's no `impl_zip_iter!()`. I somehow thought this case worked before. (also it should be `repeat` rather than `empty`)

bluss:
This should resolve the performance trouble with `izip`... Maybe it should be tested before it's merged though! itertools itself keeps some zip benchmarks on file (but they are kind of imperfect; I'm generally skeptical of leaning hard on the `black_box` function -- it blocks optimizations that are important as well.)

We still need to keep `multizip` and `Zip` to provide a named type for a multi zip, and to avoid too much breaking changes. I can update docs for them both to refer to each other and the details.

krdln:
@bluss
> itertools itself keeps some zip benchmarks on file

What does "on file" mean in this context? Haven't found anything for `izip` or `multizip` in the repo.

<details><summary>I've done quick test using the following bench:</summary>

```rust
#![feature(test)]

#[macro_use] extern crate itertools;
extern crate test;

use test::Bencher;
use test::black_box;

#[bench]
fn fmul_izip(b: &mut Bencher) {
    let aa: &[f32] = &[1.0; 1000];
    let bb: &[f32] = &[2.0; 1000];
    let cc: &mut [f32] = &mut[0.0; 1000];
    b.iter(|| {
        let aa = black_box(aa);
        let bb = black_box(bb);
        let cc = black_box(&mut *cc);
        for (a, b, c) in izip!(aa, bb, cc) {
            *c = *a * *b;
        }
    })
}

#[bench]
fn fmul_multizip(b: &mut Bencher) {
    let aa: &[f32] = &[1.0; 1000];
    let bb: &[f32] = &[2.0; 1000];
    let cc: &mut [f32] = &mut[0.0; 1000];
    b.iter(|| {
        let aa = black_box(aa);
        let bb = black_box(bb);
        let cc = black_box(&mut *cc);
        for (a, b, c) in itertools::multizip((aa, bb, cc)) {
            *c = *a * *b;
        }
    })
}
```
</details>

The results seem promising:
```
test fmul_izip     ... bench:          62 ns/iter (+/- 8)
test fmul_multizip ... bench:         798 ns/iter (+/- 49)
```

> I can update docs for them both to refer to each other and the details.

So I should leave the docs as they are in this PR, right?

bluss:
There's a heap of zip benchmarks in the benches/bench1.rs file. Not izip!(), but for other zips. Not editing docs more is fine.

bluss:
We will have to treat this as a breaking change, we have to be careful with a lot of dependents.  We're going to add the `#![no_std]` breaker at the same time (which is much more of a theoretical case, about those using `--no-default-features` for a crate with no crate features...).

bluss:
I'm sorry for taking the time here. Making a breaking update requires some more work, so I'll do the other stuff first.

bluss:
Let's go! Thanks for writing this macro :-)

PR #214, opened 2017-08-24
Remove unneeded mut qualifiers

shepmaster:
I feel like I'm missing something obvious here, but I ran `cargo build` `cargo test` `cargo doc` on 1.19 and 1.21.0-nightly (f25c2283b 2017-08-15) and it worked in all those cases...

How are these unneeded?


bluss:
Confusing! It's because the variable is already a `&mut I`. Rust must have just learned to warn for these.

PR #216, opened 2017-09-01
Add fold_mutref

cramertj:
I've wanted this function a couple times in the past, so i thought I'd suggest adding it here. Feel free to bikeshed the name-- I also considered `fold_mut`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1308,6 +1308,34 @@ pub trait Itertools : Iterator {
         Ok(start)
     }
 
+    /// Fold using a mutable reference to the accumulator.
+    ///
+    /// Rather than having to mutate the accumulator and then return it,
+    /// this function allows simple, direct mutations.
+    ///
+    /// ```
+    /// use itertools::Itertools;
+    ///
+    /// #[derive(Debug, PartialEq)]
+    /// struct State(i32);
+    ///
+    /// impl State {
+    ///     fn mutate(&mut self, x: i32) {
+    ///         self.0 += x;
+    ///     }
+    /// }
+    ///
+    /// let s = (0..10).fold_mut(State(0), |s_mut, x| s_mut.mutate(x));
+    /// assert_eq!(s, State(45));
+    /// ```
+    fn fold_mut<ACC, F>(self, mut acc: ACC, mut f: F) -> ACC
+        where F: FnMut(&mut ACC, Self::Item),
+              Self: Sized
+    {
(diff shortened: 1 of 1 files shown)


bluss:
Warning, I'm thinking aloud.. It's not bad:

```rust
let element_freq = iterable.fold_mut(HashMap::new(), |map, c| *map.entry(c).or_insert(0) += 1);
```

In a way it seems redundant with `for_each` (finally it's coming!). We have closure environments, so we don't really need to pass map, the reason for that is mostly the expression oriented way we want to code. And the expression oriented way is most natural with the existing fold, the by value version.

(Examples for [unfold 0.6.0](https://docs.rs/itertools/0.6.0/itertools/fn.unfold.html) and [unfold 0.6.2](https://docs.rs/itertools/0.6.2/itertools/fn.unfold.html) stabs at trying both of those.. the way it's expressed with closure captures is far superior for this case.)

cramertj:
It is kind of redundant with `for_each`, agreed.

cramertj:
Note that some of the nastiness comes from having to use `*` and `ref mut` to unwrap the multiple bits of state. That'll be a lot cleaner once match-ergo lands.

src/lib.rs
```diff
@@ -1308,6 +1308,35 @@ pub trait Itertools : Iterator {
...
+    /// let s = (0..10).fold_mutref(State(0), |s_mut, x| s_mut.mutate(x));
+    /// assert_eq!(s, State(45));
+    /// ```
+    fn fold_mutref<ACC, F>(&mut self, start: ACC, mut f: F) -> ACC
+        where F: FnMut(&mut ACC, Self::Item)
+    {
+        self.fold(start, |mut acc, item| {
+            f(&mut acc, item);
+            acc
+        })
```
bluss: I think a `foreach` (or `fold((), |(), elt| ..)`) based implementation is best, it's truest to the method that we don't move around the acc value at all, if that for some reason is a big inconvenience or does not optimize out.

bluss:
it's a bit similar to this method as well: [rayon's `for_each_with`](https://docs.rs/rayon/0.8.2/rayon/iter/trait.ParallelIterator.html#method.for_each_with), but there is nothing quite like it in rayon.

bluss:
Ok, sorry for letting this wait for a bit. I think it seems this offers too little in terms of a new feature, that we don't want to include it. Let me know if you have other use cases or users that change the perspective on that.

cramertj:
Seems fine to me!

PR #219, opened 2017-09-16
Add positions adaptor

### LINKED ISSUE #218: Iterator adaptor .positions(predicate)
Like an iterated version of position or a filtered version of enumerate.

The existing combination is a bit heavy on the punctuation:

```rust
self.1.iter()
      .enumerate()
      .filter(|&(_, node)| node.weight.is_none())
      .map(|(i, _)| NodeIndex::<Ix>::new(i))
```

Propose instead:

```rust
self.1.iter()
      .positions(|node| node.weight.is_none())
      .map(NodeIndex::<Ix>::new)
```

tmccombs:
To get iterators of positions of elements that match a predicate.

Fixes #218

### DIFF
--- a/src/adaptors/mod.rs
+++ b/src/adaptors/mod.rs
@@ -1502,7 +1502,6 @@ impl<I, F> Iterator for RPositions<I, F>
                 return Some(self.count);
             }
         }
-        assert_eq!(self.count, 0);
         None
     }
 
--- a/src/adaptors/mod.rs
+++ b/src/adaptors/mod.rs
@@ -1423,3 +1423,90 @@ impl<I, F, T, U, E> Iterator for MapResults<I, F>
         self.iter.size_hint()
     }
 }
+
+/// An iterator adapter to get the positions of each element that matches a predicate.
+///
+/// See [`.positions()`](../trait.Itertools.html#method.positions) for more information.
+pub struct Positions<I, F> {
+    iter: I,
+    f: F,
+    count: usize
+}
+
+/// Create a new `Positions` iterator.
+pub fn positions<I, F>(iter: I, f: F) -> Positions<I, F>
+    where I: Iterator,
+          F: FnMut(I::Item) -> bool,
+{
+    Positions {
+        iter: iter,
+        f: f,
+        count: 0
+    }
+}
+
+impl<I, F, T> Iterator for Positions<I, F>
+    where I: Iterator<Item = T>,
+          F: FnMut(I::Item) -> bool,
+{
+    type Item = usize;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        while let Some(v) = self.iter.next() {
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -1010,6 +1012,42 @@ pub trait Itertools : Iterator {
         None
     }
 
+    /// Find the positions of each element satisfying a predicate.
+    ///
+    /// Equivalent to `iter.filter(|(_,v)| pred(v)).map(|(i,_)| i)`.
```
bluss: missing enumerate in the equivalency.

I'd make the doc even more clear by saying something like "yields the index of each element matching the predicate, counted from the start of the iterator".

bluss: The thought is that we need to remind the user somehow that we count indices as we see them in the iterator.  If they give us `slice[3..].iter().positions()` we are not counting indices in the original slice etc.

src/adaptors/mod.rs
```diff
@@ -1423,3 +1423,90 @@ impl<I, F, T, U, E> Iterator for MapResults<I, F>
...
+    type Item = usize;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        while let Some(v) = self.iter.next_back() {
+            self.count -= 1;
+            if (self.f)(v) {
+                return Some(self.count);
+            }
+        }
+        assert_eq!(self.count, 0);
```
bluss: Oh. I don't think we should leave an assertion in here, it's too heavy handed. Won't it destroy a lot of loop optimizations?

bluss:
This iterator should be double ended, shouldn't it?

tmccombs:
I suppose it could implement `DoubleEndedIterator` if the source iterator is `DoubleEndedIterator + ExactSizeIterator`, if I added another counter to `Positions`. In that case `rpositions` could just be `self.positions().rev()`. What do you tink?

bluss:
That sounds good to me -- even removing rpositions seems fine (it matches enumerate that way)

bluss:
Also put the positions adaptor in the correct section of the itertools trait, it belongs with the adaptors.

> The trait Itertools: extra iterator adaptors and methods for iterators.

> This trait defines a number of methods. They are divided into two groups:

>     Adaptors take an iterator and parameter as input, and return a new iterator value. These are listed first in the trait. An example of an adaptor is .interleave()

>    Regular methods are those that don't return iterators and instead return a regular value of some other kind. .next_tuple() is an example and the first regular method in the list.

bluss:
oh another counter? We don't need that, we can use len like enumerate does.

tmccombs:
Oh, I didn't think about the fact that `len()` changes when next or next_back is called. Good call.

bluss:
Thanks! Will make a release with this soon.

bluss:
In itertools 0.6.3

PR #225, opened 2017-09-28
Add PartialEq trait bounds to IterTools::group_by

### LINKED ISSUE #224: Group_by trait bounds
In the Itertools trait, group_by has the following definition:
```
fn group_by<K, F>(self, key: F) -> GroupBy<K, Self, F>
  where
    Self: Sized,
    F: FnMut(&Self::Item) -> K
```
It then simply creates a new GroupBy struct. The only method on the GroupBy struct is into_iter, which has the following definition for the GroupBy struct:
```
impl<'a, K, I, F> IntoIterator for &'a GroupBy<K, I, F> 
  where
    I: Iterator,
    I::Item: 'a,
    F: FnMut(&I::Item) -> K,
    K: PartialEq, 
```

So in order to _use_ group_by, K must implement PartialEq; but that fact is not evident in the definition of the group_by function (No trait bounds on K).

Should it be added to group_by?

I ran into this issue when K was a struct; I had to jump through the documentation to understand why K didn't typecheck, and where rustc was getting the trait bound.

Phlosioneer:
Closes #224

Merge this pull request whenever the next round of breaking changes occur.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -371,6 +371,7 @@ pub trait Itertools : Iterator {
     fn group_by<K, F>(self, key: F) -> GroupBy<K, Self, F>
         where Self: Sized,
               F: FnMut(&Self::Item) -> K,
+	      K: PartialEq,
     {
         groupbylazy::new(self, key)
     }


bluss:
Thanks! Hopefully relatively soon. #212 seems important.

Phlosioneer:
The check for rust nightly failed, but I don't _think_ that was my change? Should I investigate that further, or was it already failing on nightly?

The specific test that failed was `correct_cartesian_product3`.

bluss:
Yeah looks like the new cartesian product code might have a bug related to fusing, I'll look at that.

Issue #143, opened 2016-09-07
Add #[must_use] where appropriate


bluss:
Add #[must_use] just like libstd.

PR #227, opened 2017-10-12
Add #[must_use] to adaptor returned by methods on Itertools trait

matematikaadit:
Related to #143. The purpose is to warn user that iterator adaptors are lazy and so it wont execute unless consumed.


bluss:
Thanks! Much appreciated

PR #228, opened 2017-10-14
Add new adaptor .merge_join_by()

srijs:
Noticed I was re-implementing something similar to this combinator in many projects, so I thought it might be worth making it reusable!

The combinator takes two iterators and merges/zips their items in ascending order. It's useful  in many contexts to implement merge behaviours with linear time complexity, including but not limited to:

- Determining the causal order of two vector clocks
- Merging timestamp-ordered streams of events, cross-referencing events with the same timestamp
- Zipping two heterogenous maps or trees (think merging `Map<K, A>` and `Map<K, B>` into `Map<K, (A, B)>`)

Let me know what you think!

### DIFF
--- a/src/either_or_both.rs
+++ b/src/either_or_both.rs
@@ -1,0 +1,10 @@
+/// Value that either holds a single A or B, or both.
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub enum EitherOrBoth<A, B> {
+    /// Both values are present.
+    Both(A, B),
+    /// Only the left value of type `A` is present.
+    Left(A),
+    /// Only the right value of type `B` is present.
+    Right(B),
+}
--- a/src/free.rs
+++ b/src/free.rs
@@ -21,6 +21,7 @@ pub use multipeek_impl::multipeek;
 #[cfg(feature = "use_std")]
 pub use kmerge_impl::kmerge;
 pub use zip_eq_impl::zip_eq;
+pub use merge_join::merge_join_by;
 #[cfg(feature = "use_std")]
 pub use rciter_impl::rciter;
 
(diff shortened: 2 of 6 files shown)


bluss:
Hey, this looks interesting. It's kind of related to this here right, without the explicit "joining" of keys? https://github.com/contain-rs/ordered_iter

Brief look says it seems like a promising shape of that kind of algorithm to bring into itertools. It seems just as much like a variant of "merge" as of "zip", but naming is hard..

srijs:
You're right, from a set theory standpoint this implements an "outer join", similar to what is implemented in `ordered_iter` as `OuterJoin`.

I did some light reading, and this algorithm is also called a ["merge join"](https://en.wikipedia.org/wiki/Sort-merge_join).

I've been struggling with the name as well, in a way this is a cross between `.merge()` and `.zip_longest()`. We could rename it to  `.merge_join()` to be in-line with the name for the algorithm linked above.

Let me know what you think!

bluss:
I think merge_join_by sounds good too. I prefer it to zip, since we are most often not making pairs here, but if you have another better name feel free to pick that too. It seems fine to merge as soon as we have the name fixed.

srijs:
@bluss renamed to `merge_join_by`.

bluss:
Thanks! Ill look at this when im back home. I don't want to expand the either crate by anything. But the topic of another rate for itertools' types has come up before.

bluss:
For future additions I recommend going directly to quickcheck tests -- they are both shorter and test more, I quite like them.

bluss:
And of course, thanks a lot for this addition!

bluss:
I'm fixing up some issues myself right now.. quicker that way, I think I didn't have time to look at the details  before :/ besides going back-and-forth is no fun. The ideas and name were the important thing here. :-)

srijs:
Please feel free to criticise future PRs in more detail, I'm always eager to learn and implement suggestions!

Thanks for merging!

bluss:
I'll do it better when I have time :smile: Sometimes it just doesn't happen

PR #232, opened 2017-11-06
FEAT: implement .update()

### LINKED ISSUE #231: [feature request] shuffle
Consider the following trait method:

```rust
pub trait Itertools: Iterator {
    fn shuffle(self, buffer_size: usize) -> Shuffle<Self> {}
}
```

The `Shuffle` iterator would create a buffer (`Vec::with_capacity(buffer_size)`) and have two states:
- empty buffer: on the next `next()` call, the buffer is filled with items from the underlying iterator and the buffer is shuffled. Fallthrough to non-empty buffer case.
- non-empty buffer: items are popped from the buffer.

I'd happily implement this if people like the idea.

lucasem:
Closes #231.

Adds the following method on Itertools:

```rust
pub trait Itertools : Iterator {
    { ... }
    fn shuffle<F>(self, buffer_size: usize, shuffler: F) -> Shuffle<Self, F> 
    where
        Self: Sized,
        Self::Item: PartialEq,
        F: FnMut(&mut Vec<Self::Item>), 
    { ... }
}
```

Shuffles portions of the iterator by collecting items into a buffer of a given size and calling a shuffling function on that buffer.

```rust
use itertools::Itertools;

let data = vec![0, 1, 2, 3, 4, 5, 6];
let shuffled = data.into_iter().shuffle(2, |v| v.reverse());
itertools::assert_equal(shuffled,
                        vec![1, 0, 3, 2, 5, 4, 6]);
```

You can use the rand crate to randomly shuffle the buffer:

```rust
extern crate rand;

use itertools::Itertools;
use rand::Rng;

let shuffle = |v: &mut Vec<_>| rand::thread_rng().shuffle(v.as_mut_slice());
let data = vec![0, 1, 4, 7, 2, 9, 5];
let shuffled: Vec<_> = data.iter().shuffle(8, shuffle).collect();
// e.g. -> [4, 5, 0, 7, 9, 1, 2]
```

### DIFF
--- a/src/adaptors/mod.rs
+++ b/src/adaptors/mod.rs
@@ -1192,3 +1192,62 @@ impl<I, F> DoubleEndedIterator for Positions<I, F>
     }
 }
 
+/// An iterator adapter to apply a mutating function to each element before yielding it.
+///
+/// See [`.update()`](../trait.Itertools.html#method.update) for more information.
+#[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
+pub struct Update<I, F> {
+    iter: I,
+    f: F,
+}
+
+/// Create a new `Update` iterator.
+pub fn update<I, F>(iter: I, f: F) -> Update<I, F>
+where
+    I: Iterator,
+    F: FnMut(&mut I::Item),
+{
+    Update { iter: iter, f: f }
+}
+
+impl<I, F> Iterator for Update<I, F>
+where
+    I: Iterator,
+    F: FnMut(&mut I::Item),
+{
+    type Item = I::Item;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if let Some(mut v) = self.iter.next() {
+            (self.f)(&mut v);
+            Some(v)
+        } else {
+            None
+        }
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        self.iter.size_hint()
+    }
+}
+
+impl<I, F> ExactSizeIterator for Update<I, F>
+where
+    I: ExactSizeIterator,
+    F: FnMut(&mut I::Item),
+{}
+
+impl<I, F> DoubleEndedIterator for Update<I, F>
+where
+    I: DoubleEndedIterator,
+    F: FnMut(&mut I::Item),
+{
+    fn next_back(&mut self) -> Option<Self::Item> {
+        if let Some(mut v) = self.iter.next_back() {
+            (self.f)(&mut v);
+            Some(v)
+        } else {
+            None
+        }
+    }
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -65,6 +65,7 @@ pub mod structs {
         TupleCombinations,
         Flatten,
         Positions,
+        Update,
     };
     #[cfg(feature = "use_std")]
     pub use combinations::Combinations;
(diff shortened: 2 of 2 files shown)


bluss:
Hi! To fit in itertools, I think we need to consider the most general form of this functionality. Can you reason about what that is and what the more generic operation of this adaptor does?

It is a benefit if the adaptor is generally applicable to different problems, otherwise I'm not sure if it fits. Thanks!

lucasem:
Of course! Its primary use is to randomize an iterator. I think this is generally applicable to plenty of problems.

For example, I could create a shuffled deck of playing cards (ignoring suit):
```rust
extern crate rand;
use itertools::Itertools;
use rand::Rng;
// random shuffle
let shuffle = |v: &mut Vec<_>| rand::thread_rng().shuffle(v.as_mut_slice());

let shuffled_deck = (0..13).cycle().take(52).shuffle(52, shuffle).collect();
```

Or to pick a random subset of size `k`:
```rust
let xs = [7, 1, 2, 3, 6, 0, 2, 6];
let k = 3;
let subset = xs.iter().shuffle(8, shuffle).take(k).collect();
```

Or many Monte Carlo algorithms.

Those are simple use cases, but the prevalence of a random shuffle in many standard libraries indicates the utility of this kind of feature.

bluss:
I was thinking of something like this:

- The method is not specific to shuffling or using rand at all, instead it's a quite general operation. That's normally good. What's the general operation? What more can it be used for?

lucasem:
That's a good question. I think it would be _ideal_ for shuffle to only take in a buffer size, but there are some important pros and cons to doing the randomization in this crate:

* Pros (_reasons for shuffle to only take buffer size as arg_):
  - Easier developer experience/prettier interface.
* Cons (_reasons for shuffle to also take shuffler function as arg_):
  - Would have to (optionally, like with std) depend on the rand crate — I'm not sure if you're willing to go for that kind of footprint.
  - Wouldn't account for different kinds of randomization — fast/simple vs cryptographically secure, etc.

It's not an obvious decision — it depends primarily on the kind of level of abstraction you want to provide itertools users. I am fine with either one, and perhaps partial to only taking in a buffer size (putting randomization in this crate).

bluss:
Yes, rand is out, we want to stick to general purpose iterator extensions without bringing in extra utilities. If you are interested in rand extensions -- I'd encourage you to create itertools-rand or such crates.

I would like to "refactor" the thinking behind the feature in this PR — I know it doesn't land very conveniently, but it seems to me, that this is a combination of chunking, collecting a vector of each chunk, mapping each such vector in place (the random shuffle) and then flattening the iterator again.

We have all those parts, but each could be expressed better:

- `theiterator.chunks(n).into_iter()`
  - with its `into_iter()` drawback
  - also possible to have a shorthand for chunks as vectors in particular
- `.map(|chunk| chunk.collect_vec())`
  - Let's hope for a std `.into_vec()` in the future as the same shorthand
- `.map(|mut chunk| {rand_shuffle(&mut chunk); chunk})`
  - Missing a "map inplace" that passes a &mut to the mapping function?
- `.flatten()`

What we have:

```rust
iter.chunks(n).into_iter()
    .map(|chunk| chunk.collect_vec())
    .map(|mut chunk| { rand_shuffle(&mut chunk); chunk})
    .flatten()
```

So what would the ideal be? Something like this seems good enough, composed of orthogonal parts:  

```rust
iter.chunks(n)
    .map(Iterator::into_vec)
    .map_inplace(rand_shuffle)
    .flatten()
```


(Other proposed names for `map_inplace` have been `update`)

lucasem:
That's a great way of looking at it. The sacrifice to convenience and brevity is sensible. I'll update the PR with something like this:

```rust
pub trait Itertools : Iterator {
    { ... }
    fn update<F>(self, updater: F) -> Update<Self, F> 
    where
        Self: Sized,
        F: FnMut(&mut Self::Item), 
    { ... }
}
```

src/adaptors/mod.rs
```diff
@@ -1192,3 +1192,56 @@ impl<I, F> DoubleEndedIterator for Positions<I, F>
...
+        if let Some(mut v) = self.iter.next() {
+            (self.f)(&mut v);
+            Some(v)
+        } else {
+            None
+        }
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        (0, self.iter.size_hint().1)
```
bluss: Cool. Can you quickly fix this to just pass through the size hint? Since we don't filter anything. This also means it can be (conditional on I) ExactSizeIterator

lucasem: 👍

bluss:
Thanks a lot!

PR #233, opened 2017-11-06
Add collect_tuple method

matklad:
Hi! This adds a consuming method `unpack_tuple`, which destructs the iterator into a tuple of specific size, making sure there are no extra items. This is useful, for example, for parsing related code, when you want to coerce unstructured data into a specific form, and mirrors Python's unpacking idiom:

```
(a, b) = input().split()  # makes sure we've got exactly two words
```

You sort-of can use `.next_tuple()` for this, but you may leave some extra input, which is often a cause of subtle bugs.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1132,6 +1132,37 @@ pub trait Itertools : Iterator {
         T::collect_from_iter_no_buf(self)
     }
 
+    /// Collects all items from the iterator into a tuple of
+    /// a specific size (up to 4).
+    ///
+    /// If number of elements inside iterator is exactly equal to the
+    /// tuple size, then the tuple is returned inside `Some`, otherwise
+    /// `None` is returned.
+    ///
+    /// ```
+    /// use itertools::Itertools;
+    ///
+    /// let iter = 1..3;
+    ///
+    /// if let Some((x, y)) = iter.collect_tuple() {
+    ///     assert_eq!((x, y), (1, 2))
+    /// } else {
+    ///     panic!("Expected two elements")
+    /// }
+    /// ```
+    fn collect_tuple<T>(mut self) -> Option<T>
+        where Self: Sized + Iterator<Item = T::Item>,
+              T: tuple_impl::TupleCollect
+    {
+        match self.next_tuple() {
+            Some(tuple) => match self.next() {
+                Some(_) => None,
+                None => Some(tuple),
+            },
+            _ => None
+        }
+    }
+
 
     /// Find the position and value of the first element satisfying a predicate.
     ///
--- a/tests/quick.rs
+++ b/tests/quick.rs
@@ -101,7 +101,7 @@ impl qc::Arbitrary for Inexact {
 /// Our base iterator that we can impl Arbitrary for
 ///
 /// By default we'll return inexact bounds estimates for size_hint
-/// to make tests harder ro pass.
+/// to make tests harder to pass.
 ///
 /// NOTE: Iter is tricky and is not fused, to help catch bugs.
 /// At the end it will return None once, then return Some(0),
@@ -537,6 +537,11 @@ quickcheck! {
             }
         }
         itertools::equal(cloned(&a).tuple_combinations::<(_, _)>(), cloned(&v))
+    }
+
+    fn collect_tuple_matches_size(a: Iter<i16>) -> bool {
+        let size = a.clone().count();
+        a.collect_tuple::<(_, _, _)>().is_some() == (size == 3)
     }
 }
 
--- a/tests/tuples.rs
+++ b/tests/tuples.rs
@@ -71,3 +71,18 @@ fn next_tuple() {
     assert_eq!(iter.next_tuple().map(|(&x, &y)| (x, y)), Some((3, 4)));
     assert_eq!(iter.next_tuple::<(_, _)>(), None);
 }
+
+#[test]
+fn collect_tuple() {
+    let v = [1, 2];
+    let iter = v.iter().cloned();
+    assert_eq!(iter.collect_tuple(), Some((1, 2)));
+
+    let v = [1];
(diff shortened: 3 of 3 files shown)


bluss:
Thanks. It makes sense, it's a common thing one wants to do.

1. This looks almost like a `T::try_from(iter)` for `T` a tuple, right? I wonder if the `.into_tuple()` name fits, but ideally it should have a fallible name, but `.try_into_tuple()` doesn't sound good.
2. Should it be fully fallible and produce a Result instead? There's no precedent for that in itertools, so probably not?
3. Even returning an option is not especially sweet. I imagine one wants to write short and sweet code, but then we gloss over error details. Imagine we use the name and signature `.unwrap_tuple(self) -> T` and too few or many elements is a panic.

Maybe (3) is even the best, otherwise it doesn't really make the code easy to read. What do you think?

```rust
use itertools::Itertools;

let s = "ABC=1";
let (front, tail) = s.split("=").next_tuple().unwrap(); // silent if there are more
let (front, tail) = s.split("=").unpack_tuple().unwrap();
let (front, tail) = s.split("=").into_tuple().unwrap();
let (front, tail) = s.split("=").unwrap_tuple();
```

matklad:
1)

Yeah, I'm open to name bikeshedding here: I think I chose unpack tuple just because it's called "tuple unpacking" in python :) This also looks like `.collect::<TUPLE>()`, so another option is `collect_tuple`. 

So, I think I like `collect_tuple` > `into_tuple` > `unpack_tuple`. I prefer `collect` more because extension from `.collect::<Vec<_>>()` seems the most natural. I don't think we should add `try` here, because `next_tuple` and `next` are fine without one :) 

3) I feel pretty strongly about not implicitly panicking here :) My [original use case](https://github.com/matklad/fall/blob/f605043b2ae9dc6e7b665869605c0344e247f51c/lang/fall/src/analysis/query/resolve_pratt_variant.rs#L57-L68) needs some special error-reporting for the `None` case. Part of the motivation here is correctness (not ignoring trailing values), and unwrapping does not really mesh up with correctness. We *could* add `unwrap_tuple()` as a convenience method, but then, it'll be ambiguous as to whether it is a `.next_tuple().unwrap()` or `.unpack_tuple().unwrap()`.


2) Ah, that's an interesting question! I think a sensible extension is to return the number of elements which were in the iterator. Then, for my use case I would be able to produce a nicer error message. But this will be different from `next_tuple`, different from the rest of the itertools, and the design of the error type seems like a can of worms. 

So, 

1) Let's use `collect_tuple`? If not, let's use `into_tuple`

2) Current signature seems best.

3) The fact that `.unpack_tuple` is ambiguous seems like a nail in the coffin for the idea.

bluss:
Thanks for showing the use case, that is valuable. I think finding the right name is important. It's not bike shedding, it's about visiting a lot of options and seeing if any of them feel really right. It's easy to avoid bike shedding when there's one person making the call, so we'll be fine.

matklad:
> I think finding the right name is important. It's not bike shedding, it's about visiting a lot of options and seeing if any of them feel really right

Totally agree with you, good naming is always important, and for itertools specifically it is crucial. I've just used "bikeshedding" word wrong here :( 

So far, I really like `collect_tuple`: it unambiguously describes the action. It's a bit long though, and does not imply that the operation is failable (but there's no infallible alternative). 

Another option is to try to invent better word than `tuple`. For `next_tuple`, a sweat alternative would be `next_n` (I can imagine it being called like `next_n::<5>` when integer generics arrive). However, `collect_n` does not sound as good, and will be inconsistent with `next_tuple`.

I'll switch PR to `collect_tuple` for now, and try to think about an even better name.

bluss:
I'm fine with either unpack_tuple or collect_tuple I think. This PR looks good to me, thanks for including tests. I'm unsure about the fusing behaviour; basically to be correct we should not call `self.next()` after we have already seen a `None`!

Digression: We could use quickcheck tests (instead) and they uncover many fusing bugs.

I think it's best here to change the match on a pair to be two nested `if let Some(_)`, so that we don't call .next() after the iterator has finished. Do you see it the same way?

matklad:
>I'm unsure about the fusing behaviour; basically to be correct we should not call self.next() after we have already seen a None!

An interesting catch! I've never paid attention to the fusing behavior of iterators before!

I believe it would be hard to observe the difference with the current implementation: its functional behavior is correct even for non-fused iterators, and the iterator is consumed, so it had to also mutate some "global" state for the caller of `collect_tuple` to detect the difference. So I failed to devise a nice property to detect extra `.next`.  

But it certainly makes sense to update the implementation (I've used nested `matches` instead of `if let`s, because `match` is shorter than `if let else`), and I've added a simple property of "next tuple works iff iterator has correct count". 

That's all that needs to be done I think?

matklad:
oups, missed `.` in docs!

matklad:
Ok, now it should be ok :)

bluss:
Side note about consuming iterators, since `&mut I` implements `Iterator` (given bounds..), one can always use `(&mut iter).collect_tuple()` to observe each next and keep the remainder. (One interesting effect is that the `self` methods of the trait, for example `fold` can't use the `I`'s own iterator implementation for fold, it has to use the `&mut I` implementation, so the default, and a lot of special case optimizations might go lost..)

Thanks a lot for the nice PR.

matklad:
Random observation: now that slice patterns are stale and you can unpack fixed-size arrays, argubly the tuple-based APIs should be reformulated in terms of arrays. Not saying it is a good idea, but it defenitelly feels more philosophically appropriate :)

bluss:
Agreed. No quirks or remaining bugs in those? Just curious :)

matklad:
>No quirks or remaining bugs in those?

I don't know: I've used them a couple of times, and they worked for simple cases.

PR #241, opened 2017-12-04
FEAT: implement .all_unique()

samueltardieu:
In day 4 of the Advent of Code 2017 (https://adventofcode.com/),
`all_unique()` would have been useful. It is already present
in some languages such as Factor's `all-unique?`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -35,6 +35,8 @@ use std::iter::{IntoIterator};
 use std::cmp::Ordering;
 use std::fmt;
 #[cfg(feature = "use_std")]
+use std::collections::HashSet;
+#[cfg(feature = "use_std")]
 use std::hash::Hash;
 #[cfg(feature = "use_std")]
 use std::fmt::Write;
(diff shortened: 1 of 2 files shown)


bluss:
Hi, I have to say I have had problem seeing the value in such a method, but I have also had trouble coming up with examples for how to do this easily with existing iterator and itertools functionality.

If we have a collection as starting point we can do the following which many users should be able to find:

```rust
let all_unique = v.iter().unique().count() == v.len();
```

(So I realized `.unique().count()` could be more efficient, and #254 does that.)

I was wondering if we should instead add methods to the `unique()` and `unique_by()` iterators, but I'm not sure we end up with the right semantics, or what we do if they are called when the `unique()` is already partially iterated.

Something like this:

```rust
let all_unique = collection.iter().unique().all_unique();
```

or even a method that tells us about both the numbers. The fidelity loss is that the `.counts()` method can not tell us about the dups that were seen before the `.unique_count()` method was called (if the `.unique()` was already partially iterated).

```rust
let (count_unique, count_dups) = collection.iter().unique().unique_count();
```

jswrenn (approved):
Sorry for the delay!

bors r+

PR #245, opened 2017-12-08
Add "Null", a type to discard FromIterator results without allocating.

birkenfeld:
I've found this useful a few times when making the iterator yield zero-sized items was not easily possible.

Not sure about the name...

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -31,7 +31,7 @@ extern crate core as std;
 
 pub use either::Either;
 
-use std::iter::{IntoIterator};
+use std::iter::{FromIterator, IntoIterator};
 use std::cmp::Ordering;
 use std::fmt;
 #[cfg(feature = "use_std")]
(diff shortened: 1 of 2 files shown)


bluss:
- If not the name null, maybe `Sink` would be good
- I think the `iter.collect::<it::Null>()` idea is a no-go. There is no value in itself in making every iterator functionality go through `collect`, so there has to be a more solid motivation for why this is useful.

jswrenn:
@birkenfeld, can you provide some motivating examples? In the meantime, I'm going to close this PR.

[`Iterator::for_each`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.for_each) covers the use-cases I can imagine.

birkenfeld:
I think my use case was a generic situation, where an iterator-collecting function was to be used in different ways, one of which only made use of the side effects, and discarded the results.

PR #250, opened 2017-12-21
Implement PeekingNext for MultiPeek

nicopap:
I am writing a parser and I recently changed the main iterator from the standard `Peekable` into a `MultiPeek`.

I was surprised I couldn't use it with `peeking_take_while`: I expected that `MutliPeek` -- being an "upgrade" over `Peekable` -- would work just as fine.

The change is very simple. I wrote the `impl` in `mutlipeek_impl.rs` because I use internals of `MultiPeek` to always check for the "true" next value that `iter.next()` will yield.

I added two assertions in the `mutlipeek` test. If this is unwelcome, I'll happily submit a new pull request without those.

### DIFF
--- a/src/multipeek_impl.rs
+++ b/src/multipeek_impl.rs
@@ -3,6 +3,7 @@
 use std::iter::Fuse;
 use std::collections::VecDeque;
 use size_hint;
+use PeekingNext;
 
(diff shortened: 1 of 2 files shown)


bluss:
Thanks, that looks fine.

Issue #238, opened 2017-11-18
Return iterator from sorted / sorted_by


dtolnay:
The [`sorted`](https://docs.rs/itertools/0.7.2/itertools/trait.Itertools.html#method.sorted) and [`sorted_by`](https://docs.rs/itertools/0.7.2/itertools/trait.Itertools.html#method.sorted_by) methods currently return a `Vec<Self::Item>`. As the example code in the `sorted_by` rustdoc shows, this is often not the most convenient.

```rust
let oldest_people_first = people
    .into_iter()
    .sorted_by(|a, b| Ord::cmp(&b.1, &a.1))
    // need a second into_iter(), which normally does not happen in iterator chains
    .into_iter()
    .map(|(person, _age)| person);
```

I think the opposite would be more intuitive -- have `sorted` return something like `std::vec::IntoIter` so that continuing the iterator chain works seamlessly, and have a method for extracting the underlying Vec efficiently if that's what they need.

```rust
// seamlessly continue the chain
iter.sorted().map(...)

// done, get a vec
iter.sorted().into_vec()
```

bluss:
Yes. Returning Vec is from a desire to not remove any possibilities or force allocation, but indeed we can improve this by switching the defaults around.

I kind of thing std is the best place to fix this, to add .into_vec() on `std::vec::IntoIter`. It has enough information to reuse the vector's original allocation.

matklad:
This will need clarification about algorithmic complexity in the documentation then, because two implementations with different performance characteristics are possible:

1) sort the whole vector at once, call `.into_iter()`
2) sort the vector lazily, using one-sided quick sort, such that `xs.sorted().take(10)` works in linear time, and not in `n log(n)`.

bluss:
Good idea. I want to recommend crate lazysort for that.

stjepang:
I'm interested in helping with this. Some thoughts:

1. We can implement `sorted` by tweaking `sort_unstable` (pattern-defeating quicksort) from the standard library. How does that sound?

2. Do we need both stable and unstable variants of `sorted`? For the stable variant, I'd probably use the same algorithm but partition elements a little bit differently.

3. When we get specialization in Rust, it'll be possible to implement a faster version of `partition` for slices by relying on `partition_in_blocks` (from `libcore/slice/sort.rs`).

4. Is there a reason why we don't have stable and unstable version of `partition`? C++ has `std::partition` and `std::stable_partition`.

bluss:
[...]

PR #251, opened 2017-12-25
Add Itertools::sorted_by_key

Xion:
Analogous to `sorted_by`, but takes a "key" function rather than a comparator. Uses `slice::sort_by_key` internally.


bluss:
Thanks. Will use a different return value **later**, in sync with #238

bluss:
Thank you!

PR #255, opened 2018-01-06
Return an iterator from sorted/sorted_by/sorted_by_key

### LINKED ISSUE #238: Return iterator from sorted / sorted_by

bluss:
Prefer returning an iterator for much more fluent use of this method.

A FromIterator special case in `Vec` means that the `std::vec::IntoIter` can be converted into a `Vec` with no cost, so for the case where a Vec is needed, we incur no cost except for the extra typing.

Fixes #238

### DIFF
--- a/src/free.rs
+++ b/src/free.rs
@@ -7,6 +7,9 @@
 use std::fmt::Display;
 use std::iter::{self, Zip};
 #[cfg(feature = "use_std")]
+type VecIntoIter<T> = ::std::vec::IntoIter<T>;
+
+#[cfg(feature = "use_std")]
 use Itertools;
 
 pub use adaptors::{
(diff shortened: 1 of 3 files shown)


stjepang:
I suggest returning a new type rather than `vec::IntoIter` because then we could implement lazy sorting.

bluss:
Part of the motivation is the (current, undocumented) free conversion between IntoIter and back to the Vec, with collect, but I suppose that can be preserved with a newtype and a custom forwarding collect impl.

PR #256, opened 2018-01-10
Let the resulting type of izip!(i, j) be std::iter::Zip<i, j>

### LINKED ISSUE #229: izip!(): Use just a.zip(b) if there are only two input parameters
Currently, `izip!()` always uses a map. It's redundant when we only have two input parameters.

Use simpler type or use more consistent type? The map closure's type is always anonymous anyway, so it can't be used for much.

bluss:
* Use the regular `Zip` iterator (without a .map()) when there are just
two inputs to `izip!()`.
* Use the regular iterator when there is just one input.

Fixes #229

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -257,13 +257,20 @@ macro_rules! izip {
         izip!(@closure ($p, b) => ( $($tup)*, b ) $( , $tail )*)
     };
 
-    ($first:expr, $second:expr $(,)*) => {
+    // unary
+    ($first:expr $(,)*) => {
         $crate::__std_iter::IntoIterator::into_iter($first)
+    };
+
+    // binary
+    ($first:expr, $second:expr $(,)*) => {
+        izip!($first)
             .zip($second)
     };
 
+    // n-ary where n > 2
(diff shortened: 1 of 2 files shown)


bluss:
cc @krdln ; I'm just wondering if you prefer the current (always .zip()s and one .map()) or if you prefer this PR.

krdln:
@bluss The no-map version is definitely better in my opinion! Less code for optimizer to remove :)

Speaking of the breaking-change label, I wanted to say it's not actually breaking, because we've never made any promise here, but now I see that docs specifically talk about zips and map, so I stand corrected. Anyway, I think docs should be updated to reflect that it's not always a `Map`. Or even just say that the return type is just an implementation detail.

bluss:
Hah, I agree it is not severe due to the common use of the macro but it is undoubtedly breaking, equivalent to changing the return type of a function.

bluss:
Good point about docs though. I changed the unary case with a corresponding simplification and updated docs. As you can see I'm piling up the breaking changes as open PRs now, so when they are all ready we can go to itertools 0.8 soon.

PR #257, opened 2018-01-11
Add Itertools::replace adaptor

jonas-schievink:
Allows replacing items from an iterator with the contents of another iterator.

This works similar to `str::replace`, but for arbitrary (clonable) types and on iterators, needing no allocation.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -89,6 +89,7 @@ pub mod structs {
     #[cfg(feature = "use_std")]
     pub use rciter_impl::RcIter;
     pub use repeatn::RepeatN;
+    pub use replace::Replace;
     pub use sources::{RepeatCall, Unfold, Iterate};
     #[cfg(feature = "use_std")]
     pub use tee::Tee;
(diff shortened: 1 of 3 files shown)


bluss:
Implementation looks good, please keep the `where` style like the rest of the crate. We can probably generalize the adaptor to use `T: PartialEq<U>` so that the two types don't have to be equal (as usual, useful for String).

On top of that, I'm not very convinced of the utility. I also wonder if there is a more general way to make a similar feature, where replace-element is a special case.

bluss:
Current replace is equivalent to this, and this is quite a mouthful (and more inefficient?). But something for the imagination:

```rust
use std::iter::once;
use itertools::Either::*;
let pattern = 1;
(0..10).flat_map(|elt| if elt != pattern { Left(once(elt)) } else { Right(vec![1, 2, 3].into_iter()) })
```

jonas-schievink:
A way to make `replace` more generic is to have it take a closure that returns a `bool` instead of hard-coding the `==` operation. I didn't find a way to make it more generic than that without arriving at something as verbose as your `flat_map` example.

bluss:
We discussed this and I wanted to decline this PR, the argument based on 1) doubt about utility 2) fitness for the proposed use case (there exists much more efficient ways to xml-escape). Thanks and it can always be added later if we find a better framing for the feature.

Issue #92, opened 2016-01-11
Group by that merges same key elements


muja:
After a quick glance I couldn't find a function like that. I believe it would be useful to have for unsorted Vecs, instead of creating multiple groups only for consecutive elements.

Thoughts?

bluss:
Sure, it's a bit like the current unique, except you collect all values that map to the same key.

matematikaadit:
I'm interested in submitting PR for this. But I'm not quite understand about the explanation. Care to gives an example?

lnicola:
Something like `GroupBy` or `ToLookup` here https://codeblog.jonskeet.uk/category/edulinq/, I think. The latter is eager, while the former is deferred until the oyter iterator is evaluated, I believe. That's in contrast with the other iterator adapters which are mostly lazy.

matematikaadit:
Sorry, not familiar with C#. A rust snippet on how the method is used and the intended result would be appreciated.

lnicola:
@matematikaadit And I don't really know Rust, so the following might make little sense. I offered that link as a guideline for the API, not for the implementation. Anyway, consider this code:

```rust
// numbers and their lengths in English:
let items = vec![(4, "zero"), (3, "one"), (3, "two"), (4, "four"), (4, "five"),
                         (3, "six"), (5, "seven"), (5, "eight"), (4, "nine"), (3, "ten")];
// group by length, associate keys with references to the original values
let grouped = items.iter().group_by(|&item| item.0).collect::<Vec<_>>();
// grouped: Vec<(usize, Vec<&({Integer}, &str)>)>

assert_eq!(grouped.len(), 3); // three distinct lengths

// the keys, in the original order, but distinct
assert_eq!(grouped[0].0, 4);
assert_eq!(grouped[1].0, 3);
assert_eq!(grouped[2].0, 5);

assert_eq!(*grouped[0].1[0].1, "zero");
assert_eq!(*grouped[0].1[4].1, "four");
assert_eq!(*grouped[2].1[3].1, "ten");

// that looks rather ugly, not that I've typed it

// another variant, also available in the .NET API:
// this one has key and value selectors
let grouped = items.iter().group_by(|&item| item.0, |&item| &item.1).collect::<Vec<_>>();
// grouped: Vec<(usize, Vec<&str>)>

// or, maybe more idiomatic and easier to implement without overloading, return keys and values
let grouped = items.iter().group_by(|&item| (item.0, item.1)).collect::<Vec<_>>();
[...]

PR #258, opened 2018-01-12
Add Itertools::to_lookup

tobz1000:
Fulfills #92.

Create a Lookup from an Iterator, in a similar manner to [C#'s ToLookup](https://msdn.microsoft.com/en-us/library/bb549073(v=vs.110).aspx).

Returns a plain `HashMap`, so no special `FromIterator` or `.iter()` implementations. My only concern is, if we do want to specialise this in future, it will require a new type and be a breaking change.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -31,6 +31,8 @@ extern crate core as std;
 
 pub use either::Either;
 
+#[cfg(feature = "use_std")]
+use std::collections::HashMap;
 use std::iter::{IntoIterator};
 use std::cmp::Ordering;
 use std::fmt;
@@ -132,6 +134,8 @@ mod groupbylazy;
 mod intersperse;
 #[cfg(feature = "use_std")]
 mod kmerge_impl;
+#[cfg(feature = "use_std")]
+mod lookup;
 mod merge_join;
 mod minmax;
 #[cfg(feature = "use_std")]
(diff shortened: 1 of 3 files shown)


bluss:
This is welcome but should be named so that it does not conflict or confuse with potential equivalents of collect to hashmap under names like into_hahmap.

src/lib.rs
```diff
@@ -488,6 +492,30 @@ pub trait Itertools : Iterator {
...
+    /// assert_eq!(lookup.get(&1), None);
+    /// assert_eq!(lookup[&2], vec![12, 42]);
+    /// assert_eq!(lookup[&3], vec![13, 33]);
+    /// ```
+    #[cfg(feature = "use_std")]
+    fn to_lookup<K, F>(self, get_key: F) -> HashMap<K, Vec<Self::Item>>
+        where Self: Iterator + Sized,
+              K: Hash + Eq,
+              F: Fn(&Self::Item) -> K
+    {
```
bluss: Naming aside (it is important too).

I would expect this to apply to iterators of pairs and use the first field as the key and the second as the value. Like `HashMap::from_iter` expects it. Then we don't have to worry about the diversity in calling mode required to extracting the key, be it by value (split the key out as a part of something) or by ref (clone/create a new key from a value).

tobz1000: I was actually thinking about just this :) haven't had much time to work on it this weekend but I was considering providing three different argument signatures:
* iterator of pairs,
* iterator with key function (i.e. the current version),
* iterator with both key and value functions.

Seems reasonable enough to only provide the first one. However, my initial reasoning for doing it as it currently is, was to match `.group_by()`. I can't see any reason why this function and that one would take different arguments. Maybe `.group_by()` should take a pair iterator too?

bluss: Good point, they seem to be in about the same situation. Maybe they differ in that the group by key is often not kept around.

It seems they should both do the same thing, and if I had thought of that, I'd have picked to mimic group_by as well.

tobz1000:
Changed name to `to_group_lookup`; not sure if it's what you had in mind. If not, let me know, or you can change it yourself after merge.

src/lib.rs
```diff
@@ -488,6 +492,28 @@ pub trait Itertools : Iterator {
...
+    /// assert_eq!(lookup[&3], vec![13, 33]);
+    /// ```
+    #[cfg(feature = "use_std")]
+    fn to_group_lookup<K, V>(self) -> HashMap<K, Vec<V>>
+        where Self: Iterator<Item=(K, V)> + Sized,
+              K: Hash + Eq,
+    {
+        lookup::to_group_lookup(self)
+    }
+
```
bluss: This is an iterator "regular" method, not an adaptor, so it should be listed in next to the other methods.

bluss:
Thanks. I'm going to suggest a name such as `.into_group_hashmap()`` or `.into_group_map()`, then we don't need to invent a new (for Rust) word like lookup.

bluss:
Edited; it's nice to say "Fulfills #92." but it is not what github prefers: https://help.github.com/articles/closing-issues-using-keywords/

bluss:
Thanks for the solid work and thanks for using quickcheck tests.

bluss:
This is finally released in itertools 0.7.7. Thanks, I think it's a very nice new feature.

PR #261, opened 2018-02-05
add is_sorted, is_sorted_by, is_sorted_by_key

gnzlbg:


gnzlbg:
@bluss looks like rand made a breaking change but did not bump the major semver version :(

src/lib.rs
```diff
@@ -953,6 +953,77 @@ pub trait Itertools : Iterator {
...
+    fn is_sorted_by<F>(mut self, mut compare: F) -> bool
+        where Self: Sized,
+              Self::Item: Ord,
+              F: FnMut(&Self::Item, &Self::Item) -> Ordering, 
+    {
+        let first = self.next();
+        if let Some(mut first) = first {
+            while let Some(second) = self.next() {
+                if compare(&first, &second) == Ordering::Greater {
+                    return false;
```
scottmcm: For a while-let with an early return, should this be `&mut self` instead, like `all` and `any`?

(nit: can it be written [in terms of `all`](https://medium.com/@veedrac/rust-is-slow-and-i-am-the-cure-32facc0fdcb#9a37), until `try_fold` is stable?)

gnzlbg: I'll give these a try!

gnzlbg: @scottmcm these should be fixed now

gnzlbg:
@scottmcm probably the solution to fixing the travis bots is to raise the minimum version of stable supported, or to disable tests that don't work on the stable version used. Bumping the minimum version of stable supported will require a major semver bump though (0.7 to 0.8).

bluss:
sad to see rand breaking that

bluss:
Hi, it looks good. Please place these new methods in the regular method section. For example after the sorted methods.

gnzlbg:
I've moved the methods. Is that place ok? I did not realize the methods were separated into sections :/

bluss:
It's in the Itertools trait doc, but it's impossible to see that doc due to the big ugly code block that starts a trait documentation page.

bluss:
> The trait Itertools: extra iterator adaptors and methods for iterators.

> This trait defines a number of methods. They are divided into two groups:

>    Adaptors take an iterator and parameter as input, and return a new iterator value. These are listed first in the trait. An example of an adaptor is .interleave()

>    Regular methods are those that don't return iterators and instead return a regular value of some other kind. .next_tuple() is an example and the first regular method in the list.

bluss:
Thanks. We need to see how the discussion for std is going. As witnessed with the flatten problems, it would suck to introduce these here only to have them in std shortly afterwards.

gnzlbg:
No worries, I have my own fork of itertools in the meantime and I can
migrate to nightly if these are incuded in std in the mean time. I just
wanted to upstream these here to show exactly which implementation I was
proposing in the Rust-Lang issue.

There were some questions about making these free functions instead but I
still do think that making these Iterator methods is the most flexible
approach.

bluss:
Superseded by work in std, rust-lang/rust/pull/55045

PR #265, opened 2018-03-04
Attempt to fix travis tests for Rust 1.12.0

bluss:
A fix that I copied from cuviper (Josh Stone) in the rayon project.

This resolves our testing & release deadlock with a workaround. Lovingly copy-pasted from @cuviper's work in rayon.


bluss:
It works for the problem it was solving - Rust 1.12. So fixing the nightly problem is another PR.

PR #268, opened 2018-03-11
Add some Debug implementations

phimuemue:
This covers a part of https://github.com/bluss/rust-itertools/issues/32

I looked for all `struct`s and `enum`s and added `#[derive(Debug)]` where possible. I added a `Debug` impl for `KMerge`, `KMergeBy`, `MergeJoinBy` using `debug_fmt_fields`, printing all fields except function callable objects.

### DIFF
--- a/src/adaptors/multi_product.rs
+++ b/src/adaptors/multi_product.rs
@@ -29,7 +29,7 @@ pub fn multi_cartesian_product<H>(iters: H) -> MultiProduct<<H::Item as IntoIter
     MultiProduct(iters.map(|i| MultiProductIter::new(i.into_iter())).collect())
 }
 
-#[derive(Clone)]
+#[derive(Clone, Debug)]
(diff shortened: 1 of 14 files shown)


bluss:
Nice. I'd skip the ones that use a closure, or fix them to ignore the closure, otherwise those debug impls almost never have any use?

phimuemue:
I agree. I skip the ones with a closure.

As a sidenote: I tried to implement `Debug` for `MultiProduct` and realized that `debug_fmt_fields` can currently not deal with it because it offers a single unnamed member (accessed by `self.0`) because `debug_fmt_fields` expects `ident`s as field names.

Is there an easy way to extend `debug_fmt_fields` to support those members, too?

bluss:
Maybe, it's a macro we can change as appopriate. We could also just use a struct with named fields.

bluss:
Thanks a lot for this, we can address the rest later.

PR #269, opened 2018-03-18
Add Itertools::tree_fold1

scottmcm:
This is useful for non-associative operations like building a tree (folding over a binop constructor) or combining floating point data (where this order keeps the magnitudes more similar).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -40,6 +40,7 @@ use std::fmt;
 use std::hash::Hash;
 #[cfg(feature = "use_std")]
 use std::fmt::Write;
+use std::mem;
 
 #[macro_use]
 mod impl_macros;
@@ -1599,6 +1600,40 @@ pub trait Itertools : Iterator {
               Self: Sized,
     {
         self.next().map(move |x| self.fold(x, f))
+    }
+
+    /// Accumulate the elements in the iterator in a balanced manner.
+    ///
+    /// This produces a result like `f(f(1, 2), f(3, 4))`, a balanced call tree,
+    /// instead of `f(f(f(1, 2), 3), 4)`, a linear sequence.
+    ///
+    /// If `f` is associative, prefer the normal `fold1` instead.
+    ///
+    /// ```
+    /// use itertools::Itertools;
+    ///
+    /// assert_eq!((0..9).fold1_balanced(|x, y| x - y), Some(-8));
+    /// let num_strings = (0..5).map(|x| x.to_string());
+    /// assert_eq!(num_strings.fold1_balanced(|x, y| format!("({}-{})", x, y)),
+    ///     Some(String::from("(((0-1)-(2-3))-4)")));
+    /// assert_eq!((0..0).fold1_balanced(|x, y| x * y), None);
+    /// ```
+    #[cfg(feature = "use_std")]
+    fn fold1_balanced<F>(self, mut f: F) -> Option<Self::Item>
+        where F: FnMut(Self::Item, Self::Item) -> Self::Item,
+              Self: Sized,
+    {
+        let hint = self.size_hint().0;
+        let cap = mem::size_of::<usize>() * 8 - hint.leading_zeros() as usize;
+        let mut stack = Vec::with_capacity(cap);
+        self.enumerate().foreach(|(mut i, mut x)| {
+            while (i & 1) != 0 {
+                x = f(stack.pop().unwrap(), x);
+                i >>= 1;
+            }
+            stack.push(x);
+        });
+        stack.into_iter().fold1(f)
     }
 
     /// An iterator method that applies a function, producing a single, final value.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -40,7 +40,8 @@ use std::fmt;
 use std::hash::Hash;
 #[cfg(feature = "use_std")]
 use std::fmt::Write;
-use std::mem;
+#[cfg(feature = "use_std")]
+use std::mem::size_of;
 
 #[macro_use]
 mod impl_macros;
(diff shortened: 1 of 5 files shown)


bluss:
Very nice algorithm, I'm mellowly neutral of its inclusion in itertools though. Some thoughts:

- It only builds one kind of tree, I'm unsure if that will often hit the use case? The example with 0..5 shows the root having 4 leaves on the left and 1 on the right, in that sense it is not balanced.
- For floating point operations I think more specialized algorithms would be appropriate (they can do compensated sums and vectorize, or both at the same time).
- A name like `tree_fold1` sounds better to me

Maybe this does have a good usage, even for example when applied to a sequence that already consists of "chunks" or items that are rather costly to process one by one.

scottmcm:
- It being neither complete nor balanced is a good point.  It's bottom-up merge sort merge order, but I don't know a great way to say that succinctly in an identifier.  (I'm also not managing to some up with a lazy way to do either complete-tree order or AVL-style-balanced order while still only using logarithmic memory.)
- Certainly for something like `sum` there are much better options than this.  The case I used this for was merging moments with https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Parallel_algorithm, which is already pretty stable, but this order keeps the weights more similar to each other for a little gain without much complexity.

I'm happy to make updates to it or close it.  I just thought it was interesting, general, and nuanced enough to be worth proposing, and itertools felt like a better fit than std.

bluss:
I think I'm tickled by it, it seems cool. It's mentioned here https://en.wikipedia.org/wiki/Fold_(higher-order_function)#Linear_vs._tree-like_folds but I can't find much else.

 I'll do review comments

src/lib.rs
```diff
@@ -1601,6 +1602,40 @@ pub trait Itertools : Iterator {
         self.next().map(move |x| self.fold(x, f))
     }
 
+    /// Accumulate the elements in the iterator in a balanced manner.
+    ///
+    /// This produces a result like `f(f(1, 2), f(3, 4))`, a balanced call tree,
+    /// instead of `f(f(f(1, 2), 3), 4)`, a linear sequence.
```
bluss: If we include a fifth argument here, we cover the asymmetry too, with `f(f(f(1, 2), f(3, 4)), 5)`, if that is readable?

Maybe we can do an ascii art that draws `f(f(f(1, 2), f(3, 4)), 5)` as a tree

src/lib.rs
```diff
@@ -40,6 +40,7 @@ use std::fmt;
 use std::hash::Hash;
 #[cfg(feature = "use_std")]
 use std::fmt::Write;
+use std::mem;
```
bluss: Skip this import or also use the same std conditional on it (to fix a warning).

src/lib.rs
```diff
@@ -1601,6 +1602,40 @@ pub trait Itertools : Iterator {
...
+    ///
+    /// ```
+    /// use itertools::Itertools;
+    ///
+    /// assert_eq!((0..9).fold1_balanced(|x, y| x - y), Some(-8));
+    /// let num_strings = (0..5).map(|x| x.to_string());
+    /// assert_eq!(num_strings.fold1_balanced(|x, y| format!("({}-{})", x, y)),
+    ///     Some(String::from("(((0-1)-(2-3))-4)")));
+    /// assert_eq!((0..0).fold1_balanced(|x, y| x * y), None);
+    /// ```
```
bluss: I think we can skip the first example that gives `-8` here. The 0..4 example should use 1..5 or 1..6 just like the initial example in this doc comment. Maybe it's easier to read with format string `"({} + {})"`, at least with the extra spacing.

(We try to keep in mind that examples serve as documentation.. if it needs better testing, we use tests for that. It's easy to be lazy here though, it happens. I wouldn't *mind* a quickcheck test for this method :slightly_smiling_face: )

src/lib.rs
bluss: Not that pretty, but why not

```text
        f
       / \
      /   \
     f     5
    / \
   /   \
  f     f
 / \   / \
1   2 3   4
```

scottmcm: Or maybe tilted?  I think of it this way:
```
        f
       /|
      f |
    _/| |
   /  | |
  f   f |
 /|  /| |
1 2 3 4 5
```
(Well, I actually think of it as that but upside down, so it's in dataflow order.)

Centril: I think there's too little air in the last version so I like the first one more -- easier to parse for me.

scottmcm: Or one that can be read left-to-right, top-to-bottom:
```text
1
 \
2-f
   \
3   \
 \   \
4-f---f
       \
5-------f
```

scottmcm: Hmm, maybe an orthogonal graph using box drawings:
```text
1 2 3 4 5 6 7
│ │ │ │ │ │ │
└─f └─f └─f │
  │   │   │ │
  └───f   └─f
      │     │
      └─────f
```

src/lib.rs
scottmcm: I realized I'd mistranslated my pop code to fold instead of rfold, and quickcheck totally found it (on the len-7 case, too, which is minimal).  Cool library!

scottmcm:
Ok, that should be much better now.  Two ideas I had but decided not to do:

1. I think I could make this core-only by using the call stack instead of a Vec (since it only goes logarithmicly deep).  That would mean not using internal iteration, though, and it definitely wouldn't be tail-recursive, so I'm not optimistic about how LLVM would do with it.

2. There's probably a way to do something like "the more accurate the lower size_hint, the closer to a complete binary tree".  But that sounded complex, and making the hint affect the output seemed like a really dangerous proposition.

src/lib.rs
bluss: oh wow, I just wanted to draw a diagram not start an ascii-off :smile:

Box drawings look nice in rustdoc, so that's fine too. The last one is good.

src/lib.rs
```diff
@@ -1601,6 +1603,86 @@ pub trait Itertools : Iterator {
...
+            while (i & 1) != 0 {
+                x = f(stack.pop().unwrap(), x);
+                i >>= 1;
+            }
+            stack.push(x);
+        });
+
+        // Every tree in the stack is bigger than all the trees that follow it
+        // combined, so rfold is better balanced here than forward.
+        stack.into_iter().rev().fold1(|a, b| f(b, a))
```
bluss: ah nice improvement

bluss:
It's well thought out, looks nice. I don't *want* to hold you out, but just using the call stack sounds very tempting. Can't we say the same thing for this version with a Vec, that the compiler won't really be able to take advantage of internal iteration in a significant way since we already have the pop/push operation in the innermost loop?

scottmcm:
Well, here's a recursion-style `no_std` version that works and seems to have ok perf.  I don't think I'm happy with it, though; it feels over-complicated to me.  I'll try a more mechanical translation tomorrow.

scottmcm:
Ok, I think I've found a `no_std` version with which I'm happy.  Let me know what you think; no rush, of course.  Small folds are probably particularly happy to not allocate.

(As for the push/pop, I suspect that when LLVM unrolled the loop, it noticed that things like `i&1` became constants, and thus folded the push/pop into nothing, or into just reserve.)

bluss:
quick benchmarks comparing the stack and vec at the WIP commit 2a44a984ab539d28cf9275eaf2dc94780e9a25cb, results in the commit log.

Interestingly (?), the tree and linear fold1 do different things for format!(), so we see a massive speedup when using tree fold instead of fold1 for this operation -- and this particular string concatenation is associative.

scottmcm:
Wow, your quick-and-dirty benchmarks are way less quick-and-dirty than mine were :laughing: 

That `format!` _is_ interesting.  I guess the allocated strings in `fold1` are length `n/2` on average, but the ones in `tree_fold1` are only length `4` on average.  I guess there's "non-associativity of performance"...

bluss:
I'm happy with this. I just installed bors in the itertools repo, so I can give you the accept power for this PR.

bors delegate+

bluss:
Thanks!

scottmcm:
And thank you, @bluss, for once again providing feedback that got my PR to a much better end state than where it started 🙂

PR #276, opened 2018-05-05
Use FNV hasher

akiekintveld:
The default `HashMap` uses SipHash, which is great for collision resistance, but can be slow in general cases where this isn't necessary. Specifically, it is much faster for small keys. This switches to the [FNV](https://crates.io/crates/fnv) hasher (using the very small `fnv` crate) for the `HashMap`s in `.unique()` and `.into_group_map()`. This results in faster Iterators with very little work.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,5 +1,5 @@
 #![warn(missing_docs)]
-#![crate_name = "itertools"]
+#![crate_name="itertools"]
 #![cfg_attr(not(feature = "use_std"), no_std)]
 
 //! Itertools — extra iterator adaptors, functions and macros.
@@ -22,7 +22,7 @@
 //!
 //! This version of itertools requires Rust 1.12 or later.
 //!
-#![doc(html_root_url = "https://docs.rs/itertools/0.7/")]
+#![doc(html_root_url="https://docs.rs/itertools/0.7/")]
 
 extern crate either;
 extern crate fnv;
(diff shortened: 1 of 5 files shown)


akiekintveld:
To back up the claim that it is faster for small keys: on my MacBook the benchmark I added runs at 26,115 ns/iter (+/- 2,329) on the default SipHash `HashMap` and at 9,890 ns/iter (+/- 549) after switching to the FNV hasher.

bluss:
Hi, thanks for this thoughtful improvement. We will not add more dependencies right now; itertools has a bunch of different functionality, and for most itertools users this will be a dependency that they don't have any benefit from. For that reason, I'll decline this PR.

src/lib.rs
```diff
@@ -1901,7 +1902,7 @@ pub trait Itertools : Iterator {
     /// assert_eq!(lookup[&3], vec![13, 33]);
     /// ```
     #[cfg(feature = "use_std")]
-    fn into_group_map<K, V>(self) -> HashMap<K, Vec<V>>
+    fn into_group_map<K, V>(self) -> FnvHashMap<K, Vec<V>>
```
bluss: Now it won't matter in this PR, but we'd usually be mindful here - the function signature changes, so it's a breaking change.

akiekintveld: Thanks for the review! I completely understand the aversion to adding more dependencies. I would like to note that `FnvHashMap` is just a type alias, so the change would technically be from `HashMap<K, Vec<V>>` to `HashMap<K, Vec<V>, FnvBuildHasher>`, so still a signature change but the hasher is more of an implementation detail.

PR #278, opened 2018-05-12
Cleanup unique implementation

akiekintveld:
`.unique()` is just a special case of `.unique_by()` where the function applied to each key is `clone`. (Keys must be `Clone` already because of the Entry API usage.) Realizing this, quite a bit of duplicate/redundant code can be eliminated. This PR also adds benchmarks for `unique()`.

### DIFF
--- a/src/unique_impl.rs
+++ b/src/unique_impl.rs
@@ -91,7 +91,7 @@ impl<I> Iterator for Unique<I>
     }
 
     fn count(self) -> usize {
-        self.iter.count()
+        count_new_keys(self.iter.used, self.iter.iter)
     }
 }
 
--- a/src/unique_impl.rs
+++ b/src/unique_impl.rs
@@ -1,6 +1,5 @@
 
 use std::collections::HashMap;
-use std::collections::hash_map::{Entry};
 use std::hash::Hash;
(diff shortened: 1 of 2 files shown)


akiekintveld:
I'm not sure the exact syntax for function types Rust wants here for 1.12.0 and 1.15.0.

bluss:
Hi, nice effort but I'd like to decline this change. Your unification uses call through function pointer to handle the cloning, and that has a negative performance impact. The current separation is the most straightforward way I know of, in that version of Rust, to avoid the overhead.

src/unique_impl.rs
```diff
@@ -83,20 +80,12 @@ impl<I> Iterator for Unique<I>
...
     fn next(&mut self) -> Option<I::Item> {
-        while let Some(v) = self.iter.iter.next() {
-            if let Entry::Vacant(entry) = self.iter.used.entry(v) {
-                let elt = entry.key().clone();
-                entry.insert(());
-                return Some(elt);
-            }
-        }
-        None
+        return self.iter.next();
```
bluss: (Cloning before we check if the element is unique, is another negative performance impact for types where Clone has a measurable cost).

PR #287, opened 2018-06-01
Add some convenience methods to EitherOrBoth

jethrogb:


jethrogb:
Ping

src/either_or_both.rs
```diff
@@ -8,3 +10,49 @@ pub enum EitherOrBoth<A, B> {
...
+
+    /// If `Right`, or `Both`, return `Some` with that value, otherwise, return `None`.
+    pub fn right(self) -> Option<B> {
+        match self {
+            Right(right) | Both(_, right) => Some(right),
+            _ => None
+        }
+    }
+
+    /// Converts from `EitherOrBoth<A, B>` to `EitherOrBoth<&A, &B>`.
```
bluss: Converts from `&EitherOrBoth<A, B>` in fact, isn't that more illuminating to the reader? Or not?

src/either_or_both.rs
```diff
@@ -8,3 +10,49 @@ pub enum EitherOrBoth<A, B> {
...
+    pub fn as_ref(&self) -> EitherOrBoth<&A, &B> {
+        match self {
```
bluss: Style wise this is preferred as `*self` and dropping the `&` on each line below. Same with mut. But that's not important enough to block merging.

src/either_or_both.rs
```diff
@@ -8,3 +10,49 @@ pub enum EitherOrBoth<A, B> {
...
+    pub fn has_left(&self) -> bool {
+        self.as_ref().left().is_some()
+    }
+
+    /// If `Right`, or `Both`, return true, otherwise, return false.
+    pub fn has_right(&self) -> bool {
+        self.as_ref().right().is_some()
+    }
+
+    /// If `Left`, or `Both`, return `Some` with that value, otherwise, return `None`.
```
bluss: "with the left value"

bluss:
Makes total sense. Let's edit the docs then it's done? Thanks

src/either_or_both.rs
jethrogb: I used the same phrasing as https://doc.rust-lang.org/std/option/enum.Option.html#method.as_ref . However, I agree that your suggestion makes sense.

bluss:
Thanks!

bluss:
bors: r+

PR #288, opened 2018-06-02
map_into method

### LINKED ISSUE #280: Support for `From::from` on the items of iterator
Hello

I find myself writing iterator method chains that contain `map(From::from)` ‒ I need to convert all the elements from one type to another. I was wondering, if there are convenience methods like `cloned` (I think that one is actually in std already) and `collect_vec`, would it make sense to have some kind of convenience method for this? I could send a pull request, but I want to ask first if it makes sense to anyone else and if someone has a good name for such method.

vorner:
To allow easy conversion of all elements.

 fixes #280 (`map_into` sounded better than `map_from`).

### DIFF
--- a/src/adaptors/mod.rs
+++ b/src/adaptors/mod.rs
@@ -1038,6 +1038,62 @@ impl_tuple_combination!(Tuple2Combination Tuple1Combination ; A, A, A ; a);
 impl_tuple_combination!(Tuple3Combination Tuple2Combination ; A, A, A, A ; a b);
 impl_tuple_combination!(Tuple4Combination Tuple3Combination ; A, A, A, A, A; a b c);
 
+/// An iterator adapter to apply `Into` conversion to each element.
+///
+/// See [`.map_into()`](../trait.Itertools.html#method.map_into) for more information.
+#[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
+pub struct MapInto<I, R> {
+    iter: I,
+    _res: PhantomData<fn() -> R>,
+}
+
+/// Create a new [`MapInto`](struct.MapInto.html) iterator.
+pub fn map_into<I, R>(iter: I) -> MapInto<I, R> {
+    MapInto {
+        iter: iter,
+        _res: PhantomData,
+    }
+}
+
+impl<I, R> Iterator for MapInto<I, R>
+    where I: Iterator,
+          I::Item: Into<R>,
+{
+    type Item = R;
+
+    fn next(&mut self) -> Option<R> {
+        self.iter
+            .next()
+            .map(|i| i.into())
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        self.iter.size_hint()
+    }
+
+    fn fold<Acc, Fold>(self, init: Acc, mut fold_f: Fold) -> Acc
+        where Fold: FnMut(Acc, Self::Item) -> Acc,
+    {
+        self.iter.fold(init, move |acc, v| fold_f(acc, v.into()))
+    }
+}
+
+impl<I, R> DoubleEndedIterator for MapInto<I, R>
+    where I: DoubleEndedIterator,
+          I::Item: Into<R>,
+{
+    fn next_back(&mut self) -> Option<Self::Item> {
+        self.iter
+            .next_back()
+            .map(|i| i.into())
+    }
+}
+
+impl<I, R> ExactSizeIterator for MapInto<I, R>
+where
+    I: ExactSizeIterator,
+    I::Item: Into<R>,
+{}
 
 /// An iterator adapter to apply a transformation within a nested `Result`.
 ///
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -83,6 +83,7 @@ pub mod structs {
         Product,
         PutBack,
         Batching,
+        MapInto,
         MapResults,
         Merge,
         MergeBy,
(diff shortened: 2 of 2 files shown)


vorner:
Added as a fixup, for easier review (a squash will be needed before merge).

I see why `size_hint` is beneficial, but won't the default `fold` just optimize out to basically the same thing?

pthariensflame:
If the underlying iterator has implemented `fold` in an optimized way, you'll take advantage of that by delegating, which might mean *O*(1) execution in some cases; the default `fold` has to rely on `next`, and so must necessarily be *O*(n) at least.

bluss:
Nice and simple, just need to make the decision to add it as a feature or not

Centril:
proptest does this, https://docs.rs/proptest/0.7.0/proptest/strategy/trait.Strategy.html#method.prop_map_into.

I think this should be proposed for libcore first and if it isn't accepted decide whether to include it in itertools or not.

Centril:
If it is added to itertools and then after the standard library, it will cause breakage for users of map_into in itertools, since Itertools is an extension trait of Iterator.

You don't have to write an RFC, a PR is enough.

vorner:
Ping. It didn't get accepted into std. What is the status here? Should this move forward, or get closed too?

Centril:
I do think this would be a useful addition myself.

Centril:
@bluss  I think it's about the clarity of being able to use turbofish here; `.map_into::<Alpha>()` reads quite well and helps type inference neatly. While you can use `.map(A::from)`, that applies not as often.

src/adaptors/mod.rs
```diff
@@ -1110,6 +1110,46 @@ impl<I, J> Iterator for Flatten<I, J>
 
 }
 
+/// An iterator adapter to apply `Into` conversion to each element.
+///
+/// See [`.map_into()`](../trait.Itertools.html#method.map_into) for more information.
+#[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
+pub struct MapInto<I, R> {
+    iter: I,
+    _res: PhantomData<R>,
```
bluss: Should be `PhantomData<fn() -> R`, we don't own an `R`, we just produce one. Implications for Send/Syncness for example, with this change `MapInto<I, X>` will be `Send` even if `X` is not.

src/lib.rs
```diff
@@ -617,6 +618,22 @@ pub trait Itertools : Iterator {
...
     }
 
+    /// Convert each item of the iterator using the `Into` trait.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// use itertools::Itertools;
+    ///
+    /// (1i32..42i32).map_into().collect::<Vec<f64>>();
```
bluss: Let's use an example with the turbofish maybe? Seems nice to put it like `.map_into::<f64>()` or is that ugly? :slightly_smiling_face:

bluss:
Ok I'm on board with this. I'd prefer if this was double ended conditionally (and exact size, and fused), just like Map is (and arguably less problematically so than Map). That can all be added on later if one wants.

bluss:
Gentle tip, say "fixes #280" in the pr description and the related issue will close automatically when this merges.

src/lib.rs
vorner: I'm not sure about that. In this context it would probably look odd (because I'd still have to use the turbofish on collect too, to tell it it's `Vec`).

Should I try to come up with some other example that makes more sense?

src/lib.rs
```diff
@@ -617,6 +618,22 @@ pub trait Itertools : Iterator {
         adaptors::step(self, n)
     }
 
+    /// Convert each item of the iterator using the `Into` trait.
+    ///
+    /// # Examples
```
bluss: Oh by the way this trait doesn't use headers right now. Traditionally, to not overcrowd the docs, the headings used to be too large compared with the methods..

vorner:
> Ok I'm on board with this. I'd prefer if this was double ended conditionally (and exact size, and fused),

I've also noticed the `Map` implements `TrustedLen` and `TrustedRandomAccess`. But the first is still unstable while the latter is hidden from the docs, so I don't know if it should be implemented.

> Gentle tip, say "fixes #280"

I'll try to remember to add it in a pre-merge rewrite of the branch (as there are fixups). I don't want to force-push in the middle of review.

src/lib.rs
bluss: Hm right, probably most users would avoid the double turbofish if they could! As you want, can be fixed later. Let's give it a go and see if we like it:

```rust
use itertools::Itertools;

(1i32..42i32).map_into::<f64>().collect::<Vec<_>>();
```

Or try a different example:

```rust
use itertools::Itertools;

let frequencies = [7, 5, 15, 5, 0, 2];
let average_value = frequencies.iter().cloned().map_into::<f64>().sum() / frequencies.len() as f64;
```

Uh, sometimes it just doesn't inspire me, the way we have to express these things (I think it has too much noise compared to the punch — `cloned()` is noise, and `iter()` is noise, both just to convert equivalentish things). Numeric applications really demand their own `.sum_as` or `.mean_value_as` which compute the accumulation using the specified type!

bluss:
You can add the `fixes` thing in just the PR description at the top here, that works fine for me.

We'll just have to skip unstable stuff in itertools for now. Those traits will not be stable in any shape until specialization comes around in some form.

src/lib.rs
vorner: I noticed itertools have `collect_vec`, so it can be used to avoid the double turbofish thing. Maybe that looks better?

bluss: yep!

bluss:
let's get this ready to be merged

vorner:
OK, rebased onto current master to avoid conflicts and squashed the fixups.

vorner:
I had to remove the FusedIterator, because on CI, 1.22 considers that one unstable.

bluss:
Ok, I see. FusedIterator is Rust 1.26. It's unlikely we leap that far in version requirements at this time, at least not for just for that trait. Thanks!

PR #289, opened 2018-06-04
Add Itertools::copied

xfix:
This is to pair with `cloned`. There are other pairs like that in Rust, like `copy_from_slice` and `clone_from_slice`. This allows to avoid accidentally calling `clone` while using `cloned` method, by allowing to use a specialization that only accepts `Copy` types.

### DIFF
--- a/src/adaptors/mod.rs
+++ b/src/adaptors/mod.rs
@@ -24,6 +24,52 @@ macro_rules! clone_fields {
         }
     );
 }
+/// An iterator that copies the elements of an underlying iterator.
+///
+/// See [`.copied()`](../trait.Itertools.html#method.copied) for more information.
+#[derive(Clone, Debug)]
+#[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
+pub struct Copied<I> {
+    it: I
+}
+
+pub fn copied<'a, I, T>(iterable: I) -> Copied<I::IntoIter>
+    where I: IntoIterator<Item=&'a T>,
+          T: 'a + Copy
+{
+    Copied { it: iterable.into_iter() }
+}
+
+impl<'a, I, T> Iterator for Copied<I>
+    where I: Iterator<Item=&'a T>,
+          T: Copy + 'a
+{
+    type Item = T;
+    #[inline]
+    fn next(&mut self) -> Option<T> {
+        self.it.next().map(|&x| x)
+    }
+
+    #[inline]
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        self.it.size_hint()
+    }
+}
+
+impl<'a, I, T> ExactSizeIterator for Copied<I>
+    where I: ExactSizeIterator<Item=&'a T>,
+          T: Copy + 'a
+{}
+
+impl<'a, I, T> DoubleEndedIterator for Copied<I>
+    where I: DoubleEndedIterator<Item=&'a T>,
+          T: Copy + 'a
+{
+    #[inline]
+    fn next_back(&mut self) -> Option<T> {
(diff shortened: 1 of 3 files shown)


bluss:
I'd vote to leave this out, it is a bit trivial and niche.

joshtriplett:
I'd really like to have this as well. I *often* find myself writing `.cloned()` for something that implements `Copy`, and I'd like to avoid accidentally doing an expensive clone if I refactor somewhere. I feel like `.copied()` would make my code more self-documenting, and enforce invariants I care about.

scottmcm:
>  There are other pairs like that in Rust, like copy_from_slice and clone_from_slice

That's a case where the Copy bound allows a different, probably-more-efficient implementation, though.  That wouldn't be the case with an `Itertools::copied`.

xfix:
Something worth noting is that `cloned` doesn't really do more than `map(|x| x.clone)` (other than being able to name an iterator struct).

rkarp:
This can actually be implemented much more simply, though I'm not sure if `deref` or `copied` would be the best name:

```rust
pub trait IteratorDerefExt {
    fn deref<'a, T>(self) -> Cloned<Self>
        where Self: Sized + Iterator<Item = &'a T>,
              T: 'a + Clone + Copy
    {
        self.cloned()
    }
    fn into_iter_deref<'a, T>(self) -> Cloned<Self::IntoIter>
        where Self: Sized + IntoIterator<Item = &'a T>,
              T: 'a + Clone + Copy
    {
        self.into_iter().cloned()
    }
}

impl<T> IteratorDerefExt for T {}
```
And yes, I'm in favor of adding this functionality as well.

bluss:
Ok this sound good to me

1. Can we check if this is something std wants to take before we do it?
2. If we take it, a free function copied would be a good addition too (like cloned).

I have a related problem that still doesn't have a good solution. It's a simple one: Declare a function that takes an iterator of `f32` values. Make the function generic so that it accepts both iterators of `&f32` and `f32` in practice. What's the best solution for this?

joshtriplett:
@bluss 
> Can we check if this is something std wants to take before we do it?

I'm all for it; I'd happily write a patch adding both `cloned` and `copied` to `core`.

> I have a related problem that still doesn't have a good solution. It's a simple one: Declare a function that takes an iterator of f32 values. Make the function generic so that it accepts both iterators of &f32 and f32 in practice.

Here's one solution:

https://play.rust-lang.org/?version=beta&mode=debug&edition=2018&gist=0b1520dcda103320ca0aeb0331157f58

```rust
fn sum_the_values<B, I>(values: I) -> f32
where
    B: std::borrow::Borrow<f32>,
    I: IntoIterator<Item = B>,
{
    let mut total = 0.0;
    for x in values {
        total += x.borrow();
    }
    total
}

fn main() {
    println!(
        "{} {}",
        sum_the_values(std::iter::repeat(1.1).take(5)),
        sum_the_values(std::iter::repeat(&1.1).take(5))
    );
}
```

bluss:
Neat. I think [this (playground)](https://play.rust-lang.org/?version=beta&mode=debug&edition=2018&gist=e93a944101233e2237d9cd74639a6172) is the expression I'd like, to make it easy to reuse. That is, to hide the Borrow magic behind a new trait and iterator adaptor that makes it simple to use.

bluss:
@joshtriplett .cloned() is already in libcore, and I guess there's a chance we can make a proposal for .copied() go through?

joshtriplett:
Given that nightly has this available now, I don't think itertools should add it. Given that, I think this can be closed.

PR #292, opened 2018-06-10
Permutations

### LINKED ISSUE #285: Permutation
Hi,

Thank you for making this awesome library. It would be great if you can support [permutate function like in python](https://docs.python.org/2/library/itertools.html#itertools.permutations)

Thanks!

tobz1000:
Fixes #285.

This implementation is based on the Python implementation specified in the issue. There are quite a few points which should be considered.

## Method name, _k_-less form

The adaptor function is called `permutations`, for user familiarity. However, since the size of the output can be specified, a more mathematically accurate name would probably be `k_permutations`. Perhaps two adaptor functions would be better: `.k_permutations(k: usize)` and `.permutations()` (which just sets `k = vals.len()`)?

## Item value ordering/distinctness

Input items are not inspected in any way; they are treated purely by their initial index. This means that:

  1. Permutations are yielded in lexicographical order of the index, not by any `PartialOrd` implementation of the items.
  2. Identical items will not be detected, and will result in some identical permutations.

__1__ can be achieved by the user by collecting-and-sorting their input.

__2__ is a little trickier, but can be achieved by filtering the output. However, I think there is a more efficient algorithm to avoid duplicated. Maybe we should provide this option?

## Permutations from source buffer/indices

In addition to the iterator adaptor, I've added `Permutations::from_vals`, to create directly from a `Vec` or a slice. This saves some cloning compared to using `source.[into_]iter().permutations(k)`.

There is also `Permutations::new(n, k)`, which is functionally equivalent to `(0..n).permutations(k)`, but a little faster (about 0.6x the run time).

But perhaps you would consider these unnecessary additions to the API?

## `PermutationSource` trait

These different implementations (from `Vec`/slice/just indices) are achieved with the trait `PermutationSource`. It's visible to the user to implement for other structures if they wish, but this is probably of limited value. There's not much harm in allowing the user to access it, but again, maybe it's just API bloat. (or a future breaking change when it's removed/changed...)

On the other hand, perhaps there are other places in the library which could benefit from taking a source generically? Any adaptor where input elements are used more than once will need to store them, and it might be more efficient to allow users to supply the memory directly. This could be done in another PR.

For completeness, I also made full implementations of the three variations, without the trait, for benchmarking. The pure-indices implementation is about 10% slower using the trait, but `Vec`s and slices are unaffected (or even a little faster...).

## `Combinations` changes

I've made small changes to the `Combinations` documentation to match `Permutations` - most significantly, replacing the letter `n` with `k`. I've also changed all uses of the variable `n` to `k` in the implementation... but maybe this is considered commit noise :)

There are some other potential changes which would bring `Combinations` and `Permutations` more in line with one another.

As mentioned above, `Combinations` might benefit from using a `_____Source` trait, to allow direct memory buffers.

My `Permutations` implementations doesn't make use of `LazyBuffer`. Perhaps the buffer is useful for iterators which never complete, and have a very small `k` value. Has any benchmarking been done?

Either way, it makes sense for both adaptors to either use or not use `LazyBuffer`. Maybe it could be integrated into the `_______Source` trait if it's useful?

---

Let me know what you think when you have the chance. (Sorry for submitting two reasonably big PRs at once. I hope you get a chance to go through it all eventually!)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -117,6 +117,8 @@ pub mod structs {
     pub use multipeek_impl::MultiPeek;
     pub use pad_tail::PadUsing;
     pub use peeking_take_while::PeekingTakeWhile;
+    #[cfg(feature = "use_std")]
+    pub use permutations::Permutations;
     pub use process_results_impl::ProcessResults;
     #[cfg(feature = "use_std")]
     pub use put_back_n_impl::PutBackN;
@@ -182,6 +184,8 @@ mod minmax;
 mod multipeek_impl;
 mod pad_tail;
 mod peeking_take_while;
+#[cfg(feature = "use_std")]
+mod permutations;
 mod process_results_impl;
 #[cfg(feature = "use_std")]
 mod put_back_n_impl;
@@ -1135,7 +1139,7 @@ pub trait Itertools : Iterator {
         adaptors::tuple_combinations(self)
     }
 
-    /// Return an iterator adaptor that iterates over the `n`-length combinations of
+    /// Return an iterator adaptor that iterates over the `k`-length combinations of
     /// the elements from an iterator.
     ///
     /// Iterator element type is `Vec<Self::Item>`. The iterator produces a new Vec per iteration,
@@ -1150,7 +1154,7 @@ pub trait Itertools : Iterator {
     ///     vec![1, 2, 4],
     ///     vec![1, 3, 4],
     ///     vec![2, 3, 4],
-    ///     ]);
+    /// ]);
     /// ```
     ///
     /// Note: Combinations does not take into account the equality of the iterated values.
@@ -1164,16 +1168,15 @@ pub trait Itertools : Iterator {
     ///     vec![2, 2],
     /// ]);
     /// ```
-    ///
     #[cfg(feature = "use_std")]
-    fn combinations(self, n: usize) -> Combinations<Self>
+    fn combinations(self, k: usize) -> Combinations<Self>
         where Self: Sized,
               Self::Item: Clone
     {
-        combinations::combinations(self, n)
+        combinations::combinations(self, k)
     }
 
-    /// Return an iterator that iterates over the `n`-length combinations of
+    /// Return an iterator that iterates over the `k`-length combinations of
     /// the elements from an iterator, with replacement.
     ///
     /// Iterator element type is `Vec<Self::Item>`. The iterator produces a new Vec per iteration,
@@ -1190,15 +1193,60 @@ pub trait Itertools : Iterator {
     ///     vec![2, 2],
     ///     vec![2, 3],
     ///     vec![3, 3],
-    ///     ]);
+    /// ]);
     /// ```
     #[cfg(feature = "use_std")]
-    fn combinations_with_replacement(self, n: usize) -> CombinationsWithReplacement<Self>
+    fn combinations_with_replacement(self, k: usize) -> CombinationsWithReplacement<Self>
     where
         Self: Sized,
         Self::Item: Clone,
     {
-        combinations_with_replacement::combinations_with_replacement(self, n)
+        combinations_with_replacement::combinations_with_replacement(self, k)
+    }
+
+    /// Return an iterator adaptor that iterates over all k-permutations of the
+    /// elements from an iterator.
+    ///
+    /// Iterator element type is `Vec<Self::Item>` with length `k`. The iterator
+    /// produces a new Vec per iteration, and clones the iterator elements.
+    ///
+    /// If `k` is greater than the length of the input iterator, the resultant
+    /// iterator adaptor will be empty.
+    ///
+    /// ```
+    /// use itertools::Itertools;
+    ///
+    /// let perms = (5..8).permutations(2);
+    /// itertools::assert_equal(perms, vec![
+    ///     vec![5, 6],
+    ///     vec![5, 7],
+    ///     vec![6, 5],
+    ///     vec![6, 7],
+    ///     vec![7, 5],
+    ///     vec![7, 6],
+    /// ]);
+    /// ```
+    ///
+    /// Note: Permutations does not take into account the equality of the iterated values.
+    ///
+    /// ```
+    /// use itertools::Itertools;
+    ///
+    /// let it = vec![2, 2].into_iter().permutations(2);
+    /// itertools::assert_equal(it, vec![
+    ///     vec![2, 2], // Note: these are the same
+    ///     vec![2, 2], // Note: these are the same
+    /// ]);
+    /// ```
+    ///
+    /// Note: The source iterator is collected lazily, and will not be
+    /// re-iterated if the permutations adaptor is completed and re-iterated.
+    #[cfg(feature = "use_std")]
+    fn permutations(self, k: usize) -> Permutations<Self>
+        where Self: Sized,
+              Self::Item: Clone
+    {
+        permutations::permutations(self, k)
     }
 
     /// Return an iterator adaptor that pads the sequence to a minimum length of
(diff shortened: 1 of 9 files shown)


ChrisJefferson:
Just to say, I'd really like a "permutation" function, so I'd like to see this PR, or something like it, merged.

estk:
@bluss Have you had a chance to look this over? This would scratch an itch for me.

bluss:
Sorry that I haven't had time. This looks good. The name sounds good to me, but probably explain to the users what a k-permutation means in the doc.

If you think something is a good addition, you can always help by jumping in and reviewing

bluss:
It's hard to compete with premutohedron performancewise maybe, since that one works with modifications in place, mainly.

tobz1000:
Is there anything else you need to move this forward? Or anyone you could nominate to review?

barabadzhi:
Is this PR dead? :worried:  
I badly seek `permutations` functionality similar to the one we have in Python.
Any merge plans? :thinking:

Avi-D-coder:
@barabadzhi it seems like all PRs are on hold, @bluss is not on very often.

src/permutations.rs
```diff
@@ -0,0 +1,265 @@
+/// An iterator to iterate through all the `k`-permutations of a series of items.
+///
+/// Source items are distinguished by their position, not value; so if there
+/// are identical items in the source, there will be some identical permutation
+/// iterations.
+///
+/// Can be constructed from an in-memory list of items directly; or from an
+/// iterator, with the
+/// [`.permuatations()`](../trait.Itertools.html#method.permutations) method.
+#[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
```
jswrenn: This annotation isn't strictly correct, given that:

> My `Permutations` implementations doesn't make use of `LazyBuffer`.

My only point of reluctance in merging this that the `Itertools::permutations` implementation collects `self`. Even without this annotation, there's a strong expectation that merely constructing an iterator adaptors are lazy and won't, for instance, loop infinitely (as this would on a cyclic iterator).

Nonetheless, I think this method is useful enough that it should be merged even in its non-lazy form, so long as:
 - this hazard is prominently documented
 - and there isn't any reason we couldn't make it lazy in the future without a breaking change.

Re. that second point: I'm worried making `PermutationSource` part of Itertool's public API exposes and commits us too much of `Permutations`'s internal implementation details.

So, I'd like to either:
1. merge a version `Permutations` that _is_ lazy, or
2. resolve my backward-compatibility concerns and merge a non-lazy `Permutations`.

tobz1000:
Agreed for keeping `PermutationSource` private.

I will have a look at re-using `LazyBuffer`, or otherwise not collecting the source iterator.

We could actually make permutations useful for infinite iterators by changing the order of permutations in a similar manner to #330. In fact, It might be a fairly simple case of flat-mapping the proposed iterator in that issue with sub-permutations where `k == n` (although maybe not the most performant way).

jswrenn:
Note https://github.com/rust-itertools/itertools/pull/340, which was modified to use `LazyBuffer`.

src/permutations.rs
```diff
@@ -0,0 +1,265 @@
...
+    fn perm_to_vec(&self, perm: &[usize]) -> Vec<usize> {
+        perm.to_vec()
+    }
+
+    fn len(&self) -> usize {
+        self.0
+    }
+}
+
+impl<T> PermutationSource for Vec<T>
```
tobz1000: It looks like this implementation is unnecessary, as collecting form an un-advanced vec `IntoIter` will use the original buffer anyway. So `vec![...].into_iter().permutations()` should be just as efficient as `Permutations::from_vals(vec![...])`.

It would be nice to remove `from_vals` from the public API altogether. I will think about getting `&[...].iter().permutations()` to also use the original buffer.

tobz1000: I've removed the code to optimise for already-collected buffers. I might re-add it in another PR in a way that works for more adaptors (`combinations` etc.) too.

tobz1000:
Infinite iterators are now handled in a similar manner to the combinations adaptors.

If iteration is stopped and restarted, the previously-collected buffer is reused, rather than running through the iterator again. This means side-effects will only occur on the first run-through. Let me know what you think of this.

src/lib.rs
```diff
@@ -1150,30 +1157,30 @@ pub trait Itertools : Iterator {
...
     /// ```
     ///
     /// Note: Combinations does not take into account the equality of the iterated values.
+    ///
     /// ```
     /// use itertools::Itertools;
     ///
     /// let it = vec![1, 2, 2].into_iter().combinations(2);
     /// itertools::assert_equal(it, vec![
-    ///     vec![1, 2], // Note: these are the same
```
jswrenn: Did you mean to delete these comments? They were recently added in #355.

tobz1000: My mistake, I'll re-add.

jswrenn:
Sorry for being slow in getting to this; it's been an exceptionally busy month. I think my only remaining reservation for merging is that the history is _very_ messy. Could you squash this down to just one commit that's rebased on master?

src/lib.rs
```diff
@@ -1135,12 +1139,15 @@ pub trait Itertools : Iterator {
...
     }
 
-    /// Return an iterator adaptor that iterates over the `n`-length combinations of
+    /// Return an iterator adaptor that iterates over the `k`-length combinations of
     /// the elements from an iterator.
     ///
     /// Iterator element type is `Vec<Self::Item>`. The iterator produces a new Vec per iteration,
     /// and clones the iterator elements.
     ///
+    /// If the input iterator is empty, or `k` is 0, or `k` greater than the
```
tobz1000: Incorrect; this does not match the behaviour of this adaptor

tobz1000:
Squashed this work onto master. I also updated some of the documentation - that change can be seen [here](https://github.com/tobz1000/rust-itertools/commit/7397d00776395271d4354ec42d7d05d64da1e89c).

Thank you for reviewing this!

tests/test_std.rs
```diff
@@ -585,12 +585,6 @@ fn combinations() {
     it::assert_equal((0..1).combinations(1), vec![vec![0]]);
     it::assert_equal((0..2).combinations(1), vec![vec![0], vec![1]]);
     it::assert_equal((0..2).combinations(2), vec![vec![0, 1]]);
-
-    // Infinite iterator should work, even though it doesn't really make sense
-    let mut infinite_combinations = (0..).combinations(2);
-    assert_eq!(infinite_combinations.next(), Some(vec![0, 1]));
-    assert_eq!(infinite_combinations.next(), Some(vec![0, 2]));
-    assert_eq!(infinite_combinations.next(), Some(vec![0, 3]));
```
jswrenn: Is there a reason you deleted these tests? They seem valid.

src/lib.rs
```diff
@@ -1997,19 +2046,21 @@ pub trait Itertools : Iterator {
     /// assert_eq!(successes, [1, 2]);
     /// assert_eq!(failures, [false, true]);
     /// ```
-    fn partition_map<A, B, F, L, R>(self, mut predicate: F) -> (A, B)
+    fn partition_map<A, B, F, L, R>(self, predicate: F) -> (A, B)
         where Self: Sized,
-              F: FnMut(Self::Item) -> Either<L, R>,
+              F: Fn(Self::Item) -> Either<L, R>,
```
jswrenn: I think something went amiss here in the rebase, too. This undos #366.

tests/test_std.rs
tobz1000: Those undos slipped through, let me try again...

tobz1000: Actually this was intentional - I replace these tests with some in `adaptors_no_collect.rs`.

jswrenn:
This looks great!

bors r+

PR #294, opened 2018-06-17
README.md: Clarify policy wrt. contribution and standard library

Centril:
See text in PR ;)

Basically we want to avoid `.flatten()` debacles and similar in the future.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -38,8 +38,8 @@ How to contribute
 - Include tests for your new feature, preferably a quickcheck test
 - Make a Pull Request
 
-For new features, please first file a PR to `rust-lang/rust <https://github.com/rust-lang/rust/>`_,
-adding your new feature to the `Iterator` trait of the standard library.
+For new features, please first consider filing a PR to `rust-lang/rust <https://github.com/rust-lang/rust/>`_,
+adding your new feature to the `Iterator` trait of the standard library, if you believe it is reasonable.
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -31,12 +31,19 @@ How to use in your crate:
...
 
-How to contribute:
+How to contribute
+-----------------
 
 - Fix a bug or implement a new thing
 - Include tests for your new feature, preferably a quickcheck test
 - Make a Pull Request
 
+For new features, please first file a PR to `rust-lang/rust <https://github.com/rust-lang/rust/>`_,
```
bluss: Thanks. Can we put some softening here, so that contributors must consider if it can be accepted into std? I think that many things are less general and are only interesting for itertools (even if we mainly like very general iterator functionality too).

bluss: Along the lines of, must consider but don't have to open a PR vs std if it's reasonable it doesn't belong there.

Centril: Perhaps something like "For new features, please first *consider* filing a PR to \<link\>, adding your new feature to the `Iterator` trait of the standard library, *if you believe it is reasonable*."
(added parts italicized)

bluss: yes

Centril:
Nuanced policy per discussion :)

bluss:
Thanks!

PR #296, opened 2018-06-27
Derive Eq and PartialEq for FoldWhile

jturner314:
This is useful for testing. (It's currently not possible to use `assert_eq!()` with `FoldWhile`.)


bluss:
Thanks!

PR #300, opened 2018-08-15
Add Itertools::{sum1, product1}

Emerentius:
This implements `sum1` and `product1` consumers that are like `sum` and `product` except they produce an `Option<T>` with `Some` for nonempty iters and `None` otherwise, just like `fold1`.
That allows one to distinguish between an empty iter and iters that sum/multiply to their neutral element.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -56,7 +56,7 @@ pub use either::Either;
 
 #[cfg(feature = "use_std")]
 use std::collections::HashMap;
-use std::iter::{IntoIterator};
+use std::iter::{IntoIterator, once};
 use std::cmp::Ordering;
 use std::fmt;
 #[cfg(feature = "use_std")]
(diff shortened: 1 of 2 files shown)


Emerentius:
>error: expected one of `)` or `,`, found `4`
   --> /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/libc-0.2.43/src/unix/mod.rs:107:46
    |
107 |     #[cfg_attr(feature = "align", repr(align(4)))]

Uh, I don't know what travis is complaining about

jswrenn:
Thanks for the contribution! Could you rebase this on master?

Emerentius:
Rebased and reworded the first line of the doc comments for both methods as commands `Iterate over all …` instead of statements `Iterates over all … ` for consistency with other methods.

PR #302, opened 2018-08-27
Bump MSRV to 1.13.0

matklad:
Current libc is not compatible wit 1.12.0, so it's probably time for
us to bump the min version as well.


bluss:
Thanks!

PR #303, opened 2018-09-03
Introduce peek_nth

davidcornu:
I've been implementing Bob Nystrom's [Crafting Interpreters](https://www.craftinginterpreters.com) in Rust as a way to learn the language and stumbled upon [`multipeek`](https://docs.rs/itertools/0.7.8/itertools/fn.multipeek.html) while looking for a way to `peek` more than one value ahead (as is required in order to [tokenize number literals](http://www.craftinginterpreters.com/scanning.html#number-literals)).

`multipeek` seemed like a great solution to this but I find it's tricky to use correctly. Because it makes `peek` stateful, you have to be careful to call [`reset_peek`](https://docs.rs/itertools/0.7.8/itertools/structs/struct.MultiPeek.html#method.reset_peek) whenever handing control of the iterator to another part of the code or risk running into bugs when `peek` doesn't return what you expect.

To work around this I implemented `peek_nth` which allows you to `peek` more than one iteration into the future without relying on a cursor. I have a working version in my project but I figured I'd submit it here for discussion.

**Example usage**

```rust
let xs = vec![1,2,3];
let mut iter = peek_nth(xs.iter());

assert_eq!(iter.peek(), Some(&&1));
assert_eq!(iter.next(), Some(&1));

// The iterator does not advance even if we call `peek_nth` multiple times
assert_eq!(iter.peek_nth(0), Some(&&2));
assert_eq!(iter.peek_nth(1), Some(&&3));
assert_eq!(iter.next(), Some(&2));

// Calling `peek_nth` past the end of the iterator will return `None`
assert_eq!(iter.peek_nth(1), None);
```

Note - `peek` equivalent to `peek_nth(0)` and is included for convenience.

----

⚠️ I'm _very_ new to Rust and took a lot of inspiration from`multipeek` to implement this. I'm more than willing to rewrite/fix any mistakes I may have made 😅.

### DIFF
--- a/src/peek_nth.rs
+++ b/src/peek_nth.rs
@@ -1,0 +1,102 @@
+use crate::size_hint;
+use crate::PeekingNext;
+use std::collections::VecDeque;
+use std::iter::Fuse;
+
+/// See [`peek_nth()`](../fn.peek_nth.html) for more information.
+#[derive(Clone, Debug)]
+pub struct PeekNth<I>
+where
+    I: Iterator,
+{
+    iter: Fuse<I>,
+    buf: VecDeque<I::Item>,
+}
+
+/// A drop-in replacement for `std::iter::Peekable` which adds a `peek_nth`
+/// method allowing the user to `peek` at a value several iterations forward
+/// without advancing the base iterator.
+///
+/// This differs from `multipeek` in that subsequent calls to `peek` or
+/// `peek_nth` will always return the same value until `next` is called
+/// (making `reset_peek` unnecessary).
+pub fn peek_nth<I>(iterable: I) -> PeekNth<I::IntoIter>
+where
+    I: IntoIterator,
+{
+    PeekNth {
+        iter: iterable.into_iter().fuse(),
+        buf: VecDeque::new(),
+    }
+}
+
+impl<I> PeekNth<I>
+where
+    I: Iterator,
+{
+    /// Works exactly like the `peek` method in `std::iter::Peekable`
+    pub fn peek(&mut self) -> Option<&I::Item> {
+        self.peek_nth(0)
+    }
+
+    /// Returns a reference to the `nth` value without advancing the iterator.
+    ///
+    /// # Examples
+    ///
+    /// Basic usage:
+    ///
+    /// ```rust
+    /// use itertools::peek_nth;
+    ///
+    /// let xs = vec![1,2,3];
+    /// let mut iter = peek_nth(xs.iter());
+    ///
+    /// assert_eq!(iter.peek_nth(0), Some(&&1));
+    /// assert_eq!(iter.next(), Some(&1));
+    ///
+    /// // The iterator does not advance even if we call `peek_nth` multiple times
+    /// assert_eq!(iter.peek_nth(0), Some(&&2));
+    /// assert_eq!(iter.peek_nth(1), Some(&&3));
+    /// assert_eq!(iter.next(), Some(&2));
+    ///
+    /// // Calling `peek_nth` past the end of the iterator will return `None`
+    /// assert_eq!(iter.peek_nth(1), None);
+    /// ```
+    pub fn peek_nth(&mut self, n: usize) -> Option<&I::Item> {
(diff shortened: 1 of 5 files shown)


davidcornu:
Ran my code against clippy and fixed all lints.

bluss:
Thanks, looks very thoughtful, I'm sorry I haven't had time to look

davidcornu:
No worries! Let me know if I can help in any way.

davidcornu:
@jswrenn is this feature something that would be of interest? Happy to fix the conflicts and get tests passing again if that's the case. Otherwise feel free to close.

jswrenn:
@davidcornu This is excellent. I'd be happy to merge it if you have time to fix up the conflicts and get the tests running.

davidcornu:
@jswrenn will do

davidcornu:
Done ✅

src/peek_nth.rs
```diff
@@ -0,0 +1,125 @@
...
+  /// assert_eq!(iter.peek_nth(1), Some(&&3));
+  /// assert_eq!(iter.next(), Some(&2));
+  ///
+  /// // Calling `peek_nth` past the end of the iterator will return `None`
+  /// assert_eq!(iter.peek_nth(1), None);
+  /// ```
+  pub fn peek_nth(&mut self, n: usize) -> Option<&I::Item> {
+    if n < self.buf.len() {
+      return Some(&self.buf[n]);
+    }
```
phimuemue, line 68: I know I'm a bit late to the party, but isn't this check redundant as per line 70?

davidcornu: Good catch! I didn't realize ranges would result in no iterations for cases where the start is larger than the end.

src/peek_nth.rs
```diff
@@ -0,0 +1,121 @@
...
+      match self.iter.next() {
+        Some(next_item) => {
+          self.buf.push_back(next_item);
+        }
+        None => return None,
+      }
+    }
+
+    Some(&self.buf[n])
+  }
```
jswrenn, line 76: I haven't tested this and I might be wrong, but I think this can be equivalently written as:
```suggestion
  pub fn peek_nth(&mut self, n: usize) -> Option<&I::Item> {
    self.buf.extend(self.iter.by_ref().take(n.saturating_sub(self.buf.len())));
    self.buf.get(n)
  }
```

src/peek_nth.rs
```diff
@@ -0,0 +1,121 @@
...
+{
+  type Item = I::Item;
+
+  fn next(&mut self) -> Option<I::Item> {
+    if self.buf.is_empty() {
+      self.iter.next()
+    } else {
+      self.buf.pop_front()
+    }
+  }
```
jswrenn, line 91: The `is_empty` check isn't strictly needed, since `pop_front()` will return `None` if `buf` is empty. That check can be removed like this:
```suggestion
  fn next(&mut self) -> Option<I::Item> {
    self.buf.pop_front().or_else(|| self.iter.next())
  }
```

src/peek_nth.rs
```diff
@@ -0,0 +1,121 @@
...
+          return None;
+        }
+      }
+    } else if let Some(r) = self.buf.get(0) {
+      if !accept(r) {
+        return None;
+      }
+    }
+    self.next()
+  }
```
jswrenn, line 120: Once again, **I have not tested this**, but I _think_ this might be equivalent:
```suggestion
  fn peeking_next<F>(&mut self, accept: F) -> Option<Self::Item>
  where
    F: FnOnce(&Self::Item) -> bool,
  {
    self.peek().filter(|elt| accept(elt))?;
    self.next()
  }
```

src/peek_nth.rs
davidcornu: Nice! This works pretty much as-is except for needing to account for `n` being 0-based.

davidcornu:
@jswrenn thanks for the comments! Definitely cleaned up the implementation.

jswrenn:
Thanks for the contribution!

bors r+

PR #310, opened 2018-10-03
Add exactly_one function

Xaeroxe:
Adds a small function I found uses for personally.  Can't be added to the core library because in the error case it allocates a vec and stores data in it.

### DIFF
--- a/src/exactly_one_err.rs
+++ b/src/exactly_one_err.rs
@@ -1,3 +1,5 @@
+use size_hint;
+
 /// Iterator returned for the error case of `IterTools::exactly_one()`
 /// This iterator yields exactly the same elements as the input iterator.
 ///
@@ -37,5 +39,16 @@ where
             .take()
             .or_else(|| self.first_two.1.take())
             .or_else(|| self.inner.next())
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let mut additional_len = 0;
+        if self.first_two.0.is_some() {
+            additional_len += 1;
+        }
+        if self.first_two.1.is_some() {
+            additional_len += 1;
+        }
+        size_hint::add_scalar(self.inner.size_hint(), additional_len)
     }
 }
--- a/src/exactly_one_err.rs
+++ b/src/exactly_one_err.rs
@@ -1,0 +1,41 @@
+/// Iterator returned for the error case of `IterTools::exactly_one()`
+/// This iterator yields exactly the same elements as the input iterator.
+///
+/// During the execution of exactly_one the iterator must be mutated.  This wrapper
+/// effectively "restores" the state of the input iterator when it's handed back.
+///
+/// This is very similar to PutBackN except this iterator only supports 0-2 elements and does not
+/// use a `Vec`.
+#[derive(Debug, Clone)]
+pub struct ExactlyOneErr<T, I>
+where
+    I: Iterator<Item = T>,
+{
+    first_two: (Option<T>, Option<T>),
+    inner: I,
+}
+
+impl<T, I> ExactlyOneErr<T, I>
+where
+    I: Iterator<Item = T>,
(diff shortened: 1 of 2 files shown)


Centril:
Relevant: https://github.com/rust-lang/rust/pull/55355.

bluss:
I'd prefer this in Centril's version. And now we know, std doesn't want this method, so we can take it.

Centril:
(Feel free to copy the contents from my PR if you want)

Xaeroxe:
@bluss Why do we prefer returning nothing in the case of there being more than 1?  Since both versions consume the iterator it seems to me we're just needlessly hiding valuable debugging info.

Xaeroxe:
If we simply don't like memory allocation for the error vec then I suppose we could make this take `&mut` instead and change the `Err` type to instead be `(Self::Item, Self::Item)`.

Centril:
In all the cases where I needed this I didn't want to get the remaining elements so the allocation shouldn't be pushed on folks imo; If you want to give back the iterator you can do: `-> Result<Self::Item, (Self::Item, Self)>` and then you can use `rest` in `Err(x, rest)` to get the remaining bits if you need that.

bluss:
We can return an iterator or similar, we already have one for iterator + one item in itertools, PutBack, so that s a nice one to put in the error case

Xaeroxe:
@bluss I've rewritten this to use an iterator in the error case.

bluss:
Oh right, we need to put back two elements. Either use putbackn or a custom iterator in that case, I don't mind the latter

Centril:
I think this turned out nicely. :)

Xaeroxe:
@bluss how's this?

Xaeroxe:
putbackn uses a vec internally which puts us back at the original problem, so I went for a custom iterator.

Xaeroxe:
@bluss Just pinging you in case this got lost.  Based on your github history it seems the past few days have been very busy for you.

Xaeroxe:
@bluss Believe this is ready to merge now.

src/exactly_one_err.rs
```diff
@@ -0,0 +1,58 @@
...
+    first_two: (Option<I::Item>, Option<I::Item>),
+    inner: I,
+}
+
+impl<I> ExactlyOneError<I>
+where
+    I: Iterator,
+{
+    /// Creates a new `ExactlyOneErr` iterator.
+    pub fn new(first_two: (Option<I::Item>, Option<I::Item>), inner: I) -> Self {
```
bluss: this should be pub(crate) and not public, but I'll fix it up

bluss:
Sorry, we need to add at test for this iterator before it merges. Then I agree it's ready. Can be a simple quickcheck test.

Xaeroxe:
@bluss is the doc comment test not sufficient?

bluss:
@Xaeroxe I realize I had thought about tests before and accepted the doctests, so I'm sorry that I didn't remember that. I shouldn't change my mind.

But in principle, no, features should have tests. Preferably quickcheck tests! They are very good at finding bugs. I feel strongly that testing documentation is often misunderstood — we test documentation to make sure that the examples we give actually work. We should make understandable examples, and they are there for documentation, not testing.

bluss:
So since it's annoying to change one's mind, I'll merge since I had already decided some point prior to just let it slide.

bluss:
bors r+

Thank you!

Xaeroxe:
@bluss well I mean you made a really good case and I agree with you, so I'm happy to add the unit tests if you want.

bluss:
@Xaeroxe ok, that can be a new PR if you want to

bluss:
Wheels of bors are already turning

Xaeroxe:
@bluss I've opened #324 , thanks!

Issue #293, opened 2018-06-11
Infinite sequence of all combinations


svenstaro:
Given a list of inputs `["A", "B", "C"]` I want an iterator that gives me an infinite combination output so that we get `["A", "B", "C", "AA", "AB", ..., "CCCC", ...`. I don't believe this is currently easily possible using itertools.

svenstaro:
Perhaps this is not "combinations" mathematically speaking but rather a product.

tobz1000:
This can probably be done with some adaption of `MultiProduct`. To add a bit more flexibility, it could take a meta-iterator, incremenally yielding items from the first iterator, then the product of the first two, etc. until termination (or not, if the meta-iterator is infinite).

e.g.
```rust
for x in (0..4).map(|_| 0..=1).incremental_product() {
    // [0] -> [1] -> [0, 0] -> [0, 1] ... [1, 1, 0, 1] -> [1, 1, 1, 0] -> [1, 1, 1, 1]
}
```

Then your use case could be achieved with
```rust
repeat(&["A", "B", "C"]).incremental_product()
```

PR #311, opened 2018-10-03
Infinite Sequence of All Combinations

EthanTheMaster:
Adds feature requested by #293 

Helper iterator adapter to make calculating all combinations easier.

### DIFF
--- a/src/product_combination.rs
+++ b/src/product_combination.rs
@@ -28,7 +28,8 @@ impl<I: Iterator + Clone> ProductCombination<I> {
                     break;
                 },
                 Some(ref mut pair) => {
-                    let (ref mut iter, ref mut item) = pair;
+                    let iter = &mut pair.0;
+                    let item = &mut pair.1;
 
                     let new_item = iter.next();
                     match new_item {


src/product_combination.rs
```diff
@@ -0,0 +1,67 @@
...
+impl<I> Iterator for ProductCombination<I>
+    where I: Iterator, I::Item: Clone
+{
+    type Item = Vec<I::Item>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.increment();
+
+        //Map each digit from the base data.len() to its corresponding element
+        let mut result: Self::Item = self.internal_state.iter().map(|idx| {
```
tobz1000: A short-form closure is more idiomatic here: 
```rust
self.internal_state.iter().map(|&idx| self.data[idx].clone())
```

I don't think `result` needs to be `mut` either.

.gitignore
```diff
@@ -1,2 +1,3 @@
 /target/
 Cargo.lock
+.idea/
```
tobz1000: This .gitignore change doesn't belong in a feature PR

EthanTheMaster:
@tobz1000 Thanks for the feedback! Support for huge vectors and infinitely size iterators has been added. Rather than collecting the iterator on construction, I just store the iterator and clone it whenever I needed to retrieve data. By doing this, the size of the iterator becomes irrelevant and doing things like mappings for infinitely sized iterators just work. However, cloning the iterator every time there is a retrieval comes with a performance hit as you have to iterate through n elements to reach the nth item.

src/product_combination.rs
```diff
@@ -0,0 +1,80 @@
...
+    fn increment(&mut self) {
+        let mut pointer: usize = 0;
+        let mut tack_on = false;
+        loop {
+            match self.internal_state.get_mut(pointer) {
+                None => {
+                    tack_on = true;
+                    break;
+                },
+                Some(ref mut pair) => {
```
tobz1000: We can't use default match binding in Itertools since it's still on Rust 1.12, but the tuple deref/item re-ref can still be done within the pattern match:
```rust
Some(&mut (ref mut iter, ref mut item)) => {
```

src/product_combination.rs
```diff
@@ -0,0 +1,80 @@
...
+}
+
+impl<I> Iterator for ProductCombination<I>
+    where I: Iterator + Clone, I::Item: Clone
+{
+    type Item = Vec<I::Item>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.increment();
+        Some(self.internal_state.iter().map(|ref pair| pair.1.clone()).collect())
```
tobz1000: The item yielded by `iter()` is already a reference; you don't need to use `ref` in the closure. Also, the item can be obtained with a pattern match:
```rust
.map(|&(_, ref item)| item.clone())
```

.gitignore
fyrchik: just a tip: you can put all your IDE-related ignores in `.git/info/exclude`

PR #323, opened 2018-12-11
Add {min,max}_set(_by{_key)?)? functions

zayenz:
The function min_set returns a Vec of all the minimum values. All variants for max and with key extraction and comparison functions are added also.

Since the functions need to return an unknown number of values and the values are not known until the iterator is finished, the function needs to allocate memory. Therefore Vec is used for returning the values. From the guidelines for itertools, this also makes it unlikely to be included in the standard library.

The reason for adding these functions is that I often find situations where I want to know all the minimum values. 

The naming is not optimal; `min_set` indicates that a set is returned while a `Vec` is returned. Another name is `all_min`, but I think that is less explanatory and not as discoverable. YMMV.

While an empty `Vec` could be used to handle the case of an empty iterator, I think that it is more safe to push that check into the type-system.

### DIFF
--- a/src/extrema_set.rs
+++ b/src/extrema_set.rs
@@ -1,0 +1,42 @@
+/// Implementation guts for `min_set`, `min_set_by`, and `min_set_by_key`.
+pub fn min_set_impl<I, K, F, L>(mut it: I,
+                                mut key_for: F,
+                                mut lt: L) -> Option<Vec<I::Item>>
+    where I: Iterator,
+          F: FnMut(&I::Item) -> K,
+          L: FnMut(&I::Item, &I::Item, &K, &K) -> bool,
+{
+    let (mut result, mut current_key) = match it.next() {
+        None => return None,
+        Some(element) => {
+            let key = key_for(&element);
+            (vec![element], key)
+        }
+    };
+
+    for element in it {
+        let key = key_for(&element);
+        if lt(&element, &result[0], &key, &current_key) {
+            result.clear();
+            result.push(element);
+            current_key = key;
+        } else if !lt(&result[0], &element, &current_key, &key) {
+            result.push(element);
+        }
+    }
+
+    Some(result)
+}
+
+/// Implementation guts for `ax_set`, `max_set_by`, and `max_set_by_key`.
+pub fn max_set_impl<I, K, F, L>(it: I,
+                                key_for: F,
+                                mut lt: L) -> Option<Vec<I::Item>>
+    where I: Iterator,
+          F: FnMut(&I::Item) -> K,
+          L: FnMut(&I::Item, &I::Item, &K, &K) -> bool,
+{
+    min_set_impl(it, key_for, |it1, it2, key1, key2| lt(it2, it1, key2, key1))
+}
+
+
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -168,6 +168,8 @@ mod combinations;
 mod combinations_with_replacement;
 mod exactly_one_err;
 mod diff;
+#[cfg(feature = "use_std")]
+mod extrema_set;
 mod format;
 #[cfg(feature = "use_std")]
 mod group_map;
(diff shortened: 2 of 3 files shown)


zayenz:
Just wondering if there is any interest in this feature?

fenhl:
I would find this feature useful, although wrapping in `Option` seems like a strange decision, as `Vec` methods that require elements to be present already return `Option`. I foresee many more `and_then` chains in my code if this is merged as is.

src/extrema_set.rs
```diff
@@ -0,0 +1,42 @@
...
+        None => return None,
+        Some(element) => {
+            let key = key_for(&element);
+            (vec![element], key)
+        }
+    };
+
+    for element in it {
+        let key = key_for(&element);
+        if lt(&element, &result[0], &key, &current_key) {
```
phimuemue: Would it make sense to replace `lt` (returning `bool`) by `cmp` (returning `Ordering`) so that we could avoid the second call to `lt` in line 23?

src/extrema_set.rs
```diff
@@ -0,0 +1,42 @@
+/// Implementation guts for `min_set`, `min_set_by`, and `min_set_by_key`.
+pub fn min_set_impl<I, K, F, L>(mut it: I,
+                                mut key_for: F,
+                                mut lt: L) -> Option<Vec<I::Item>>
```
phimuemue: What do you think about returning an empty `Vec` instead of `None`? I think it would make things a bit more clear, especially since the types do not indicate that a `Some` always contains a non-empty vector.

phimuemue: Do you think we could/should be a bit more generic than `Vec` here? I.e. should we try to be like `collect` (which returns something `FromIterator<Self::Item>` and lets the caller choose)?

src/extrema_set.rs
```diff
@@ -0,0 +1,42 @@
...
+        } else if !lt(&result[0], &element, &current_key, &key) {
+            result.push(element);
+        }
+    }
+
+    Some(result)
+}
+
+/// Implementation guts for `ax_set`, `max_set_by`, and `max_set_by_key`.
+pub fn max_set_impl<I, K, F, L>(it: I,
```
phimuemue: I am not sure if I would keep `max_set_impl` or if I would instead have a single `extremal_value_set_impl` (substituting both `min_set_impl` and `max_set_impl`) that would be called with an appropriate function.


This way we would have a bit fewer customization points: We would essentially always be calling `extremal_value_set_impl`, only the functor would change. I.e. we would e.g. call `extremal_value_set_impl`
* with `x<y` in `min_set`
* with `x>y` in `max_set`

src/lib.rs
```diff
@@ -1964,6 +1966,194 @@ pub trait Itertools : Iterator {
...
+    /// assert_eq!(a.iter().min_set(), Some(vec![&1]));
+    ///
+    /// let a = [1, 2, 3, 4, 5];
+    /// assert_eq!(a.iter().min_set(), Some(vec![&1]));
+    ///
+    /// let a = [1, 1, 1, 1];
+    /// assert_eq!(a.iter().min_set(), Some(vec![&1, &1, &1, &1]));
+    /// ```
+    ///
+    /// The elements can be floats but no particular result is guaranteed
```
phimuemue: To be honest, I often run into situations where I like to simply compare floats. But most of the time, I am happy that Rust tells me that `f64` is not `Ord`.

`Iterator::min` also requires `Ord`. Do you think we should stick to that requirement and require the caller to resolve the float issues?

zayenz: I agree, floats not being `Ord` is annoying, but also great.

In this case, I again copied the style and wording from `minmax`.

src/extrema_set.rs
zayenz: Sure, that would work well also. As said above, I'm happy to do the change, but will not have the time to do so for a couple of weeks.

src/extrema_set.rs
zayenz: I personally think that returning an Option is a good choice. In application code that I write, when using max or min functionality of sequences, an empty sequence is often an indication that something has gone awry and should be forced to be handled in another way. I imagine that in most cases, the Option-ness can be handled by the `?`-operator. Also, it matches the signature for max/min from the standard library, which returns an `Option` of an `Item`. I can naturally change it, but I do think that Option makes sense here.

As for why Vec and not some kind of iterator. The number of results is unknown before computing the result, and the result is unknown before going through the whole input iterator, which implies that some kind of temporary storage area is needed. Since the data is already in a Vec, I thought that the least convoluted thing was to just give the user back that. We can of course wrap the already constructed `Vec` in something that exposes it like a `FromIterator`, I have no real opinion here.

src/extrema_set.rs
zayenz: That should work nicely.

As it is, I essentially copied the structure from the `minimax` set of functions, where I see now that multiple comparisons are also made.

I'm happy to make the change, but will probably not have the time to do so in the next couple of weeks (currently on vacation).

zayenz:
> I would find this feature useful, although wrapping in `Option` seems like a strange decision, as `Vec` methods that require elements to be present already return `Option`. I foresee many more `and_then` chains in my code if this is merged as is.

See my reply here: https://github.com/rust-itertools/itertools/pull/323#discussion_r305603347

src/extrema_set.rs
phimuemue: Regarding the `Vec`: I naively thought about substituting the whole `Vec` even within the implementation (basically allowing to customize the kind of temporary storage), so that no conversion would be required upon returning it.

However, I see that `FromIterator` alone is possibly a bit too narrow to acchieve that. Maybe any container satisfying `Default + Extend` or `FromIterator + Extend` could be used?

src/extrema_set.rs
phimuemue: Ah, I see. That didn't occur to me when I looked at `minmax` last time. I could, however, imagine that things are a bit more complicated there because it needs to look for minimum and maximum simultaneously.

Maybe I should go back and have a look at `min_max`.

src/extrema_set.rs
zayenz: Since the code by neccesity (unless double iteration is used) _provisionally_ accumulates an unbounded number of result items, and discards these when a new provisionally best results is found, something more than `FromIterator + Extend` would be needed AFAIK. 

My guess would be that anything that also has the possibility to clear the results is no better than the interal `Vec`, and would also be less clear. I'd be glad to learn I'm wrong though!

src/extrema_set.rs
phimuemue: I had a look at `minmax` and think that it needs to do more stuff since it searches for both the minimum and the maximum. Here, I think we can use less calls to compare elements.

src/extrema_set.rs
```diff
@@ -0,0 +1,42 @@
...
+        Some(element) => {
+            let key = key_for(&element);
+            (vec![element], key)
+        }
+    };
+
+    for element in it {
+        let key = key_for(&element);
+        if lt(&element, &result[0], &key, &current_key) {
+            result.clear();
```
phimuemue: Referring to https://github.com/rust-itertools/itertools/pull/323#discussion_r310403165:

> Since the code by neccesity (unless double iteration is used) _provisionally_ accumulates an unbounded number of result items, and discards these when a new provisionally best results is found, something more than `FromIterator + Extend` would be needed AFAIK.
> 
> My guess would be that anything that also has the possibility to clear the results is no better than the interal `Vec`, and would also be less clear. I'd be glad to learn I'm wrong though!

I guess you are right that a `Vec` is the best choice for many (if not most) cases. However, if someone deliberately wants e.g. a `HashSet`, I think we ideally want to avoid the conversion of a `Vec` to a `HashSet`. I.e. I suggest that `max_set` should be usable somewhat like to `collect`:

```
let v = vec![0,2,0,1,2,1,0,2,0,1,1];
let i = vec.iter().cloned();
let as_vec      = i.clone().collect::<Vec<_>>();
let as_hashset  = i.clone().collect::<HashSet<_>>();
let max_vec     = i.clone().max_set::<Vec<_>>();
let max_hashset = i.clone().max_set::<HashSet<_>>(); // should imho work with HashSet internally
````

We could achieve that by replacing (https://github.com/rust-itertools/itertools/pull/323/commits/feea62176e8aa70c53debc5e91751a0fe7635a44#diff-1941d90b206003901f556eb26de20abcR20-R21)

```
result.clear();
result.push(element);
```

by

```
result = Some(element).into_iter().collect();
```

Similarily, we could replace (https://github.com/rust-itertools/itertools/pull/323/commits/feea62176e8aa70c53debc5e91751a0fe7635a44#diff-1941d90b206003901f556eb26de20abcR13)

```
(vec![element], key)
```

by

```
(Some(element).into_iter().collect(), key)
```

All of these would only require `FromIterator` (as far as I can see). Unfortunately, I do not know the performance implications of those changes, but I would hope that the compiler is smart enough to optimize it decently.

phimuemue: By the way: Is there a simple way to quote parts of a patch? (I just copy-pasted everything manually.)

phimuemue:
@zayenz Do you want to drive this to completion? (No worries if your answer is "no".)

If so, could you:
* Make the returned type simply a `Vec`? (No `Option` to denote emptiness.)
* Have the relevant functions require `Ord`? (Instead of `PartialOrd`; we can relax the constraints later if actually needed.)
* Avoid duplicate calls to `lt` by accepting a `FnMut(...)->Ordering` in `min_set_impl`?
* Resort to internal iteration (using `fold`/`for_each` instead of a manual `for`-loop)?

zayenz:
@phimuemue I'd forgotten about this one. Yes, I would like to fix this as I think the methods are quite useful.

I will take a look at updating the code, and then seeing about the changes mid next week.

phimuemue:
This is superseded by https://github.com/rust-itertools/itertools/pull/613.

PR #329, opened 2019-02-02
implement Iterator::nth for Combinations

### LINKED ISSUE #301: Feature: faster skipping combinations with nth
In `impl Iterator for Combinations`, only the `next` method is defined. It shouldn't be too difficult to permit skipping through combinations, so that `vec![1, 2, 3, 4].into_iter().combinations(2).nth(4)` doesn't unnecessarily waste time going through the first four iterations just to get to the fifth.

To do this, `Combinations` would need a field that kept track of how many iterations have already been made (this may replace the `first` bool field), its `indices` field would need to be settable immediately depending on the iteration number, and `nth` should thus be defined (save some pool-filling).

fyrchik:
Hello!
I have implemented `nth` method for `Combinations` iterator.

Closes #301 .
Let me know, what do you think. Other changes in tests are due to `rustfmt`.

### DIFF
--- a/src/combinations.rs
+++ b/src/combinations.rs
@@ -3,6 +3,8 @@ use std::fmt;
 use super::lazy_buffer::LazyBuffer;
 
 /// An iterator to iterate through all the `k`-length combinations in an iterator.
+/// Note: it iterates over combinations in lexicographic order and
+/// thus may not work as expected with infinite iterators.
 ///
 /// See [`.combinations()`](../trait.Itertools.html#method.combinations) for more information.
 #[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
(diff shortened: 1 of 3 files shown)


src/combinations.rs
```diff
@@ -40,63 +43,118 @@ pub fn combinations<I>(iter: I, n: usize) -> Combinations<I>
...
+        let mut result = Vec::with_capacity(self.n);
+        for i in self.indices.iter() {
+            result.push(self.pool[*i].clone());
+        }
+        Some(result)
+    }
+}
+
+// C(n,k) is binomial coefficient.
+fn c_n_k(n: usize, k: usize) -> usize {
```
zayenz: You might want to consider another way to implement binomial coefficients, since this will overflow fairly quickly, for example when k is 21 or higher.

Some googling shows a fairly nice variant of binomial coefficients in num-integer: https://docs.rs/num-integer/0.1.39/src/num_integer/lib.rs.html#772-791

fyrchik: Nice catch, thank you!
I was thinking how to do it better, the link you provided looks good.

fyrchik:
I have rebased this on `master`.

jswrenn:
Thanks! I'll take a look at this soon.

jswrenn:
@fyrchik I've filed a PR adding a quickcheck test on your feature branch: https://github.com/fyrchik/rust-itertools/pull/1

That test case is finding differences between the output of your specialized `Combinations::nth` and the current unspecialized method.

The specialized version of `Combinations::nth` needs to behave identically to the current unspecialized method before I can merge this PR.

fyrchik:
I have reimplemented this and got rid of binomial coefficients.
We are still advancing iterator one-by-one, however there is no intermediate allocations.
I have also added `quickcheck` tests from @jswrenn to be sure that new implementations behaves similar to the old.

jswrenn:
Thanks! Could you resolve the merge conflict? Also, note that the behavior of `combinations(0)` changed in #383.

fyrchik:
I have rebased on master. `cargo +nightly test` finished successfully.

Philippe-Cholet:
Hi @fyrchik, `Combinations` has evolved a bit since you worked on this.
_Do you want to update your work here?_ I would promptly respond/review.
We now have tests about specializations so I don't think we need your tests.

PR #331, opened 2019-02-19
Use assert_ne now that the minver is higher

scottmcm:
I saw that itertools 0.8.0 is now on rust 1.24, so I think this is ok now?

(Not worth a point release, but I think still good to do.)


bluss:
Thanks! Do you want to help maintain this crate? Basically looking for new owners, to be honest.

PR #335, opened 2019-03-01
Powerset iterator adaptor

willcrozi:
Implements a [powerset](https://en.wikipedia.org/wiki/Power_set) iterator adaptor that iterates over all subsets of the input iterator's elements. Returns vectors representing these subsets. Internally uses `Combinations` of increasing length.

I've taken the strategy of using a 'position' field that acts both as a means to to detect the special case of the first element and also allows optimal `size_hint()` implementation.

Additionally there is a commit to improve performance that alters `Combinations` implementation slightly. I've added Combinations benchmark as a stand-alone commit to allow checking for performance regressions. `Powerset` performance after this commit improves some cases (with small sizes of `n`) by 10-30%

This is my first attempt at a Rust contribution, happy to put in whatever discussion/work to get this merged. Cheers!

### DIFF
--- a/src/combinations.rs
+++ b/src/combinations.rs
@@ -31,18 +31,48 @@ impl<I> fmt::Debug for Combinations<I>
 pub fn combinations<I>(iter: I, k: usize) -> Combinations<I>
     where I: Iterator
 {
-    let mut pool: LazyBuffer<I> = LazyBuffer::new(iter);
-
-    for _ in 0..k {
-        if !pool.get_next() {
-            break;
-        }
-    }
+    let mut pool = LazyBuffer::new(iter);
+    pool.prefill(k);
 
     Combinations {
         indices: (0..k).collect(),
         pool,
         first: true,
+    }
+}
+
+impl<I: Iterator> Combinations<I> {
+    /// Returns the length of a combination produced by this iterator.
+    #[inline]
+    pub fn k(&self) -> usize { self.indices.len() }
+
+    /// Returns the (current) length of the pool from which combination elements are
+    /// selected. This value can change between invocations of [`next`].
+    ///
+    /// [`next`]: #method.next
+    #[inline]
+    pub fn n(&self) -> usize { self.pool.len() }
+
+    /// Resets this `Combinations` back to an initial state for combinations of length
+    /// `k` over the same pool data source. If `k` is larger than the current length
+    /// of the data pool an attempt is made to prefill the pool so that it holds `k`
+    /// elements.
+    pub(crate) fn reset(&mut self, k: usize) {
+        self.first = true;
+
+        if k < self.indices.len() {
+            self.indices.truncate(k);
+            for i in 0..k {
+                self.indices[i] = i;
+            }
+
+        } else {
+            for i in 0..self.indices.len() {
+                self.indices[i] = i;
+            }
+            self.indices.extend(self.indices.len()..k);
+            self.pool.prefill(k);
+        }
     }
 }
 
@@ -53,7 +83,7 @@ impl<I> Iterator for Combinations<I>
     type Item = Vec<I::Item>;
     fn next(&mut self) -> Option<Self::Item> {
         if self.first {
-            if self.pool.is_done() {
+            if self.k() > self.n() {
                 return None;
             }
             self.first = false;
--- a/src/combinations.rs
+++ b/src/combinations.rs
@@ -53,6 +53,10 @@ impl<I: Iterator> Combinations<I> {
     #[inline]
     pub fn n(&self) -> usize { self.pool.len() }
 
+    /// Returns a reference to the source iterator.
+    #[inline]
+    pub(crate) fn src(&self) -> &I { &self.pool.it }
+
     /// Resets this `Combinations` back to an initial state for combinations of length
     /// `k` over the same pool data source. If `k` is larger than the current length
     /// of the data pool an attempt is made to prefill the pool so that it holds `k`
--- a/src/lazy_buffer.rs
+++ b/src/lazy_buffer.rs
@@ -24,10 +24,6 @@ where
         self.buffer.len()
     }
 
-    pub fn is_done(&self) -> bool {
-        self.done
-    }
-
     pub fn get_next(&mut self) -> bool {
         if self.done {
             return false;
@@ -42,6 +38,17 @@ where
                 self.done = true;
                 false
             }
+        }
+    }
+
+    pub fn prefill(&mut self, len: usize) {
+        let buffer_len = self.buffer.len();
+
+        if !self.done && len > buffer_len {
+            let delta = len - buffer_len;
+
+            self.buffer.extend(self.it.by_ref().take(delta));
+            self.done = self.buffer.len() < len;
         }
     }
 }
--- a/src/powerset.rs
+++ b/src/powerset.rs
@@ -1,7 +1,9 @@
 use std::fmt;
+use std::usize;
 use alloc::vec::Vec;
 
 use super::combinations::{Combinations, combinations};
+use super::size_hint;
 
 /// An iterator to iterate through the powerset of the elements from an iterator.
 ///
@@ -10,20 +12,22 @@ use super::combinations::{Combinations, combinations};
 #[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
 pub struct Powerset<I: Iterator> {
     combs: Combinations<I>,
+    // Iterator `position` (equal to count of yielded elements).
+    pos: usize,
 }
 
 impl<I> Clone for Powerset<I>
     where I: Clone + Iterator,
           I::Item: Clone,
 {
-    clone_fields!(combs);
+    clone_fields!(combs, pos);
 }
 
 impl<I> fmt::Debug for Powerset<I>
     where I: Iterator + fmt::Debug,
           I::Item: fmt::Debug,
 {
-    debug_fmt_fields!(Powerset, combs);
+    debug_fmt_fields!(Powerset, combs, pos);
 }
 
 /// Create a new `Powerset` from a clonable iterator.
@@ -31,7 +35,10 @@ pub fn powerset<I>(src: I) -> Powerset<I>
     where I: Iterator,
           I::Item: Clone,
 {
-    Powerset { combs: combinations(src, 0) }
+    Powerset {
+        combs: combinations(src, 0),
+        pos: 0,
+    }
 }
 
 impl<I> Iterator for Powerset<I>
@@ -43,14 +50,34 @@ impl<I> Iterator for Powerset<I>
 
     fn next(&mut self) -> Option<Self::Item> {
         if let Some(elt) = self.combs.next() {
+            self.pos = self.pos.saturating_add(1);
             Some(elt)
         } else if self.combs.k() < self.combs.n()
             || self.combs.k() == 0
         {
             self.combs.reset(self.combs.k() + 1);
-            self.combs.next()
+            self.combs.next().map(|elt| {
+                self.pos = self.pos.saturating_add(1);
+                elt
+            })
         } else {
             None
+        }
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        // Total bounds for source iterator.
+        let src_total = size_hint::add_scalar(self.combs.src().size_hint(), self.combs.n());
+
+        // Total bounds for self ( length(powerset(set) == 2 ^ length(set) )
+        let self_total = size_hint::pow_scalar_base(2, src_total);
+
+        if self.pos < usize::MAX {
+            // Subtract count of elements already yielded from total.
+            size_hint::sub_scalar(self_total, self.pos)
+        } else {
+            // Fallback: self.pos is saturated and no longer reliable.
+            (0, self_total.1)
         }
     }
 }
--- a/src/powerset.rs
+++ b/src/powerset.rs
@@ -1,0 +1,56 @@
+use std::fmt;
+use alloc::vec::Vec;
+
+use super::combinations::{Combinations, combinations};
+
+/// An iterator to iterate through the powerset of the elements from an iterator.
+///
+/// See [`.powerset()`](../trait.Itertools.html#method.powerset) for more
+/// information.
+#[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
+pub struct Powerset<I: Iterator> {
+    combs: Combinations<I>,
+}
+
+impl<I> Clone for Powerset<I>
+    where I: Clone + Iterator,
+          I::Item: Clone,
+{
+    clone_fields!(combs);
+}
+
+impl<I> fmt::Debug for Powerset<I>
+    where I: Iterator + fmt::Debug,
+          I::Item: fmt::Debug,
+{
+    debug_fmt_fields!(Powerset, combs);
+}
+
+/// Create a new `Powerset` from a clonable iterator.
+pub fn powerset<I>(src: I) -> Powerset<I>
+    where I: Iterator,
+          I::Item: Clone,
+{
+    Powerset { combs: combinations(src, 0) }
+}
+
+impl<I> Iterator for Powerset<I>
+    where
+        I: Iterator,
+        I::Item: Clone,
+{
+    type Item = Vec<I::Item>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if let Some(elt) = self.combs.next() {
+            Some(elt)
+        } else if self.combs.k() < self.combs.n()
+            || self.combs.k() == 0
+        {
+            self.combs.reset(self.combs.k() + 1);
+            self.combs.next()
+        } else {
+            None
+        }
+    }
+}
--- a/src/size_hint.rs
+++ b/src/size_hint.rs
@@ -3,6 +3,7 @@
 
 use std::usize;
 use std::cmp;
+use std::u32;
 
 /// **SizeHint** is the return type of **Iterator::size_hint()**.
 pub type SizeHint = (usize, Option<usize>);
@@ -71,6 +72,20 @@ pub fn mul_scalar(sh: SizeHint, x: usize) -> SizeHint {
     let (mut low, mut hi) = sh;
     low = low.saturating_mul(x);
     hi = hi.and_then(|elt| elt.checked_mul(x));
+    (low, hi)
+}
+
+/// Raise `base` correctly by a **`SizeHint`** exponent.
+#[inline]
+pub fn pow_scalar_base(base: usize, exp: SizeHint) -> SizeHint {
+    let exp_low = cmp::min(exp.0, u32::MAX as usize) as u32;
+    let low = base.saturating_pow(exp_low);
+
+    let hi = exp.1.and_then(|exp| {
+        let exp_hi = cmp::min(exp, u32::MAX as usize) as u32;
+        base.checked_pow(exp_hi)
+    });
+
     (low, hi)
 }
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -67,3 +67,7 @@ harness = false
 [[bench]]
 name = "bench1"
 harness = false
+
+[[bench]]
+name = "combinations"
+harness = false
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -71,3 +71,7 @@ harness = false
 [[bench]]
 name = "combinations"
 harness = false
+
+[[bench]]
+name = "powerset"
+harness = false
--- a/benches/combinations.rs
+++ b/benches/combinations.rs
@@ -1,0 +1,125 @@
+use criterion::{black_box, criterion_group, criterion_main, Criterion};
+use itertools::Itertools;
+
+// approximate 100_000 iterations for each combination
+const N1: usize = 100_000;
+const N2: usize = 448;
+const N3: usize = 86;
+const N4: usize = 41;
+const N14: usize = 21;
+
+fn comb_for1(c: &mut Criterion) {
+    c.bench_function("comb for1", move |b| {
+        b.iter(|| {
+            for i in 0..N1 {
+                black_box(vec![i]);
+            }
+        })
+    });
+}
+
+fn comb_for2(c: &mut Criterion) {
+    c.bench_function("comb for2", move |b| {
+        b.iter(|| {
+            for i in 0..N2 {
+                for j in (i + 1)..N2 {
+                    black_box(vec![i, j]);
+                }
+            }
+        })
+    });
+}
+
+fn comb_for3(c: &mut Criterion) {
+    c.bench_function("comb for3", move |b| {
+        b.iter(|| {
+            for i in 0..N3 {
+                for j in (i + 1)..N3 {
+                    for k in (j + 1)..N3 {
+                        black_box(vec![i, j, k]);
+                    }
+                }
+            }
+        })
+    });
+}
+
+fn comb_for4(c: &mut Criterion) {
+    c.bench_function("comb for4", move |b| {
+        b.iter(|| {
+            for i in 0..N4 {
+                for j in (i + 1)..N4 {
+                    for k in (j + 1)..N4 {
+                        for l in (k + 1)..N4 {
+                            black_box(vec![i, j, k, l]);
+                        }
+                    }
+                }
+            }
+        })
+    });
+}
+
+fn comb_c1(c: &mut Criterion) {
+    c.bench_function("comb c1", move |b| {
+        b.iter(|| {
+            for combo in (0..N1).combinations(1) {
+                black_box(combo);
+            }
+        })
+    });
+}
+
+fn comb_c2(c: &mut Criterion) {
+    c.bench_function("comb c2", move |b| {
+        b.iter(|| {
+            for combo in (0..N2).combinations(2) {
+                black_box(combo);
+            }
+        })
+    });
+}
+
+fn comb_c3(c: &mut Criterion) {
+    c.bench_function("comb c3", move |b| {
+        b.iter(|| {
+            for combo in (0..N3).combinations(3) {
+                black_box(combo);
+            }
+        })
+    });
+}
+
+fn comb_c4(c: &mut Criterion) {
+    c.bench_function("comb c4", move |b| {
+        b.iter(|| {
+            for combo in (0..N4).combinations(4) {
+                black_box(combo);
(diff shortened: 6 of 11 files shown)


.travis.yml
```diff
@@ -29,5 +29,6 @@ branches:
   only:
     - master
     # bors branches
+    - powerset
```
jswrenn, line 32: What's the reason for this?

jswrenn (commented):
I want to get this into the next Itertools release! Sorry for the long delay in review!

.travis.yml
willcrozi: Great to hear, thanks! I'll remove that and push again, it was me just trying to get the CI building cleanly on my fork before updating this pull request (I forgot that the 'powerset' branch was still active for this pull request).

I've rewritten this feature and will post some further notes later on today.

willcrozi:
> I want to get this into the next Itertools release! Sorry for the long delay in review!

Great to hear, thanks!

I've rewritten this feature since the original pull request. I think the code is ready and was intending on writing some explanatory notes, which I can post here later today.

It seems the CI build for the Powerset benchmark times out (it completes successfully on my fork). I'll take a look into this.

willcrozi:
Travis CI jobs 2 and 3 are both getting stuck and terminated after a 10 minute limit. I've reduced the length of the powerset benchmarks (needed doing anyway TBH) but this seems related to caching.

>```
>Setting up build cache
>
>adding /home/travis/.cargo to cache
>creating directory /home/travis/.cargo
>adding /home/travis/build/rust-itertools/itertools/target to cache
>creating directory /home/travis/build/rust-itertools/itertools/target
>adding /home/travis/.rustup to cache
>creating directory /home/travis/.rustup
>adding /home/travis/.cache/sccache to cache
>creating directory /home/travis/.cache/sccache
>
>No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
>
>Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received
>
>The build has been terminated```

willcrozi:
My own use case for this adaptor relates to generating passwords to feed into tools such as hashcat. Since a powerset is essentially a chain of combinations of increasing length over the input set, I've made use of the `Combinations` iterator. 

The additions to `Combinations` and `LazyBuffer` are to allow a `Powerset` to efficiently expand and reuse a `Combinations` rather than repeatedly creating new ones (with the associated cost of cloning of the source buffer). This makes a difference in performance for my use case (lots of small powersets within large iterator chains), but for most uses this is likely amortized for larger powersets.

`Powerset`'s need to detect the special case of its first element, combined with the usefulness of a functional size hint for large powersets, seems to me to warrant `Powerset`'s `pos` field.

`size_hint::two_exp()` was added for `Powerset`'s size hint but is quite specific and could be folded into `Powerset`'s size_hint method instead.

The benchmarks for `Combinations` was added to check performance regression due to the current modifications. The `Powerset` benchmark was added for completeness.

Let me know what you think.

src/powerset.rs
```diff
@@ -0,0 +1,149 @@
...
+// Represents the inner state of a `Powerset` iterator.
+#[derive(Clone, Debug)]
+enum Inner<I: Iterator> {
+    // Buffering mode, source iterator still yielding elements.
+    Buffer(I, Vec<I::Item>),
+    // Combination mode, yielding combinations of buffered elements.
+    Combs(Combinations<I>),
+    // Temporary state for in-place switching of Buffer->Combs.
+    Empty_,
+}
```
phimuemue, line 54: Couldn't we just always have a `Combinations`, starting with `k=0`, and increment `k` after the `k`-combinations are exhausted, until we reach `k==n+1`?

This would obviate many of the case distinctions on `inner`, I guess.

Or is there a reason why this `enum`/case distinction is strictly necessary?

src/powerset.rs
```diff
@@ -0,0 +1,149 @@
...
+
+impl<I> Powerset<I>
+    where I: Iterator,
+          I::Item: Clone,
+{
+    // Upgrades the inner state to Inner::Combinations and returns the first
+    // combination, should only be called when self.inner value is Inner::Buffer.
+    #[inline]
+    fn upgrade_and_get_next(&mut self) -> Option<Vec<I::Item>> {
```
phimuemue, line 133: If we really need this, I think this function should be inlined, as it makes clear that it is only called for `Inner::Buffer`, possibly allowing further simplifications.

src/size_hint.rs
```diff
@@ -74,6 +74,31 @@ pub fn mul_scalar(sh: SizeHint, x: usize) -> SizeHint {
...
+    let (mut low, mut hi) = exp;
+
+
+    // saturating 'raise two by exponent'
+    let shl_range = 1_usize.leading_zeros() as usize;
+    low = if low <= shl_range {
+        1 << low
+    } else {
+        usize::MAX
+    };
```
phimuemue, line 90: Isn't there `saturating_pow` or similar?

src/size_hint.rs
```diff
@@ -74,6 +74,31 @@ pub fn mul_scalar(sh: SizeHint, x: usize) -> SizeHint {
...
+
+    // checked 'raise two by exponent'
+    hi = match hi {
+        Some(hi) if hi <= shl_range  => Some(1 << hi),
+        _ => None
+    };
+
```
phimuemue, line 97: Isn't there `checked_pow` or similar?

src/combinations.rs
```diff
@@ -27,22 +27,70 @@ impl<I> fmt::Debug for Combinations<I>
...
+    pub fn k(&self) -> usize { self.indices.len() }
+
+    /// Returns the (current) length of the pool from which combination elements are
+    /// selected. This value can change between invocations of `next()` and `init()`.
+    #[inline]
+    pub(crate) fn n(&self) -> usize { self.pool.len() }
+
+    /// Returns a reference to the source iterator.
+    #[inline]
+    pub(crate) fn src(&self) -> &I { &self.pool.it }
```
phimuemue, line 61: Do we really need (all of) these? E.g. at some point, we stored `k` explicitly, and we threw it out because we then always had multiple ways (namely `k` vs. `indices.len` to compute the very same thing), which just bloated the iterator.

src/combinations.rs
```diff
@@ -27,22 +27,70 @@ impl<I> fmt::Debug for Combinations<I>
...
+            reset_limit = prev_k;
+        }
+
+        // Reset old indices.
+        for i in 0..reset_limit {
+            self.indices[i] = i;
+        }
+
+        self.first = true;
+        filled
```
phimuemue, line 93: Nitpick: I think it would be nicer if we could go with fewer variables here:

Something like:
* `self.first=true // always done on reset`
* if `k<indices.len`, truncate and set items to `0,1,...,k`, return `true`
* otherwise `pool.prefill`, `indices.extend(indices.len..k)`, set first items to `0,1,...,k`, return `false`

This way, we would avoid the back-and-forth wrt `reset_limit`, and avoid the `mut filled`. As said, this seems more a stylistic thing.

src/combinations.rs
```diff
@@ -53,7 +101,7 @@ impl<I> Iterator for Combinations<I>
     type Item = Vec<I::Item>;
     fn next(&mut self) -> Option<Self::Item> {
         if self.first {
-            if self.pool.is_done() {
+            if self.pool.is_done() && self.k() > self.n() {
```
phimuemue, line 104: Could you elaborate why this became necessary?

phimuemue (commented):
Thanks for this! A powerset adaptor probably comes in very handy.

I sympathize with the idea of reusing `Combinations`, but I think it would be easier to avoid all the `inner` special cases.

willcrozi:
Hi, thanks for the detailed review!

> I sympathize with the idea of reusing `Combinations`, but I think it would be easier to avoid all the inner special cases.

Agreed, please see my replies to your specific comments.

It seems there's a few issues that weave together here.

What I'm thinking:
- Scrap the `Inner` enum and move to using increasing length `Combinations` from the outset, simplifying `Powerset`'s `next()` impl as you suggest.
- Decide between the following:
  - Keep `size_hint` impl (and therefore `k()`, `n()` and `src()` methods on `Combinations`
  - Scrap `size_hint` (removing `Powerset::pos`) and decide between:
    - Keeping `k()` and `n()` on `Combinations`
    - Add `done` field to `Powerset`
    - Some other way for `Combinations` to indicate a completed state?

Any idea on how to fix the the CI failures on jobs 2 and 3? (It gets stuck setting up `sccache` by the looks of it).

src/powerset.rs
willcrozi: Yes, I was torn on this and went with the 'inner' approach due to a performance benefit for use cases involving lots of small powersets. I think you're right, it's probably not worth the extra complexity.

src/powerset.rs
willcrozi: Originally it was inline in the impl of `Iterator::next()` but moved here to reduce indentation and improve readability.  We can scrap this anyway if we just use the increasing `Combinations` approach.

src/size_hint.rs
willcrozi: Yes! I forgot to check if it `saturating_pow` had hit stable since the last time I worked on this.

Thinking about this... would it make sense to make this more general and add a `base` argument? e.g. `pow(base: usize, exp: SizeHint)`. Not sure about naming though if we consider a complementary function that raises a base `SizeHint` to an exponent might be needed in the future?

Of course adding this at all will depend on whether we think having a `size_hint' impl for `Powerset` is worth the costs.

src/size_hint.rs
willcrozi: Yes also now in stable, will update.

src/combinations.rs
willcrozi: These are mainly a consequence of handing most of our information over to `Combinations`. We could remove `src()` at the expense of losing a decent `size_hint()` but I think `k()` and `n()` have more justification:

- `k()` and `n()` are used within `Powerset` to detect iterator completion. The alternatives I could see were a `Powerset::done` field (bloat) or just allowing a completed `Powerset` to keep increasing the size of `k` for its `Combinations` on every call to `next()` (wasteful, could allocate).
- A combination's `k` and `n` are also part of the formal notation for combinations in set-theory, so I thought it might make sense to have them visible.
- `k()` and `n()` are also necessary for the `size_hint()` impl.
- Could potentially make code more readable e.g. `self.indices.len()` vs `self.k()` 

I understand that we shouldn't jump through too many hoops for `size_hint` (and there are quite a few hoops here!), but since powersets can become large very quickly as input size increases, having one might be important for users.

src/combinations.rs
willcrozi: Agreed, I was attempting to avoid duplicating the 'reset loop' but your suggestion makes it a lot cleaner. I might be able to remove the need to return `bool` depending on the details of other changes.

src/combinations.rs
willcrozi: It seems that previously the combined condition `first && pool.done` served to indicate that `k` was larger than `n` and that no items would be yielded. With my changes this is no longer the case. Now when `Combinations` is reused internally within `Powerset` `pool.done` is always `true`, even during the first call to `next()`, since we are working with the same pool data.

There are other ways to do this, but this way seemed to make the most sense.

willcrozi:
I've simplified the `Powerset` implementation as per suggestions (with`k()` and `n()` still included for now).

The first additional commit excludes impl of `size_hint` and the second includes it for comparison.

I'll take some benchmarks when I get the chance to compare to the original approach.

src/lazy_buffer.rs
```diff
@@ -16,8 +16,7 @@ where
         LazyBuffer {
             it,
             done: false,
-            buffer: Vec::new(),
-        }
+            buffer: Vec::new() }
```
phimuemue, line 19: (Nitpick) Is this formatting change intentional?

src/powerset.rs
```diff
@@ -0,0 +1,85 @@
...
+
+                let result = self.combs.next();
+                if result.is_some() {
+                    self.pos = self.pos.saturating_add(1);
+                }
+
+                result
+            } else {
+                None
+            }
```
phimuemue, line 66: (Nitpick) Should we get rid of this level of indentation by using `else if self.combs.k()...` in line 54?

src/combinations.rs
phimuemue: Couldn't we then omit the `self.pool.is_done` check?

src/lazy_buffer.rs
```diff
@@ -44,6 +43,25 @@ where
...
     }
+
+    pub fn prefill(&mut self, len: usize) -> bool {
+        if self.done {
+            return self.buffer.len() >= len;
+        }
+
+        if len > self.buffer.len() {
+            let extra = len - self.buffer.len();
+            for _ in 0..extra {
```
phimuemue, line 54: Could we avoid `extra` and instead iterate over `self.buffer.len() .. len`?

src/powerset.rs
```diff
@@ -0,0 +1,85 @@
...
+        I::Item: Clone,
+{
+    type Item = Vec<I::Item>;
+
+    fn next(&mut self) -> Option<Vec<I::Item>> {
+        if let Some(elt) = self.combs.next() {
+            self.pos = self.pos.saturating_add(1);
+            Some(elt)
+        } else {
+            if self.combs.k() < self.combs.n() || self.combs.k() == 0 {
```
phimuemue, line 55: Is the `self.combs.k()==0` check necessary? If `k==0`, then line 51 should yield the empty subset (`[]`), and the next step should increase `k` to see if we need to continue - in particular also if `k<n`, right?

src/lazy_buffer.rs
```diff
@@ -44,6 +43,25 @@ where
             }
         }
     }
+
+    pub fn prefill(&mut self, len: usize) -> bool {
```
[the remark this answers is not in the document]
phimuemue: Possibly for another PR: Could we unify `prefill` and `get_next`?

phimuemue (commented):
Hi there! Thanks for addressing the points - imho this improved the PR quite a bit!

> Decide between the following:
> 
>     * Keep `size_hint` impl (and therefore `k()`, `n()` and `src()` methods on `Combinations`
> 
>     * Scrap `size_hint` (removing `Powerset::pos`) and decide between:
>       
>       * Keeping `k()` and `n()` on `Combinations`
>       * Add `done` field to `Powerset`
>       * Some other way for `Combinations` to indicate a completed state?

I think we could start out without `size_hint` - and if we really need it, possibly implement it for `Combinations` and compute `Powerset`'s `size_hint` in terms of `Combinations`' `size_hint`. However, let's wait for @jswrenn's opinion.

*As a side note:* Could you rebase your commits, so that they are easier to review? It would make subsequent reviews much easier. Maybe separate your work along the lines of:
* Pure formatting changes (if necessary)
* Typos
* Actual implementation
* Benchmarks

@jswrenn We are increasingly confronted with huge PRs that could be easily split into smaller ones. Should we encourage this somewhere in our guidelines? (Do we have such guidelines?)

src/lazy_buffer.rs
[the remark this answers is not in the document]
willcrozi: > I think the return value is not used anywhere, so we should possibly omit it for simplicity.

Yes, I've removed this.

> Possibly for another PR: Could we unify `prefill` and `get_next`?

Yes would be good. Maybe implementing one in terms of the other is the way to go?

src/lazy_buffer.rs
willcrozi: Oops, fixed!

src/powerset.rs
willcrozi: Good idea!

src/lazy_buffer.rs
willcrozi: yes, missed that!

src/powerset.rs
willcrozi: Yes it's necessary I think. `k == 0` is a special case (first element) where `k == n == 0` and so `k < n` is false. Without it we'd never try to increase `k`.

The alternative is to change the preceding part of the condition on line 55 to be `k <= n` which would have the effect of a pointless `Combinations` that has `k > n` at the end of a `Powerset`'s iteration.

Since it's the special case I've placed it last so that it get's short-circuited and split the condition across two lines for clarity.

willcrozi:
> I think we could start out without `size_hint` - and if we really need it, possibly implement it for `Combinations` and compute `Powerset`'s `size_hint` in terms of `Combinations`' `size_hint`. However, let's wait for @jswrenn's opinion.

Fair enough, though I'm not sure it's practical to compute a `size_hint` for Combinations since the formula involves three factorials: [`(n! / k!(n - k)!)`](https://en.wikipedia.org/wiki/Combination#Number_of_k-combinations). A powerset's length is much easier to compute directly: `2^n`.

> _As a side note:_ Could you rebase your commits, so that they are easier to review? It would make subsequent reviews much easier. Maybe separate your work along the lines of:
>
>     * Pure formatting changes (if necessary)
>     * Typos
>     * Actual implementation
>     * Benchmarks

No problem! I've split into two commits (impl and benchmarks). I've also made the doc comments more in line with the rest of the library.

src/combinations.rs
willcrozi: Yes, now removed.

src/combinations.rs
```diff
@@ -31,18 +31,56 @@ impl<I> fmt::Debug for Combinations<I>
 pub fn combinations<I>(iter: I, k: usize) -> Combinations<I>
     where I: Iterator
 {
-    let mut pool: LazyBuffer<I> = LazyBuffer::new(iter);
+    Combinations::new(iter, k)
+}
+
+impl<I: Iterator> Combinations<I> {
+    #[inline]
+    pub(crate) fn new(iter: I, k: usize) -> Self {
```
willcrozi, line 39: I'm thinking maybe this constructor should be put inline into combinations::combinations() function?

src/combinations.rs
```diff
@@ -31,18 +31,56 @@ impl<I> fmt::Debug for Combinations<I>
...
 
-    for _ in 0..k {
-        if !pool.get_next() {
-            break;
+        Combinations {
+            indices: (0..k).collect(),
+            pool,
+            first: true,
         }
     }
```
willcrozi, line 49: Maybe this constructor should be folded into the adaptor function above it?

src/powerset.rs
phimuemue: Ok. But now we have special case k>n only for n==k==0. I guess ideally we would just increment k unconditionally and let Combinations handle the case k>n - which it already does anyway.

The only caveat I could imagine with this simple approach are overflow related issues. Maybe @jswrenn can steer you in the right direction.

src/combinations.rs
```diff
@@ -53,7 +91,7 @@ impl<I> Iterator for Combinations<I>
     type Item = Vec<I::Item>;
     fn next(&mut self) -> Option<Self::Item> {
         if self.first {
-            if self.pool.is_done() {
+            if self.k() > self.n() {
```
jswrenn, line 94: Can you walk me through why this line changed?

src/combinations.rs
```diff
@@ -31,18 +31,56 @@ impl<I> fmt::Debug for Combinations<I>
...
-    Combinations {
-        indices: (0..k).collect(),
-        pool,
-        first: true,
+    /// Returns the length of a combination produced by this iterator. This value can
+    /// change between invocations of [`reset`].
+    ///
+    /// [`reset`]: #method.reset
+    #[inline]
+    pub fn k(&self) -> usize { self.indices.len() }
```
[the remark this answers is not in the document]
willcrozi: Good point. I'll remove the reference to it.

src/lazy_buffer.rs
```diff
@@ -44,6 +40,18 @@ where
             }
         }
     }
+
+    pub fn prefill(&mut self, len: usize) {
+        if !self.done && len > self.buffer.len() {
+            for _ in 0..(len - self.buffer.len()) {
```
phimuemue, line 46: could we just check for `!self.done` in line 45 and then iterate `for _ in len .. self.buffer.len()` in the next line, making the second check (`len > self.buffer.len()`) unnecessary?

src/powerset.rs
```diff
@@ -0,0 +1,56 @@
...
+impl<I> fmt::Debug for Powerset<I>
+    where I: Iterator + fmt::Debug,
+          I::Item: fmt::Debug,
+{
+    debug_fmt_fields!(Powerset, combs);
+}
+
+/// Create a new `Powerset` from a clonable iterator.
+pub fn powerset<I>(src: I) -> Powerset<I>
+    where I: Iterator,
```
phimuemue, line 31: Should this be `IntoIterator` instead of `Iterator`? (Afaik the other methods use `IntoIter`.)

phimuemue (commented):
Spotted some minor things to bring `powerset` in line with other methods, etc.

src/combinations.rs
willcrozi: Sure!

Previously, before my changes, the combined condition `first && pool.done` served to indicate insufficient items in the source iterator, to ever yield any elements from a `Combinations`. This worked because at construction of a `Combinations` we would ask for precisely `k` items and asking for the `k+1`th item would cause `pool.done` to be set.

With my changes we are reusing the `Combinations` (via the `reset()` method), and this reuse is 'triggered' by exhausting the source iterator. Therefore it's now possible for a `Combinations` to exist where all of the following hold:
  - source iterator is exhausted, i.e. `pool.done == true`
  - `k <= n == true`
  - we have not yielded any items yet, i.e. `first == true`

With the upshot being that `first && pool.done` no longer reliably indicates that we have no combinations to yield. 

Effectively `self.first && self.pool.done` was overloaded to mean `k < n` (i.e. 'no combinations are possible'), by relying on a particular way of constructing `Combinations`. Changing the second part of the condition to `k > n` means that we reliably detect an insufficient pool size once again.

willcrozi: Additionally, for me, the (new) code above at lines 55 and 56 just reads more clearly as 'do a one time check that we have a big enough pool to yield combinations of length `k`'

src/powerset.rs
willcrozi: `combinations` and all the others I've looked at so far have `I: Iterator` in the where clause

src/lazy_buffer.rs
willcrozi: You're right that the second check can be removed.

It would need to be for _ in self.buffer.len()..len though (range expressions reversed). Would this be as readable in your opinion?

src/powerset.rs
willcrozi: The thing stopping me from taking the approach of incrementing `k` unconditionally (i.e. pushing to `indices`) is that it would cause `indices` to allocate/copy unnecessarily when it's at capacity.

src/lazy_buffer.rs
jswrenn: > It would need to be for _ in self.buffer.len()..len though (range expressions reversed). Would this be as readable in your opinion?

I think `self.buffer.len()..len` is probably fine.

src/lazy_buffer.rs
```diff
@@ -44,6 +40,18 @@ where
...
+                if let Some(elt) = self.it.next() {
+                    self.buffer.push(elt);
+                } else {
+                    self.done = true;
+                }
+            }
+        }
+    }
```
jswrenn, line 54: Calling `.next()` in a loop performs very poorly for iterators like `Chain`. Could you avoid that by writing something like this, maybe?
```suggestion
    pub fn prefill(&mut self, len: usize) {
        let buffer_len = self.buffer.len();
        if !self.done && len > buffer_len {
            let delta = len - buffer_len;
            self.buffer.extend(self.it.by_ref().take(delta));
            self.done = self.buffer.len() == len;
        }
    }
```

willcrozi: > Calling `.next()` in a loop performs very poorly for iterators like `Chain`. Could you avoid that by writing something like this, maybe?

Interesting to know. I had been looking for a way of doing this but didn't think to combine `by_ref()` with `take()`. Useful technique!

I'll use this except the final line should be: `self.done = self.buffer.len() < len;` I.e. set `self.done` if the buffer wasn't filled to length `len`.

willcrozi:
Just to check what's the preferred approach when I've got changes based on PR review feedback:

- Add a new commit specifically addressing the feedback?
or
- Rebase then force-push, splitting review changes across commits according to category (typos, implementation, benchmarks etc)?

willcrozi:
I've pushed changes addressing the resolved feedback so far.

I've split the benchmarks into two commits and ordered them so that it's easier to check any effect from the changes to introduced to `Combinations`. I've also tweaked the doc comment for `Combinations::reset()` to state  _"If `k` is larger than the current length of the data pool an attempt is made to prefill..."_, which is more correct.

As far as I can see the outstanding issues are:
- Whether we perform the extra check for the special case of `k == 0` (see discussion [here](https://github.com/rust-itertools/itertools/pull/335#discussion_r495486069))
- Whether we want a `size_hint()` implementation (see discussion [here](https://github.com/rust-itertools/itertools/pull/335#discussion_r494832058) and [here](https://github.com/rust-itertools/itertools/pull/335#issuecomment-699572140))

Let me know what you think and if there are any more!

src/powerset.rs
jswrenn: Let's avoid the needless allocation.

jswrenn:
> Fair enough, though I'm not sure it's practical to compute a `size_hint` for `Combinations` since the formula involves three factorials: [`(n! / k!(n - k)!)`](https://en.wikipedia.org/wiki/Combination#Number_of_k-combinations). A powerset's length is much easier to compute directly: `2^n`.

I'm basically convinced by this that we should have a `size_hint` method. I don't see any harm in it, either.

jswrenn:
It's about time this gets merged. Worst case scenario, we can always fix issues in subsequent releases! Thanks for contributing this.

bors r+

jswrenn:
bors retry

jswrenn:
My attempt at making bors happy with the new CI isn't going well, so bear with me here...

willcrozi:
> It's about time this gets merged. Worst case scenario, we can always fix issues in subsequent releases! Thanks for contributing this.

No problem, great to see it finally merged!

Thanks @jswrenn and @phimuemue for the all the feedback and bearing with me. :+1:

PR #340, opened 2019-04-06
feat: add combinations_with_replacement

tommilligan:
As part of a personal project, I wanted to generate combinations of a set `n` of length `k`, but where elements may be repeated (i.e. may be present more than once in the output). I was not able to find an implementation of this in Rust anywhere - please correct me if I'm wrong!

The `CombinationsWithReplacement` iterator consumes a given `Iterator` of `Clone`-able items, and yields combinations of type `Vec<I::Item>` (the same api as `Combinations`). Suggestions for a more concise name welcome.

The algorithm is based on [this Stackoverflow question](https://stackoverflow.com/questions/127704/algorithm-to-return-all-combinations-of-k-elements-from-n) and [sample C code](https://stackoverflow.com/questions/561/how-to-use-combinations-of-sets-as-test-data#794)

### DIFF
--- a/benches/combinations_with_replacement.rs
+++ b/benches/combinations_with_replacement.rs
@@ -1,0 +1,34 @@
+#![feature(test)]
+
+extern crate itertools;
+extern crate test;
+
+use itertools::Itertools;
+use test::{black_box, Bencher};
+
+#[bench]
+fn comb_replacement_n10_k5(b: &mut Bencher) {
+    b.iter(|| {
+        for i in (0..10).combinations_with_replacement(5) {
+            black_box(i);
+        }
+    });
+}
+
+#[bench]
+fn comb_replacement_n5_k10(b: &mut Bencher) {
+    b.iter(|| {
+        for i in (0..5).combinations_with_replacement(10) {
+            black_box(i);
+        }
+    });
+}
+
+#[bench]
+fn comb_replacement_n10_k10(b: &mut Bencher) {
+    b.iter(|| {
+        for i in (0..10).combinations_with_replacement(10) {
+            black_box(i);
+        }
+    });
+}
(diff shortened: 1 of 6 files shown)


tommilligan:
Ping @jswrenn @phimuemue for review or suggestions

jswrenn:
Thanks for the contribution! My major point of reluctance in merging this is that it's not lazy: calling `combinations_with_replacement` on a cyclic iterator will never produce a new iterator. There's a strong expectation that iterator adaptors are lazy (and `Combinations` is) and that merely constructing them isn't a linear time operation.

tommilligan:
Good point, didn't consider this as it doesn't really make sense to get combinations for an infinite iterator. However I take your point, I've:
- moved `LazyBuffer` to it's own module
- updated `CombinationsWithReplacement` to use `LazyBuffer` internally
- added tests for using infinite iterators with `combinations` and `combinations_with_replacement` adaptors

jswrenn:
Looks good! Thanks!

bors r+

tommilligan:
It should, but it built successfully in CI with `cargo build --verbose --no-default-features`. I'm not sure why it wasn't caught.

jswrenn:
Well spotted, @pczarn. I'm not sure why this passed the CI, either. I'm tied up preparing for a conference right now, so I won't be able to investigate this for about another week, unfortunately.

tommilligan:
@pczarn I've just double checked the diff, and it is marked as `#[cfg(feature = "use_std")]` everywhere I can see - possibly just a GitHub UI issue yesterday? Unless you can point me at a specific line, I don't see anything that's incorrect.

pczarn:
It's fine - I viewed a single commit instead of the entire diff.

Issue #282, opened 2018-05-17
performance of all_equal


RSabet:
Here is a comparison of `itertools::all_equal` with two alternatives for a vector consisting of 5_000_000 zeroes followed by 5_000_000 ones on rust playground:

- using `itertools::all_equal`: 607 msecs
- using `any`: 487 msecs
- using a simple for loop: 536 msecs

i would love if `all_qual` was as fast

[playground](https://play.rust-lang.org/?gist=06e8a79766f7f51c6ce73898982a70b9&version=stable&mode=debug)

bluss:
@RaminSabet That's a very reasonable concern. Note that we must use Release mode compiles to make meaningful performance comparisons. I'm not sure you'll have a different conclusion with that, but you'll have faster timings across the board, and the code will need to be fixed to not optimize out some parts of the benchmark.

RSabet:
You are right, trying to hide the values in the vector from the optimizer one can use:

    let mut v: Vec<i32> = (0..5_000_000).map(|_| rng.gen_range(1, 2)).collect();
    v.extend((0..5_000_000).map(|_| rng.gen_range(2, 3)) .collect::<Vec<i32>>());
again 5_000_000 ones followed by 5_000_000 twos

and now in release mode:

- using `itertools::all_equal`: 5 msecs
- using `any`: 3 msecs
- using a simple for loop: 0 msecs (guess the compiler still found a way to optimize)

[playground](https://play.rust-lang.org/?gist=9c0e122d555a1f080996345b2cce9540&version=stable&mode=release)

timvermeulen:
`all_equal`'s implementation is currently suboptimal:
```rust
fn all_equal(&mut self) -> bool
    where Self::Item: PartialEq,
{
    self.dedup().nth(1).is_none()
}
```
This doesn't necessarily iterate the entire iterator, but in @RSabet's example code, it does evaluate all the 2s.

Philippe-Cholet:
Solved by #342.

PR #342, opened 2019-04-26
make all_equal() faster

fyrchik:
Hello!
This PR adresses #282 issue. Variant with `dedup` does not short circuit, but one with `all` does.
I have also added some benchmarks and test for an empty iterator.
```
test all_equal                                ... bench:     999,832 ns/iter (+/- 217,245)
test all_equal_default                        ... bench:   4,814,277 ns/iter (+/- 315,335)
test all_equal_for                            ... bench:   2,096,174 ns/iter (+/- 165,596)
```

Let me know, what do you think.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1333,9 +1333,13 @@ pub trait Itertools : Iterator {
     /// assert!(data.into_iter().all_equal());
     /// ```
     fn all_equal(&mut self) -> bool
-        where Self::Item: PartialEq,
+        where Self: Sized,
+              Self::Item: PartialEq,
     {
-        self.dedup().nth(1).is_none()
+        match self.next() {
+            None => true,
+            Some(a) => self.all(|x| a == x),
+        }
     }
 
     /// Consume the first `n` elements from the iterator eagerly,
--- a/tests/test_std.rs
+++ b/tests/test_std.rs
@@ -116,6 +116,7 @@ fn dedup_by() {
 
 #[test]
 fn all_equal() {
+    assert!("".chars().all_equal());
     assert!(!"AABBCCC".chars().all_equal());
     assert!("AAAAAAA".chars().all_equal());
     for (_key, mut sub) in &"AABBCCC".chars().group_by(|&x| x) {
--- a/tests/test_std.rs
+++ b/tests/test_std.rs
@@ -117,6 +117,7 @@ fn dedup_by() {
 #[test]
 fn all_equal() {
     assert!("".chars().all_equal());
+    assert!("A".chars().all_equal());
     assert!(!"AABBCCC".chars().all_equal());
     assert!("AAAAAAA".chars().all_equal());
     for (_key, mut sub) in &"AABBCCC".chars().group_by(|&x| x) {
--- a/benches/bench1.rs
+++ b/benches/bench1.rs
@@ -731,3 +731,34 @@ fn cartesian_product_nested_for(b: &mut test::Bencher)
         sum
     })
 }
+
+#[bench]
+fn all_equal(b: &mut test::Bencher) {
+    let mut xs = vec![0; 5_000_000];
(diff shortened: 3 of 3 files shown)


timvermeulen:
Looks good!

axelf4:
An alternative implementation could be:

```rust
self.try_fold(None, |acc, x| acc.filter(|&prev| x != prev).xor(Some(Some(x))))
```

timvermeulen:
@axelf4 I haven't done any benchmarks, have you by any chance? I'd be worried that that would perform worse than this PR's implementation because of the extra branching in the closure body. But it's very hard to tell without benchmarks.

src/lib.rs
```diff
@@ -1310,9 +1310,13 @@ pub trait Itertools : Iterator {
     /// assert!(data.into_iter().all_equal());
     /// ```
     fn all_equal(&mut self) -> bool
-        where Self::Item: PartialEq,
+        where Self: Sized,
```
phimuemue: I think requiring `Sized` is ok. Can anyone with more expertise confirm this is unproblematic?

tests/test_std.rs
```diff
@@ -100,6 +100,7 @@ fn dedup() {
 
 #[test]
 fn all_equal() {
+    assert!("".chars().all_equal());
```
phimuemue: Good one.

src/lib.rs
timvermeulen: There are already lots of methods you can't call on an unsized iterator, so I don't think this is problematic.

timvermeulen: I believe the documentation (sometimes?) leaves out `Self: Sized` bounds because they’re so common 😕

timvermeulen: I believe the documentation (sometimes?) leaves out `Self: Sized` bounds because they’re so common 😕

jswrenn: Per [Reddit](https://www.reddit.com/r/rust/comments/cgh56g/hey_rustaceans_got_an_easy_question_ask_here/eukw73y/?context=1&depth=3), I think this is a bug. I've filed an issue: https://github.com/rust-lang/rust/issues/62899.

tests/test_std.rs
fyrchik: Yes, it will be a nice addtition. I have added it.

bluss: Adding a quickcheck test would be a great alternative to either of those (something for another PR)

bluss:
This should not need to require Self: Sized, and I'd prefer to rewrite it so that it is not a breaking change (in my opinion).

I'm not sure I understand the short circuit argument - in what sense does this short circuit more than dedup does? They should both visit the same number of iterator elements.

timvermeulen:
@bluss The `dedup` version does also short-circuit, but only after the second "group" has been iterated entirely, rather than right after the first element of the second group.

bluss:
Oh. We are long overdue for this improvement then.

One major part of benchmark speed up could be that `all` is explicitly unrolled in the slice iterator (slice iterator specific, in its `try_fold`). That's nice to use, but we can note that it is not a general improvement. In addition to that we have the improvement of using internal iteration (all/try_fold etc) which is a bit more general.

Looking at those upsides it seems ok to have a breaking change. I'm not sure in what context it would be a detectable breaking change, it should only change which implementation is picked when we call this method on a trait object.

... and it turns out the trait object question is moot, because trait object support has been broken in Itertools 0.8.0, `dyn Itertools` is not possible at the moment precisely because of some missing Self: Sized bounds.

That's something we could revisit, but let's only add back trait object support if we can find a good reason. Clearly there was no test for it.

With that, I don't think adding `Self: Sized` is a breaking change.

jswrenn:
If the trait object question is moot, I'm happy to merge this!

bors r+

PR #344, opened 2019-05-09
adding trait CopyIntoSlice

frehberg:
Related to issue https://github.com/bluss/rust-itertools/issues/343

Given an iterator or chain, this methods permits direct unrolling into a slice.

The method "copy_into_slice" will return the number of elements having been copied/unrolled into the slice. If the capacity of the slice is exceeded, the method will return with Error::EndOfSlice

### DIFF
--- a/src/collect_into_slice.rs
+++ b/src/collect_into_slice.rs
@@ -7,13 +7,13 @@ pub enum Error {
     EndOfSlice
 }
 
-/// Trait copying all iterator's elements into one slice, returning number of elements
+/// Trait collectin all iterator elements into one slice, returning number of elements
 ///
 /// The method will return Error::EndOfSlice if slice is exceeded
 ///
 /// # Examples
 /// ```
-///    use itertools::CopyIntoSlice;
+///    use itertools::CollectIntoSlice;
 ///
 ///    let req_method_get = b"GET ";
 ///    let req_method_head = b"HEAD ";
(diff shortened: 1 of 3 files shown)


frehberg:
> "Progress is the way from the primitive via the complicated towards the simple."
[Wernher von Braun]

Saying this: Instead of the newly defined Iterator method, I found a better, more simple solution to unroll chained byte-sequences into a byte-buffer, just based on common functionality. 

 ```rust
   let req_method_get = b"GET ";
    let req_method_head = b"HEAD ";
    let uri = b"/infotext.html HTTP/1.1";

    let mut request = [0 as u8; 128];

    // chaining various byte-arrays together, for example two here:
    let mut chained_byte_sequence =  req_method_get.iter()
        .chain(uri.iter());
    
    // take a ref and zip against the destination buffer and while iterating via fold, each 
    // element is counted.
    let  nwritten =
        chained_byte_sequence
        .by_ref()
        .zip(request.as_mut())
            .fold(0, |cnt, (item, slot) | {
                *slot = item.clone(); cnt+1
            });

    // finally, the iterator of the chained_byte_sequence must be empty
    if chained_byte_sequence.next()!=None {
        /* slice too short */ panic!();
    } else {
        println!("{:?}", &request[0..nwritten].as_ref());
    };
```
Thanks to everybody, commenting. I think is the best, most flexible solution and good code-sample. I am very happy with this solution.

AFAICS the traits CopyIntoSlice or CollectIntoSlice would be redundant! 
This ticket is obsolet and I will close it

PR #346, opened 2019-05-28
Implement DoubleEndedIterator for Zip

svenknobloch:
Implements the DoubleEndedIterator trait for Zip. Based on the standard library's implementation.

### DIFF
--- a/tests/zip.rs
+++ b/tests/zip.rs
@@ -75,5 +75,3 @@ fn zip_eq_panic2()
 
     zip_eq(&a, &b).count();
 }
-
-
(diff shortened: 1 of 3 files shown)


tests/zip.rs
```diff
@@ -42,6 +43,20 @@ fn test_double_ended_zip_longest() {
     assert_eq!(it.next(), None);
 }
 
+#[test]
+fn test_double_ended_zip() {
```
jswrenn: Could you formulate this as a `quickcheck` test?

I'd like to see that for any two inputs, `multizip((a, b)).collect_vec().rev()` is equal to (something like) `multizip((a, b)).rfold(Vec::new(), |vec, e| { vec.push(e); vec })`

PR #350, opened 2019-06-15
Add circular_tuple_windows

ed-bassett:
As described in https://github.com/bluss/rust-itertools/issues/349, I think `circular_tuple_windows` would be broadly useful enough to be added to Itertools. It relies on `tuple_windows` and cloning, so I'm sure it could not be included in the standard library.

I'm not sure if this is the best way to implement the function, but I thought I would give it a shot so that at the very least you'd have something functional to work from.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -127,7 +127,7 @@ pub mod structs {
     pub use crate::sources::{RepeatCall, Unfold, Iterate};
     #[cfg(feature = "use_std")]
     pub use crate::tee::Tee;
-    pub use crate::tuple_impl::{TupleBuffer, TupleWindows, Tuples};
+    pub use crate::tuple_impl::{TupleBuffer, TupleWindows, CircularTupleWindows, Tuples};
     #[cfg(feature = "use_std")]
     pub use crate::unique_impl::{Unique, UniqueBy};
     pub use crate::with_position::WithPosition;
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -583,6 +583,40 @@ pub trait Itertools : Iterator {
...
+    /// assert_eq!(Some((3, 4, 1)), it.next());
+    /// assert_eq!(Some((4, 1, 2)), it.next());
+    /// assert_eq!(None, it.next());
+    ///
+    /// // this requires a type hint
+    /// let it = (1..5).circular_tuple_windows::<(_, _, _)>();
+    /// itertools::assert_equal(it, vec![(1, 2, 3), (2, 3, 4), (3, 4, 1), (4, 1, 2)]);
+    /// ```
+    fn circular_tuple_windows<T>(self) -> CircularTupleWindows<Self, T>
+        where Self: Sized + Clone + Iterator<Item = T::Item> + ExactSizeIterator,
```
jswrenn: Are there any issues with eliminating the `ExactSizeIterator` bound?

ed-bassett: I added `ExactSizeIterator` so that the call to `len()` on line 198 on in `tuple_impl` would be performant. Using `tuple_windows(iter.cycle()).take(len)` was a neat way of getting the extra elements wrapped onto the end, but if removing the `ExactSizeIterator` constraint is preferable we could potentially go with something like the following, at the expense of a slight reduction in readability:
```
pub struct CircularTupleWindows<I, T: Clone>
    where I: Iterator<Item = T::Item> + Clone,
          T: TupleCollect + Clone,
          <T as TupleCollect>::Item: Debug
{
    iter: TupleWindows<Chain<Chain<IntoIter<<T as TupleCollect>::Item>, I>, IntoIter<<T as TupleCollect>::Item>>, T>,
}

pub fn circular_tuple_windows<I, T>(mut iter: I) -> CircularTupleWindows<I, T>
    where I: Iterator<Item = T::Item> + Clone,
          T: TupleCollect + Clone,
          T::Item: Clone,
          <T as TupleCollect>::Item: Debug
{
    let first_elements = iter.by_ref().take(<T as TupleCollect>::num_items() - 1).collect::<Vec<_>>();
    let iter = tuple_windows(first_elements.clone().into_iter().chain(iter).chain(first_elements.into_iter()));

    CircularTupleWindows {
        iter: iter,
    }
}

impl<I, T> Iterator for CircularTupleWindows<I, T>
    where I: Iterator<Item = T::Item> + Clone,
          T: TupleCollect + Clone,
          T::Item: Clone,
          <T as TupleCollect>::Item: std::fmt::Debug
{
    type Item = T;

    fn next(&mut self) -> Option<T> {
        self.iter.next()
    }
}
```

what do you think?

Boscop:
Any update on this? :)

ed-bassett:
> Any update on this? :)

Just waiting on feedback from @jswrenn on which way to go regarding the `ExactSizeIterator` bound

jswrenn:
Sorry for dropping the ball on this!

I think I can merge this as-is. There's no reason we can't drop the bound later if it's unduly prohibitive.

bors r+

ed-bassett:
Any chance I can squash in the fixup before it's merged? I was holding off while it was waiting on review, but I don't really know the process around getting things merged

jswrenn:
Sure! Just let me know when you're ready.

ed-bassett:
Good to go

PR #360, opened 2019-07-24
Simplify Merge/MergeBy

phimuemue:
This is a follow-up on https://github.com/rust-itertools/itertools/pull/357#issuecomment-514685494.

* I removed `struct Merge` and introduced `type Merge<I, J>=MergeBy<I, J, MergeLte>`. To do this, I had to adjust `merge_by_new` a bit. `merge_by_new` is not exported by the crate, so I hope this is fine.
* Then, I inlined `MergeCore` into `MergeBy`, because its sole purpose seemed to be to supply the guts for `next` and `size_hint`.

I did not change generic parameter names in those commits. (We may think about this in subsequent steps.)

### DIFF
--- a/src/adaptors/mod.rs
+++ b/src/adaptors/mod.rs
@@ -463,64 +463,6 @@ impl<I> ExactSizeIterator for Step<I>
     where I: ExactSizeIterator
 {}
 
-
-struct MergeCore<I, J>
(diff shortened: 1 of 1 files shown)


jswrenn:
Thanks!

bors r+

PR #363, opened 2019-08-04
Implement unique_combinations.

meltinglava:
Some questions that needs to be answared:
* Document differently?
* Any optimalisations?
* The algorithm only requires equal Items to be grouped beside each
  other. This is achived by sorting. However if we there is an
  effisient way to group equals then we can have this require `Eq`
  insted of `Ord`


src/unique_combinations.rs
```diff
@@ -0,0 +1,112 @@
...
+    }
+}
+
+impl<I> Iterator for UniqueCombinations<I>
+where
+    I: Iterator,
+    I::Item: Clone + Ord,
+{
+    type Item = Vec<I::Item>;
+    fn next(&mut self) -> Option<Self::Item> {
```
phimuemue: I may be missing something, but shouldn't `next` do something very similar to `Combinations::next`? (I.e. incrementing the rightmost index, possibly skipping values if the referenced values would lead to duplicates.)

If so, could we implement it similarly to `Combinations::next`?

tests/test_std.rs
```diff
@@ -575,6 +576,17 @@ fn combinations() {
         vec![3, 4],
         ]);
 
+    let it = vec![1, 2, 2, 3, 4].into_iter().unique_combinations(3);
```
phimuemue: Should we assert that `unique_combinations` yields the "same" (modulo ordering) elements that would be obtained by manually filtering elements from `combinations`?

src/unique_combinations.rs
```diff
@@ -0,0 +1,112 @@
...
+/// An iterator to iterate through all the `n`-length combinations in an iterator.
+///
+/// See [`.unique_combinations()`](../trait.Itertools.html#method.unique_combinations) for moref information.
+#[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
+pub struct UniqueCombinations<I: Iterator> {
+    len: usize,
+    position: Vec<usize>,
+    pool: Vec<I::Item>,
+    first: bool,
+    next_none: bool,
```
phimuemue: We sometimes have members with similar semantics. They are usually called `done`. Should we call this one `done`, as well?

src/lib.rs
```diff
@@ -1153,24 +1157,39 @@ pub trait Itertools : Iterator {
...
     /// ```
     /// use itertools::Itertools;
     ///
-    /// let it = vec![1, 2, 2].into_iter().combinations(2);
+    /// let it = vec![1, 2, 2].into_iter().unique_combinations(2);
     /// itertools::assert_equal(it, vec![
-    ///     vec![1, 2], // Note: these are the same
-    ///     vec![1, 2], // Note: these are the same
+    ///     vec![1, 2],
     ///     vec![2, 2],
```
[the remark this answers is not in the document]
meltinglava: Think this one got removed with the rebase, thow now might be the time to have them there. I am not sure on the matter.

src/unique_combinations.rs
meltinglava: The difference is that combinations only bumps based on the value of the indices, and not the values of the pool.

meltinglava:
I did a rebase so diffs should now make more sence to the current codebase.

There have happend a lot of changes to combinations since the initial release of this pr, so we should see if we can use some of the techniques applied to combinations. 

As of the LazyBuffer type, i think if this should be merged with unique combinations, this is to major of a rewrite and think it should be a separate pr if so. 

Thanks @phimuemue for reviewing. 
As for the last ones I have not marked as resolved. Since the rebase and all of the refer to combinations, can you check if they still apply.

src/unique_combinations.rs
meltinglava: I agree that the first `if` and the `else if` should look the same. However the example there with the 2 `loop`-s that are just used as `jump` statements. I find that code a lot harder to read. 

My version shows the difference that either you bump the back index, or you find what index that can be bumped and sets the next ones after that index to the values following that index. 
If you trace your `if`-s you will see that they mostly run in one of these cases rarely bought (from what i found it was just to exit the `'outer` loop). Lines like line 35 is only useful in bumping the back number, as it will be overwritten in the `for` loop in the other cases.

tests/test_std.rs
meltinglava: I think that this will not always be true, as combinations explore all combinations before getting a new number for the iterator. We see all the numbers as fast as possible as we want to find the end. This is why the `LasyBuffer` is used. to not need that many next calls from the iterator.

src/unique_combinations.rs
phimuemue: Hi there! Thanks a lot for bringing the `if`s in line - I really appreciate that and I think it looks better. Also very nice that you got rid of the `done` field.

I really try to grasp the algorithm and understand the distinct cases. Maybe you can help me with the following questions:

* Lines 77/78 (`lastpos` and `val`) seem to be analogous to lines 93/94 (`i` and `current`). If this is the case, would it be more readable if we called them the same, used the same logic wrt subtracting 1 from `i`/`lastpos`, and avoided the `mut`ability in the second case?
* Related: Is `for j in lastpos + 1..pool_len {...}` (line 79 and following) analogous to `while next == current` (line 96 and following)? If so, should we implement them in a similar fashion?

If all of the above are true:
* Do we then have the situation that the `else`-branch (starting at line 91) is basically a special case (avoiding setting the next one's indices) of the more general case (starting at 75)? If this is the case, would it make sense to leave out this special case (as `Combinations::next` does)?

tests/test_std.rs
phimuemue: I am not sure if I understand you correctly. But shouldn't `unique_combinations` give the same elements that you would get by manually removing duplicates from `combinations` (possibly sorting the yielded permutations)?

meltinglava:
i have autoformat in my editorconfig on save. I can create an pr with formating for all the files.

src/unique_combinations.rs
meltinglava: first point:
They look alike but the line 77 one has a i that is a count to count backwards. The other i is the position in the pool that contains the next item that is to be.I think we would get some benefit of calling them what they actually are, instead of something that is ambiguous.

phimuemue: Hi again. I took your PR and did some syntactic adjustments (see [here](https://github.com/meltinglava/itertools/compare/ed0de3e383e09519cce7eeb6d141e48509771cbe...phimuemue:uniq_comb) for the individual steps).

In the end, I arrived at something where both cases actually use the same code, so we may be able to use similar code, and possibly could avoid special cases here.

tests/test_std.rs
meltinglava: This is true as long as the iterator that combinations is done over is sorted in the first place. Might be a good test with quicksort.

src/unique_combinations.rs
meltinglava: I found the solution now. Thanks for the branch.

src/lib.rs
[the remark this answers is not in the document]
meltinglava: We might create a table with all types of combinations that shows the same iterator and different output for each of them.

tests/test_std.rs
meltinglava: After some thinking is think we should make thise, however without garanteeing the order of the items that comes out in either cases. So sort bought on input and output. Will implement that this weekend

meltinglava:
@phimuemue is there any other blockers than getting formating done on master for this PR (besides the unit testing)?

phimuemue:
> @phimuemue is there any other blockers than getting formating done on master for this PR (besides the unit testing)?

As always, @jswrenn has the last word. But I still think that the implementation could/should be closer to Combinations::next.

I tried to see if my intuition is wrong, but taking your implementation and transforming piece by piece leads to something very similar to Combinations::next (see my sketch https://github.com/meltinglava/itertools/compare/master...phimuemue:uniq_comb2). (One thing I like about Combinations::next is that the early-outs are only for the `None` case, while your implementation has one for `generate` and one for `None`.)

tests/test_std.rs
```diff
@@ -625,20 +680,44 @@ fn combinations_with_replacement() {
...
         (0..0).combinations_with_replacement(2),
         <Vec<Vec<_>>>::new(),
     );
+    it::assert_equal((1..3).unique_combinations(0), vec![vec![]]);
+}
+
+#[test]
+fn combinations_and_unique_combinations_has_all_unique_values() {
+    let mut rng = &mut rand::thread_rng();
+    let a = [1, 2, 3, 4, 5];
```
phimuemue: Cool to have this kind of test. Just not sure if having non-determinism in tests is such a good idea, especially if the random test is only run once. (I may be wrong on this one though, if we already have precedence for this kind of test.)

meltinglava: My biggest problem here was that i did not find out how you made sure quickcheck made vectors that was garanteed to contain duplicates.

meltinglava:
> One thing I like about Combinations::next is that the early-outs are only for the `None` case, while your implementation has one for generate and one for `None`

I agree to the theory here, thow technicly the `break` statment here is just sugar for returning in that case. The loops job in all the cases are searching, hence break out / returning when found. Do not find the other one to be drasticly better than the other one. 

The reason I am hessitent to do it that way, is that I do not feel like the code is better in terms of readability and logic. If I am not the only one that think this is true, then it is my opinion that `combinations` should be brought closer to whatever `unique_combinations` end up looking like if this is a goal that we want. 

I have found that it takes quite alot of time to figgure out what an algorithm does when its all based on one mutable number that deals with all the cases.

Here are the pros and cons of each as i see it.

Pros unique:
* No `mut` variables used.

Cons:
* At the moment not the looking like `combinations` (atm becasue this can change in the future)

Neither one:
* Same amount of early_`return` + `break` + `continue` statements (4 in bought cases)

Do anyone have anything to add?

phimuemue:
Thanks for your feedback.

> > One thing I like about Combinations::next is that the early-outs are only for the `None` case, while your implementation has one for generate and one for `None`
> 
> I agree to the theory here, thow technicly the `break` statment here is just sugar for returning in that case. The loops job in all the cases are searching, hence break out / returning when found. Do not find the other one to be drasticly better than the other one.

Regarding the `break`: If we follow my sketch we should get rid of the break here, by actually exploiting what we know: As stated in line 74

    if self.pool[bump_source] < self.pool[bump_target] { // must be true for at least one bump_target

we know that the loop initiated in line 73 (`for bump_target in bump_source + 1..pool_len`) will not be exhausted without success. This is because line 63 (`self.pool[self.indices[i]] == self.pool[i + pool_len - indices_len]`) already established an element for whith the loop from line 73 will terminate. We could incorporate this information in the loop of line 73 to get rid of the `break`, so that it reads more like "find bump_target, increment `indices(i)` to it, and adjust the remaining `indices`".

> 
> The reason I am hessitent to do it that way, is that I do not feel like the code is better in terms of readability and logic. 

You may have a valid point there: I may be biased (because I just compared your new code to the existing one), so another opinion on this could very well inform our choice there.

> If I am not the only one that think this is true, then it is my opinion that `combinations` should be brought closer to whatever `unique_combinations` end up looking like if this is a goal that we want.

I agree: If `UniqueCombinations` is found to be the simpler variant, and `Combinations` would profit from this, we should change it.

> 
> I have found that it takes quite alot of time to figgure out what an algorithm does when its all based on one mutable number that deals with all the cases.

You're right: These algorithms are surprisingly tricky to formulate in simple terms. But isn't the essence as follows:

* Starting at the end, find an index `i` in `indices` that can be incremented.
  (In `Combinations`, we have `self.indices[i] == i + self.pool.len() - self.indices.len()`, whereas in `UniqueCombinations` we could have essentially the same "tunneled through `self.pool`": `self.pool[self.indices[i]] == self.pool[i + pool_len - indices_len]`).

* Once we have found this index, increment `index(i)`.
  (In `Combinations` this is simply `self.indices[i] += 1;`, while in the `UniqueCombinations` we could derive a variant without breaks/early-outs that achieves something that increments `self.indices(i)` to something so that the referenced `pool` value changes.)

* Increment the indices after `i`. For this I would expect the code to be the same in both variants.

I think current `Combinations` resembles this logic quite well. Do you have another (possibly simpler) view on the problem?

meltinglava:
Just thought how hard it would be doing this with rust iterators. It is qute descriptive if when comments are added. Here we have almost all of combinations, thow i could not get the .iter_mut to return mutable referances. Do also think what I did inside the for loop is wrong. thow this is just an example.
```rust
// search for bumpable index
return self
    .indices
    .iter()
    .rev()
    .enumerate()
    .find(|(offset, &index)| self.pool[index] != self.pool[pool_len - 1 - offset])
    .map(|(offset, &index)| (offset, index))
    .clone()
    // when found bump that index and increase the following values
    .map(|(offset, index)| {
        for (o, &mut i) in self
            .indices
            .iter_mut()
            .rev()
            .take(offset + 1)
            .rev()
            .enumerate()
        {
            *i = index + offset + o;
        }
    })
```
What i like about this version is that its only the`find` function does exactly what it says, find the value that fits the predicate, and it does the `break` / early `return` handeling

PR #364, opened 2019-08-13
Express KMerge in terms of KMergeBy

phimuemue:
This is a follow up to https://github.com/rust-itertools/itertools/pull/360:
* First, I brought `KMergeBy` in line with `KMerge` by `impl Clone` (if possible).
* Then, I substitute `KMerge` by a specialized `KMergeBy`.
* At last, I express `kmerge` in terms of `kmerge_by`.


jswrenn:
Thanks!

bors r+

PR #369, opened 2019-09-12
add rustfmt support to CI

### LINKED ISSUE #368: Consider applying rustfmt to the code?
It seems that the code here doesn't follow the [Rust style guide](https://github.com/rust-dev-tools/fmt-rfcs/blob/master/guide/guide.md). Maybe consider applying rustfmt and add CI step to enforce it?

I think that would make people more comfortable to contribute, as they wouldn't need to think about what code style should they use, and can just rely on the tool to figure out.

SirWindfield:
This PR adds support for `rustfmt`. The CI script installs the binary on the stable release only and checks if the code is formatted correctly.

Closes #368.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,11 @@ sudo: false
 matrix:
   include:
     - rust: 1.24.0
-    - rust: stable
+    - name: Rust stable
+      rust: stable
+      before_script:
+        - rustup component add rustfmt
+        - cargo fmt -- --check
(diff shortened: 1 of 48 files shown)


jswrenn:
I appreciate the contribution! I'm going to close this for now per the rationale in https://github.com/rust-itertools/itertools/issues/368#issuecomment-536741166.

We will definitely revisit rustfmt in the future.

PR #378, opened 2019-10-30
Use `slice::iter` instead of `into_iter` to avoid future breakage

LukasKalbertodt:
`an_array.into_iter()` currently just works because of the autoref
feature, which then calls `<[T] as IntoIterator>::into_iter`. But
in the future, arrays will implement `IntoIterator`, too. In order
to avoid problems in the future, the call is replaced by `iter()`
which is shorter and more explicit.

A crater run showed that your crate is affected by a potential future change. See https://github.com/rust-lang/rust/pull/65819 for more information.

### DIFF
--- a/benches/bench1.rs
+++ b/benches/bench1.rs
@@ -688,7 +688,7 @@ fn multi_cartesian_product_iterator(b: &mut test::Bencher)
 
     b.iter(|| {
         let mut sum = 0;
-        for x in xs.into_iter().multi_cartesian_product() {
+        for x in xs.iter().multi_cartesian_product() {
             sum += x[0];
             sum += x[1];
             sum += x[2];
@@ -704,7 +704,7 @@ fn multi_cartesian_product_fold(b: &mut test::Bencher)
 
     b.iter(|| {
         let mut sum = 0;
-        xs.into_iter().multi_cartesian_product().fold((), |(), x| {
+        xs.iter().multi_cartesian_product().fold((), |(), x| {
             sum += x[0];
             sum += x[1];
             sum += x[2];


LukasKalbertodt:
Not to be rushing anyone, but would it be possible to release this as 0.8.2? Lots of crates are depending on `itertools` and having this fix out in the wild as soon as possible would be a good thing :)

jswrenn:
Done! Sorry for the delay.

Since the affected code is part of a benchmark, would this breakage actually affect library users?

LukasKalbertodt:
Oh gosh, I'm so very sorry! As I tried fixing as many crates as possible, I seem to have paid too little attention to what I have been changing. You are right, the benchmark should not affect users.

Most failures looked [like this](https://crater-reports.s3.amazonaws.com/pr-65819/try%23b34fa46cc6fa5d46c6adcb041ebd28c7f7a374f3/gh/ejpcmac.diceware/log.txt):

```
[INFO] [stderr] error[E0631]: type mismatch in function arguments
[INFO] [stderr]    --> /opt/rustwide/cargo-home/registry/src/github.com-1ecc6299db9ec823/proptest-0.7.0/src/arbitrary/_std/io.rs:131:23
[INFO] [stderr]     |
[INFO] [stderr] 131 |     ].into_iter().map(Clone::clone).map(Just))
[INFO] [stderr]     |                       ^^^^^^^^^^^^
[INFO] [stderr]     |                       |
[INFO] [stderr]     |                       expected signature of `fn(std::io::ErrorKind) -> _`
[INFO] [stderr]     |                       found signature of `for<'r> fn(&'r _) -> _`
[INFO] [stderr] 
[INFO] [stderr] error[E0599]: no method named `map` found for type `core::iter::Map<core::array::IntoIter<std::io::ErrorKind, 18usize>, for<'r> fn(&'r _) -> _ {<_ as core::clone::Clone>::clone}>` in the current scope
[INFO] [stderr]    --> /opt/rustwide/cargo-home/registry/src/github.com-1ecc6299db9ec823/proptest-0.7.0/src/arbitrary/_std/io.rs:131:37
[INFO] [stderr]     |
[INFO] [stderr] 131 |     ].into_iter().map(Clone::clone).map(Just))
[INFO] [stderr]     |                                     ^^^ method not found in `core::iter::Map<core::array::IntoIter<std::io::ErrorKind, 18usize>, for<'r> fn(&'r _) -> _ {<_ as core::clone::Clone>::clone}>`
[INFO] [stderr]     |
[INFO] [stderr]     = note: the method `map` exists but the following trait bounds were not satisfied:
[INFO] [stderr]             `&mut core::iter::Map<core::array::IntoIter<std::io::ErrorKind, 18usize>, for<'r> fn(&'r _) -> _ {<_ as core::clone::Clone>::clone}> : core::iter::Iterator`
[INFO] [stderr]             `core::iter::Map<core::array::IntoIter<std::io::ErrorKind, 18usize>, for<'r> fn(&'r _) -> _ {<_ as core::clone::Clone>::clone}> : core::iter::Iterator`
[INFO] [stderr] 
```

It's in proptest 0.7.0, so this particular case might already be fixed. I will investigate in a few hours. I saw your comment shortly before I had to leave for an appointment. Sorry again for the confusion and the fairly useless release that I asked for. I will comment here once I know more.

LukasKalbertodt:
Okay, I have understood the situation. Finally. 

My root regression analysis never actually checked how many regressions a particular root regression caused, but just sorted by reverse dependencies. I just adjusted the analysis to account for that (which I should have done already some time ago). And it turns out: no regressions are caused by `itertools`. So yes, it was just in the benchmark and at no time, users of this crate were affected by this. So we are done here.

Again, please excuse my hastily actions and that I asked you to release a new version without properly checking if it is worth anything. I'm very sorry and try to be more careful in the future!

PR #380, opened 2019-11-05
Combination for zero sized array

### LINKED ISSUE #361: Combination on zero-sized array
Combination(0) seems to return an empty array.
However, the result differs when the original list is empty or not.

I think it would be better to produce [] instead of None.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=dad32c5aeae27dcd9b14ce8475b5fa79

```{.rs}
use itertools::*; // 0.8.0
fn main() {
    let a: Vec<usize> = vec![1usize];
    let b: Vec<usize> = vec![];
    println!("{:?}", a.iter().combinations(0).next()); // Some([])
    println!("{:?}", b.iter().combinations(0).next()); // None
}
```

https://www.wolframalpha.com/input/?i=0+choose+0
https://www.wolframalpha.com/input/?i=1+choose+0

leeopop:
Fixes #361

1. Add corner case handling for `combinations`
1. Add test cases for `combinations`, `combinations_with_replacement`, and `permutations`.

Zero-sized array is property handled in other combinatorics operators.

### DIFF
--- a/tests/test_std.rs
+++ b/tests/test_std.rs
@@ -637,7 +637,7 @@ fn combinations_with_replacement() {
     // Empty pool
     it::assert_equal(
         (0..0).combinations_with_replacement(2),
-        vec![vec![]],
+        <Vec<Vec<_>>>::new(),
     );
 }
 
--- a/tests/test_std.rs
+++ b/tests/test_std.rs
@@ -599,6 +599,7 @@ fn combinations_of_too_short() {
 #[test]
 fn combinations_zero() {
     it::assert_equal((1..3).combinations(0), vec![vec![]]);
+    it::assert_equal((0..0).combinations(0), vec![vec![]]);
 }
 
 #[test]
@@ -620,6 +621,11 @@ fn combinations_with_replacement() {
     // Zero size
     it::assert_equal(
         (0..3).combinations_with_replacement(0),
+        <Vec<Vec<_>>>::new(),
+    );
+    // Zero size on empty pool
+    it::assert_equal(
+        (0..0).combinations_with_replacement(0),
         <Vec<Vec<_>>>::new(),
     );
     // Empty pool
--- a/tests/test_std.rs
+++ b/tests/test_std.rs
@@ -627,17 +627,17 @@ fn combinations_with_replacement() {
     // Zero size
     it::assert_equal(
         (0..3).combinations_with_replacement(0),
-        <Vec<Vec<_>>>::new(),
+        vec![vec![]],
     );
     // Zero size on empty pool
     it::assert_equal(
         (0..0).combinations_with_replacement(0),
-        <Vec<Vec<_>>>::new(),
+        vec![vec![]],
     );
     // Empty pool
     it::assert_equal(
         (0..0).combinations_with_replacement(2),
-        <Vec<Vec<_>>>::new(),
+        vec![vec![]],
     );
 }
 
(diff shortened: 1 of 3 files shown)


tests/test_std.rs
```diff
@@ -622,6 +629,11 @@ fn combinations_with_replacement() {
         (0..3).combinations_with_replacement(0),
         <Vec<Vec<_>>>::new(),
     );
+    // Zero size on empty pool
+    it::assert_equal(
+        (0..0).combinations_with_replacement(0),
+        <Vec<Vec<_>>>::new(),
+    );
```
andersk: This looks wrong. The empty vector is a valid combination of zero elements in `0..0`, so `(0..0).combinations_with_replacement(0)` should be `vec![vec![]]` (like `(0..0).combinations(0)` and `(0..0).permutations(0)`), not `<Vec<Vec<_>>>::new() == vec![]`.

leeopop: @andersk 

It seems that the original test cases of `combinations_with_replacement` has the same test cases, and I just made a copy of it.
Should fixing `combinations_with_replacement` and its test cases be included in this PR or should I make a separate PR?

leeopop: I found that python itertools implementation shows value 1 (single empty set) for following inputs.
```
>>> len(list(itertools.combinations_with_replacement([], 0)))
1
>>> len(list(itertools.combinations([], 0)))
1
>>>
```

tests/test_std.rs
```diff
@@ -620,12 +627,17 @@ fn combinations_with_replacement() {
...
+    // Zero size on empty pool
+    it::assert_equal(
+        (0..0).combinations_with_replacement(0),
+        vec![vec![]],
     );
     // Empty pool
     it::assert_equal(
         (0..0).combinations_with_replacement(2),
-        <Vec<Vec<_>>>::new(),
+        vec![vec![]],
```
andersk: Now this is wrong. `vec![]` is only a valid combination of length 0; it’s not a valid combination of length 2.

andersk:
Here’s a simpler patch (against `master`) that achieves the correct results by removing special cases instead of by adding them:

```diff
--- a/src/combinations.rs
+++ b/src/combinations.rs
@@ -51,13 +51,11 @@ impl<I> Iterator for Combinations<I>
     type Item = Vec<I::Item>;
     fn next(&mut self) -> Option<Self::Item> {
         let mut pool_len = self.pool.len();
-        if self.pool.is_done() {
-            if pool_len == 0 || self.k > pool_len {
-                return None;
-            }
-        }
 
         if self.first {
+            if self.pool.is_done() && self.k > pool_len {
+                return None;
+            }
             self.first = false;
         } else if self.k == 0 {
             return None;
--- a/src/combinations_with_replacement.rs
+++ b/src/combinations_with_replacement.rs
@@ -66,7 +66,7 @@ where
         // If this is the first iteration, return early
         if self.first {
             // In empty edge cases, stop iterating immediately
-            return if self.k == 0 || self.pool.is_done() {
+            return if self.pool.is_done() && self.k > self.pool.len() {
                 None
             // Otherwise, yield the initial state
             } else {
```

andersk:
I opened #383 which fixes this while simplifying the code even further.

leeopop:
@andersk Yep, my PR was very rough as I'm not used to the internal implementation.
If you don't mind, you may close this PR and move to the new PR.

andersk:
Thanks for inspiring me to take a closer look. (But I’m just a contributor like you; I don’t have any special permission to close PRs.)

leeopop:
moved to #383

PR #381, opened 2019-11-06
Add Itertools::intersperse_with

gin-ahirsch:
Pretty self-explanatory. Useful for non-clonable elements or otherwise stateful interspersing.
`intersperse()` could be implemented via `intersperse_with()`, allowing the change to be more minimal. I haven't done it currently, but if you don't see any downsides I'd amend this change to do `self.intersperse_with(|| element)` for `intersperse()` instead.

### DIFF
--- a/src/intersperse.rs
+++ b/src/intersperse.rs
@@ -62,7 +62,7 @@ impl<I> Iterator for Intersperse<I>
         Self: Sized, F: FnMut(B, Self::Item) -> B,
     {
         let mut accum = init;
-        
+
         if let Some(x) = self.peek.take() {
             accum = f(accum, x);
         }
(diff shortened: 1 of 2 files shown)


FauxFaux:
I came here to write exactly this.

I planned to:
 * make the existing intersperse implementation call this one. Copy-pasting the code is presumably more efficient, though.
 * not use `FnMut`, but `Fn`. I guess `FnMut` is more general.

phimuemue:
>    make the existing intersperse implementation call this one.

I second this suggestion. We do it in other places, too, and my most recent tests indicate that the compiler is actually smart enough to inline things properly (at least with `--release`).

As a side note: Is possibly even `Interleave` similar enough so that it could share code with `Intersperse`?

gin-ahirsch:
It may not actually be possible to implement `intersperse()` through `intersperse_with()`:
```
error[E0308]: mismatched types
   --> src/lib.rs:401:31
    |
396 |     fn intersperse<F>(self, element: Self::Item) -> IntersperseWith<Self, F>
    |                    - this type parameter
...
401 |         self.intersperse_with(|| element)
    |                               ^^^^^^^^^^ expected type parameter `F`, found closure
```
Basically the same underlying issue as why #377 can't just return with a new closure - requiring `impl Trait` in return position.
The same will be true for `interleave()`.

gin-ahirsch:
Actually the issue is not just `impl Trait` in return position, but `impl Trait` in return position *for traits*. I'm not sure this is even on the map for Rust, as a Trait's interface is expected to be "fixed".
I tried using inherent trait methods (`impl dyn Itertools`), but that doesn't like methods with generic type parameters that don't have the restriction of `Self: Sized`:
```
error[E0038]: the trait `Itertools` cannot be made into an object
    --> src/lib.rs:2590:9
     |
1394 |     fn find_position<P>(&mut self, mut pred: P) -> Option<(usize, Self::Item)>
     |        ------------- method `find_position` has generic type parameters
...
2590 | impl<Item> dyn Itertools<Item=Item> {
     |            ^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Itertools` cannot be made into an object
```
I'm not sure why that would be problematic if not invoked from the inherent methods, but it seems that it is.
Even if that would be solved somehow, we cannot invoke generic methods at all though:
```
error: the `intersperse_with` method cannot be invoked on a trait object
    --> src/lib.rs:2467:14
     |
2467 |         self.intersperse_with(|| element)
     |              ^^^^^^^^^^^^^^^^
```
It's interesting that the method can accept `self` by value though, considering its type would be `dyn Itertools`.

jswrenn:
This looks good to me. Could you resolve the merge conflict?

jswrenn:
Thanks!

bors r+

PR #382, opened 2019-11-10
Add derive(Clone) where possible

andersk:


TomCrypto:
Would be great to have this merged, this PR is quite necessary when you have an iterator chain that eventually needs to be cloned (e.g. for `.cycle()`) but some itertools combinator in the chain fails to be `Clone`...

phimuemue:
Looks good to me. Maybe we can simplify the manual `impl`s after merging this (e.g. via a macro similar to `debug_fmt_fields`).

jswrenn:
Thanks!

bors r+

PR #383, opened 2019-11-12
Fix combinations(0) and combinations_with_replacement(0).

### LINKED ISSUE #361: Combination on zero-sized array

andersk:
The empty combination is always a valid combination of length 0.

Fixes #361 while also simplifying the code.

### DIFF
--- a/src/combinations.rs
+++ b/src/combinations.rs
@@ -51,13 +51,11 @@ impl<I> Iterator for Combinations<I>
     type Item = Vec<I::Item>;
     fn next(&mut self) -> Option<Self::Item> {
         let mut pool_len = self.pool.len();
-        if self.pool.is_done() {
-            if pool_len == 0 || self.k > pool_len {
-                return None;
-            }
-        }
 
         if self.first {
+            if self.pool.is_done() {
+                return None;
+            }
             self.first = false;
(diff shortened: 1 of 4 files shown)


jswrenn:
Thanks for the contribution! `combinations` has been around long enough that users _might_ be relying on this behavior. As such, I'm going to treat this like a breaking change, rather than a mere bug-fix, and postpone merging this until the next major version release.

jswrenn:
Thanks!

bors r+

PR #390, opened 2019-12-19
Extension trait for a range column-row iterator

jacobrosenthal:
Extension trait for a range, which accepts another range to make an iterator that would give x,y tuple 'visiting' every possibility. I wrote it for some pixel based stuff currently because I couldnt find it anywhere else. 

```
use itertools::column_row::RangeExt;
fn main() {
    for (x, y) in (0..3).rows(0..5) {
        println!("x:{:?} y:{:?}", x, y);
    }
}
```
produces
```
x:0 y:0
x:0 y:1
x:0 y:2
x:0 y:3
x:0 y:4
x:1 y:0
x:1 y:1
x:1 y:2
x:1 y:3
x:1 y:4
x:2 y:0
x:2 y:1
x:2 y:2
x:2 y:3
x:2 y:4
```
Currently doesnt support RangeTo which is a shame. But from looking at the code base I know my implementation doesnt fit here at all currently. Im happy to bikeshed literally everything about this pr, but wanted to come with something working. Let me know if it has a home here.

Thanks

### DIFF
--- a/src/column_row.rs
+++ b/src/column_row.rs
@@ -1,0 +1,46 @@
+use std::ops::Range;
+
+pub struct ColumnRowIter<T: Copy + core::cmp::PartialOrd + core::ops::AddAssign + From<u8>> {
+    range_row: Range<T>,
+    row_pos: T,
+    range_column: Range<T>,
+    col_pos: T,
+}
+
+impl<T> Iterator for ColumnRowIter<T>
+where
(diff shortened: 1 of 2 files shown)


phimuemue:
Isn't this [`cartesian_product`](https://docs.rs/itertools/0.8.2/itertools/trait.Itertools.html#method.cartesian_product)?

jacobrosenthal:
I think so! I googled, then asked a bunch of people, and still came up short. 
thanks for all the tools!

PR #391, opened 2020-01-02
Simplify `Clone` implementations

phimuemue:
Use a macro `clone_fields`, similar to `debug_fmt_fields`.

I replaced existing implementations of `clone_fields` and simplified the macro's parameters.


jswrenn:
Thanks!

bors r+

PR #397, opened 2020-01-10
don't require ParitalEq to the Item of DedupBy

KeenS:
`PartialEq` is't needed because `dedup_pred` take its place.


jswrenn:
Thanks!

bors r+

PR #403, opened 2020-02-06
Add CombinationsByLen iterator

### LINKED ISSUE #330: [Suggestion] Make another Combination iterator
Current implementation makes it hard to iterate over **all** combinations of infinite iterator.
E.g. every element from `(1..).combinations(3)` will always start with `[1,2,...]` so there is no way to ever get `[2,3,4]`.

But there is another way to implement this:
1. Iterate over all combinations which uses only indices < n (in lexicographic order).
2. Iterate over the rest.

This way our example will be `[1,2,3], [1,2,4], [1,3,4], [2,3,4], [1,2,5], ...`

It is practical to also have this implementation alongside with current.

fyrchik:
Closes #330 .

Sometimes we need to iterate over all possible combinations
of an infinite iterator. This means that every combination
must be eventually met. This PR implements such a possibility.


tests/test_std.rs
```diff
@@ -558,6 +558,51 @@ fn concat_non_empty() {
...
+    assert_eq!(it.next(), Some(vec![2,3,4]));
+    assert_eq!(it.next(), Some(vec![1,2,5]));
+    assert_eq!(it.next(), Some(vec![1,3,5]));
+    assert_eq!(it.next(), Some(vec![1,4,5]));
+    assert_eq!(it.next(), Some(vec![2,3,5]));
+    assert_eq!(it.next(), Some(vec![2,4,5]));
+    assert_eq!(it.next(), Some(vec![3,4,5]));
+    assert_eq!(it.next(), Some(vec![1,2,6]));
+}
+
```
phimuemue: Should we add some tests that actually work on infinite iterators?

src/combinations_by_len.rs
```diff
@@ -0,0 +1,154 @@
...
+}
+
+#[derive(Debug)]
+struct LazyBuffer<I: Iterator> {
+    it: I,
+    done: bool,
+    buffer: Vec<I::Item>,
+}
+
+impl<I> LazyBuffer<I> where I: Iterator
```
phimuemue: This looks very similar to the already existing `LazyBuffer`. Could we use the existing one?

tests/test_std.rs
fyrchik: The last test does so (iterator is `(1..)` by 3).
Sadly in test we can test only for a finite number of elements! I have decided to stop on 11.

src/combinations_by_len.rs
```diff
@@ -0,0 +1,154 @@
+use std::fmt;
+use std::ops::Index;
+
+/// An iterator to iterate through all the `n`-length combinations in an iterator.
+/// It works even for infinite iterators.
+///
+/// See [`combinations_by_len()`](../trait.Itertools.html#fn.combinations_by_len) for more information.
+#[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
+pub struct CombinationsByLen<I: Iterator> {
+    n: usize,
```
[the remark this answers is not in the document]
fyrchik: I think `n` is shorter and maybe a bit faster. `n` is also used in other iterators so I'd stick with it despite some duplication.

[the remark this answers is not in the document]
fyrchik: If #404 will be merged, I will change this also.

fyrchik:
@phimuemue thanks for the feedback! I have updated PR.

tests/test_std.rs
phimuemue: Ah, thank you, I misread that.

PR #404, opened 2020-02-06
`Combinations::k` is redundant

phimuemue:
If I am not mistaken, we can get rid of `Combinations::k` by simply referring to `indices.len` instead.

Also, we could simplify some things in `Combinations::next`.


jswrenn:
Thanks!

bors r+

PR #406, opened 2020-02-10
Into_group_map_by

Sagebati:
Add the into_group_map_by , tried with returning an iterator instead of an hashmap, can ben reverted on demand.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -65,6 +65,7 @@ use std::hash::Hash;
 use std::fmt::Write;
 #[cfg(feature = "use_std")]
 type VecIntoIter<T> = ::std::vec::IntoIter<T>;
+#[cfg(feature = "use_std")]
 use std::iter::FromIterator;
 
 #[macro_use]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -680,7 +680,7 @@ pub trait Itertools : Iterator {
     /// let it = (0..8).step(3);
     /// itertools::assert_equal(it, vec![0, 3, 6]);
     /// ```
-    #[deprecated(note="Use std .step_by() instead", since = "0.8")]
+    #[deprecated(note="Use std .step_by() instead", since="0.8")]
     #[allow(deprecated)]
     fn step(self, n: usize) -> Step<Self>
         where Self: Sized
@@ -858,7 +858,7 @@ pub trait Itertools : Iterator {
         where Self: Sized,
               Self::Item: IntoIterator,
               F: FnMut(&<Self::Item as IntoIterator>::Item,
-                  &<Self::Item as IntoIterator>::Item) -> bool
+                       &<Self::Item as IntoIterator>::Item) -> bool
     {
         kmerge_by(self, first)
     }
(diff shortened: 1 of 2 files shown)


phimuemue:
Hi! I see that this has a certain value. The following questions came to my mind:

* Should we implement `into_group_map_by` in terms of `into_group_map`? (This way, any improvements to `into_group_map` would automatically be obtained by `into_group_map_by`.)
* I think it is a good idea to postpone `into_group_map_by_fold` for now, as it does quite a bit, and maybe we could do that more efficient.
* I am not sure, but it looks like there are some commits that are not necessary for this PR. For review, it is maybe convenient to rebase the commits first.

SamBlaise:
> Should we implement into_group_map_by in terms of into_group_map? (This way, any improvements to into_group_map would automatically be obtained by into_group_map_by.)

I understand it introduces inconsistency in the API, but for me the API of into_group_map was too specific from the beginning. I think both ways are an good idea. (in terms of into_group_map and not).

Yeah, I couldn't get a proper implementation for group_map_by_fold. I'm sorry I'm not a git pro, and I'll look into it, definitively there is some useless commits.

jswrenn:
Thanks for the contribution!

Is there a strong reason this should produce an iterator as opposed to the HashMap itself?

I agree with all three of @phimuemue points. In the interest of keeping itertools as simple as possible, I'd like to see the `into_group_map_by` method just be a wrapper around `into_group_map`.

SamBlaise:
The only reason was to facilitate chaining more methods without using into_iter(). 
Now into_group_map_by uses into_group_map and returns an HashMap.

phimuemue:
Hi @Sagebati! I'm trying to resolve some PRs. Are you still interested in getting this into itertools?

Sagebati:
Hi, I'll do it!

Sagebati:
Formatting done !

jswrenn:
https://github.com/rust-itertools/itertools/pull/406/commits/273b382e56d5930dcd53a8640a960a9f1d1f64e3 inadvertently reverted the return type change you made in https://github.com/rust-itertools/itertools/pull/406/commits/a0f329a02a9d48f2db06a08e6b8081dd44957b11. Could you make `into_group_map_by` return a `HashMap` again?

Sagebati:
Ready!

src/lib.rs
```diff
@@ -65,7 +65,6 @@ use std::hash::Hash;
 use std::fmt::Write;
 #[cfg(feature = "use_std")]
 type VecIntoIter<T> = ::std::vec::IntoIter<T>;
-#[cfg(feature = "use_std")]
 use std::iter::FromIterator;
```
phimuemue, line 68: Is the `use_std` condition intentionally gone?

Sagebati: No, I'm embarrassed, I corrected it.

phimuemue: Never mind! :)

Sagebati:
The "waiting_on_author" is still there maybe an error ?

jswrenn:
Thanks for the ping! This looks good to me!

bors r+

Issue #405, opened 2020-02-06
Travis CI problems


phimuemue:
As of writing this, `cargo bench` apparently has problems. Looking at https://travis-ci.org/rust-itertools/itertools/jobs/647077808#L425, I think it is a problem with `criterion`.

jswrenn:
These issues seem to have been introduced by #396. Oddly, Travis was fine with that pull request.

phimuemue:
Not sure, but didn't the original use `criterion 0.3.0`, while it now tries `0.3.1`?

Philippe-Cholet:
Seems resolved a long time ago and we don't use Travis CI anymore.

PR #408, opened 2020-02-11
Temporary fix for CI

phimuemue:
Referring to https://github.com/rust-itertools/itertools/issues/405: As far as I understood, `dev-dependencies` are for examples, tests, and benches.

I tried to fix CI by fixing `criterion 0.3.0` for the time being so that it does not get in the way of other things in the queue.

From looking at https://github.com/bheisler/criterion.rs, it seems as if they bumped required rust version to 1.33, which I would not do for itertoos.

As noted in the patch, I hope that we can relax the version requirement one day.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ either = { version = "1.0", default-features = false }
 
 [dev-dependencies]
 rand = "0.7"
-criterion = "0.3"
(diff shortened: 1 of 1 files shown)


jswrenn:
Thank you for this stop-gap! I'm happy to merge it.

I'm still mulling over what a good long-term solution is. My latest idea is to mark `criterion` as an optional dependency behind a `benchmark` feature, and add a `[[bench]]` section to `Cargo.toml` listing `benchmark` as a required feature.

PR #411, opened 2020-02-16
Use field init shorthand

phimuemue:
I suggest that we use field init shorthand ([available since Rust 1.17](https://doc.rust-lang.org/nightly/edition-guide/rust-2018/data-types/field-init-shorthand.html)) whenever possible, as it seems easier and canonical to me.


jswrenn:
Thanks!!!

bors r+

PR #412, opened 2020-02-16
Position max, min and minmax methods

### LINKED ISSUE #410: Feature request: arg_min and arg_max
Currently there is no function to find the index of the minimum or the maximum (somehow analogous to C++ [min_element](https://en.cppreference.com/w/cpp/algorithm/min_element) and [max_element](https://en.cppreference.com/w/cpp/algorithm/max_element)).

In my project I ended up writing `arg_max`, `arg_max_by_key`, `arg_max_by`, `arg_min`, `arg_min_by_key`, `arg_min_by`.

Would a PR adding these be accepted? I do think `arg_max` is maths terminology, not iterator terminology, so I think something like `fn max_position(&mut self) -> Option<(usize, Self::Item)>` would be more consistent with for example [`find_position`](https://docs.rs/itertools/0.8.2/itertools/trait.Itertools.html#method.find_position). My vote would be `{max,min}_position{,_by_key,_by}` for a total of six methods, each returning `Option<(usize, Self::Item)>`.

tspiteri:
The following nine methods are provided:
  * `position_max`, `position_max_by_key`, `position_max_by`; return `Option<usize>`
  * `position_min`, `position_min_by_key`, `position_min_by`; return `Option<usize>`
  * `position_minmax`, `position_minmax_by_key`, `position_minmax_by`; return `MinMaxResult<usize>`

The `position_max` and `position_min` methods are implemented using `Iterator::enumerate` and `Iterator::max_by`/`min_by`.

The `position_minmax` methods are implemented using `Iterator::enumerate` and `minmax::minmax_impl` rather than `Itertools::minmax_by`, so that the comparison constraint can be relaxed from `Ord` to `PartialOrd`, just like for the `minmax` methods.

Closes #410.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2514,7 +2514,7 @@ pub trait Itertools : Iterator {
         where Self: Sized, K: PartialOrd, F: FnMut(&Self::Item) -> K
     {
         use MinMaxResult::{NoElements, OneElement, MinMax};
-        match minmax::minmax_impl(self.enumerate(), |x| key(&x.1), |_, _, xk, yk| xk < yk) {
+        match self.enumerate().minmax_by_key(|e| key(&e.1)) {
             NoElements => NoElements,
             OneElement(x) => OneElement(x.0),
             MinMax(x, y) => MinMax(x.0, y.0),
@@ -2557,11 +2557,7 @@ pub trait Itertools : Iterator {
         where Self: Sized, F: FnMut(&Self::Item, &Self::Item) -> Ordering
     {
         use MinMaxResult::{NoElements, OneElement, MinMax};
-        match minmax::minmax_impl(
-            self.enumerate(),
-            |_| (),
-            |x, y, _, _| Ordering::Less == compare(&x.1, &y.1)
-        ) {
+        match self.enumerate().minmax_by(|x, y| compare(&x.1, &y.1)) {
             NoElements => NoElements,
             OneElement(x) => OneElement(x.0),
             MinMax(x, y) => MinMax(x.0, y.0),
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2254,6 +2254,172 @@ pub trait Itertools : Iterator {
         )
     }
 
+    /// Return the position of the maximum element in the iterator.
+    ///
+    /// If several elements are equally maximum, the position of the
+    /// last of them is returned.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use itertools::Itertools;
+    ///
+    /// let a: [i32; 0] = [];
+    /// assert_eq!(a.iter().position_max(), None);
+    ///
+    /// let a = [-3, 0, 1, 5, -10];
+    /// assert_eq!(a.iter().position_max(), Some(3));
+    ///
+    /// let a = [1, 1, -1, -1];
+    /// assert_eq!(a.iter().position_max(), Some(1));
+    /// ```
+    fn position_max(self) -> Option<usize>
+        where Self: Sized, Self::Item: Ord
+    {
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -2254,6 +2254,320 @@ pub trait Itertools : Iterator {
...
+    /// let a = [1, 1, -1, -1];
+    /// assert_eq!(a.iter().position_minmax(), MinMax(2, 1));
+    /// ```
+    ///
+    /// [`MinMaxResult`]: enum.MinMaxResult.html
+    fn position_minmax(self) -> MinMaxResult<usize>
+        where Self: Sized, Self::Item: PartialOrd
+    {
+        use MinMaxResult::{NoElements, OneElement, MinMax};
+        match minmax::minmax_impl(self.enumerate(), |_| (), |x, y, _, _| x.1 < y.1) {
```
jswrenn: Could this be written more simply as:
```suggestion
        match self.enumerate().minmax_by_key(|e| e.1) {
```

src/lib.rs
```diff
@@ -2254,6 +2254,320 @@ pub trait Itertools : Iterator {
...
+    /// assert_eq!(a.iter().position_minmax_by_key(|x| x.abs()), MinMax(0, 3));
+    /// ```
+    ///
+    /// [`MinMaxResult`]: enum.MinMaxResult.html
+    /// [`position_minmax`]: #method.position_minmax
+    fn position_minmax_by_key<K, F>(self, mut key: F) -> MinMaxResult<usize>
+        where Self: Sized, K: PartialOrd, F: FnMut(&Self::Item) -> K
+    {
+        use MinMaxResult::{NoElements, OneElement, MinMax};
+        match minmax::minmax_impl(self.enumerate(), |x| key(&x.1), |_, _, xk, yk| xk < yk) {
```
jswrenn: Could this be written as:
```suggestion
        match self.enumerate().minmax_by_key(|e| key(&e.1)) {
```

src/lib.rs
```diff
@@ -2254,6 +2254,320 @@ pub trait Itertools : Iterator {
...
+    /// [`position_minmax`]: #method.position_minmax
+    fn position_minmax_by<F>(self, mut compare: F) -> MinMaxResult<usize>
+        where Self: Sized, F: FnMut(&Self::Item, &Self::Item) -> Ordering
+    {
+        use MinMaxResult::{NoElements, OneElement, MinMax};
+        match minmax::minmax_impl(
+            self.enumerate(),
+            |_| (),
+            |x, y, _, _| Ordering::Less == compare(&x.1, &y.1)
+        ) {
```
jswrenn: Could this be written as:
```suggestion
        match self.enumerate().minmax_by(|x, y| compare(&x.1, &y.1)) {
```

src/lib.rs
tspiteri: The other two can be, but this tries to move `e` because `e.1` does not implement `Copy`. Incidentally, this was the first of the three I implemented and did not find a way to do this with the same trait constraints without going to `minmax_impl`, and then for the other two I just repeated, though the other two can be implemented simpler as suggested.

This function
1. cannot use `minmax` because the `self.enumerate()` item does not implement `PartialOrd`.
2. cannot use `minmax_by_key` because the `self` item does not implement `Copy`.
3. cannot use `minmax_by` because from `PartialOrd` we cannot get `Ordering` without unwrapping. Unwrapping with panic is different behavior from the `minmax` methods. Unwrapping with default could be possible, but then that makes this even more hairy than using `minmax_impl`.

jswrenn:
Thanks for the contribution!

bors r+

PR #414, opened 2020-02-29
Fix "missing dyn" warning

phimuemue:
This should fix the warning about a deprecated missing `dyn`.


phimuemue:
Wow. That was quick.

BTW: Thanks a lot for your good work on maintaining this crate! - it helps me quite a bit in my projects.

jswrenn:
> BTW: Thanks a lot for your good work on maintaining this crate! - it helps me quite a bit in my projects.

It would be a significantly harder task without your code reviews! I appreciate your feedback on incoming PRs tremendously.

PR #415, opened 2020-02-29
Simplify `Combinations::next`

phimuemue:
* `pool.len` looks up the length of a vector, ([which is just a field lookup](https://doc.rust-lang.org/src/alloc/vec.rs.html#1335-1337)), so I see no real value in manually doing the same computation.

* Checking `pool.is_done` before calling `pool.get_next` is redundant, as `get_next` does the same check internally.


jswrenn:
Thanks!

bors r+

PR #418, opened 2020-03-06
Add a stability specification to the docs for unique and unique_by

mulkieran:


mulkieran:
It turns out that some of my code is relying on this stability property, and it appears to be true at this time. I'ld like to know for certain if the stability is a guaranteed property or if it might change in future. If you accept this PR, then I'll consider the property to be guaranteed.

phimuemue:
Hi there. I think if we want to cast that in stone, we should have tests to cover this property.

Do you happen to know if we have such tests?

mulkieran:
The doc test for ```unique_by``` and the test for ```unique_by``` in ```test_std.rs``` could have different expected outcomes according to the current unrestrictive definition. But if the stability criterion is accepted, then the only correct solution is the one that the tests do have.

phimuemue:
Maybe the bounds we require for `unique`/`unique_by` (i.e. `Eq + Hash`) already implies enough to deduce that it is stable?

mulkieran:
> Maybe the bounds we require for `unique`/`unique_by` (i.e. `Eq + Hash`) already implies enough to deduce that it is stable?

That could never ensure this property. But I realize that my doc change is not as strict as what I intended to say. Basically, I want to ensure that the first element in the iterator of a set of duplicate elements is the one that is retained. I'm going to try a restatement.

jswrenn:
Thanks for the contribution!

bors r+

PR #420, opened 2020-03-07
Changelog for 0.9.0

jswrenn:


matklad:
For semver breaking bumps (0.8 -> 0.9), it might be useful to specifically call out which changes are breaking.

PR #422, opened 2020-03-09
Bump to 0.9 in README.rst

cuviper:


cuviper:
Hmm, actually I just noticed that the master `Cargo.toml` doesn't reflect the published 0.9.0 either -- was this just not pushed?

jswrenn:
Good catch. Yeah, I didn't push the release commit (b085c523681146151866ead66ae8a4b8967fd005).

meltinglava:
Looks like the git tag is also missing.

PR #423, opened 2020-03-09
Added `Itertools::dedup_with_count()` and `Itertools::dedup_by_with_count()`

### LINKED ISSUE #393: [RFC] `dedup()`/`dedup_by()` with count
Every once in a while I find myself re-implementing a "dedup with count", an iterator that instead of yielding `T` yields `(usize, T)`:

```rust
let data = vec![10, 10, 20, 30, 30, 30, 20];
assert_eq!(
    data.into_iter().dedup_with_count().collect(),
    vec![(2, 10), (1, 20), (3, 30), (1, 20)]
);
```

If you agree that these (`dedup_with_count()`/`dedup_by_with_count()`) are useful methods I can open a PR with the implementation soon.

orium:
Fixes #393.

### DIFF
--- a/src/adaptors/mod.rs
+++ b/src/adaptors/mod.rs
@@ -607,18 +607,18 @@ impl<I, J, F> Iterator for MergeBy<I, J, F>
 }
 
 #[derive(Clone, Debug)]
-pub struct CoalesceCore<I>
+pub struct CoalesceCore<I, T>
     where I: Iterator
 {
     iter: I,
-    last: Option<I::Item>,
+    last: Option<T>,
 }
 
-impl<I> CoalesceCore<I>
+impl<I, T> CoalesceCore<I, T>
     where I: Iterator
 {
-    fn next_with<F>(&mut self, mut f: F) -> Option<I::Item>
-        where F: FnMut(I::Item, I::Item) -> Result<I::Item, (I::Item, I::Item)>
+    fn next_with<F>(&mut self, mut f: F) -> Option<T>
+        where F: FnMut(T, I::Item) -> Result<T, (T, T)>
     {
         // this fuses the iterator
         let mut last = match self.last.take() {
@@ -652,7 +652,7 @@ impl<I> CoalesceCore<I>
 pub struct Coalesce<I, F>
     where I: Iterator
 {
-    iter: CoalesceCore<I>,
+    iter: CoalesceCore<I, I::Item>,
     f: F,
 }
 
@@ -705,7 +705,7 @@ impl<I, F> Iterator for Coalesce<I, F>
 pub struct DedupBy<I, Pred>
     where I: Iterator
 {
-    iter: CoalesceCore<I>,
+    iter: CoalesceCore<I, I::Item>,
     dedup_pred: Pred,
 }
 
--- a/src/adaptors/mod.rs
+++ b/src/adaptors/mod.rs
@@ -10,7 +10,7 @@ pub use self::multi_product::*;
 
 use std::fmt;
 use std::mem::replace;
-use std::iter::{Fuse, Peekable, FromIterator};
+use std::iter::{Fuse, Peekable, FromIterator, FusedIterator};
 use std::marker::PhantomData;
 use crate::size_hint;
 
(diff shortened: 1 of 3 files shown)


src/adaptors/mod.rs
```diff
@@ -803,6 +803,88 @@ impl<I, Pred> Iterator for DedupBy<I, Pred>
     }
 }
 
+/// An iterator adaptor that removes repeated duplicates, while keeping a count of how many
+/// repeated elements were present. This will determine equality using a comparison function.
+///
+/// See [`.dedup_by_with_count()`](../trait.Itertools.html#method.dedup_by_with_count) or
+/// [`.dedup_with_count()`](../trait.Itertools.html#method.dedup_with_count) for more information.
+#[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
+pub struct DedupByWithCount<I, Pred>
```
orium: Should this be on the `adaptors` module?

orium:
@jswrenn Can you take a look when you have some time?

jswrenn:
Sorry for the delay and for the contribution! My only nit is the name—it suggests that `count` will be the second component of the tuple. Perhaps `count_dedup` and `count_dedup_by`?

(This is yet another case where having rust-lang/rfcs#2584 would be helpful!)

phimuemue:
Hi there! Should we try to implement this in terms of (a generalized) `Dedup`/`DedupBy`?

As far as I can see, it (naturally) shares quite a bit of implementation with `Dedup`, and we could possibly express these similarities in code. (E.g. why does `Dedup` use `CoalesceIter` internally, while `DedupWithCount` goes with `Peekable`? Could this lead to subtle differences in behavior?)

If we actually want implement it in terms of `Dedup`, one way I could imagine is to equip `DedupBy` with an additional customization point (read: type parameter) that can be used to do a side-computation in `next_with` resp. `fold`. In the plain old `Dedup`/`DedupBy`-case, these side-computations could just be no-ops, while for `DedupWithCount`/`DedupByWithCount` it could do the counting.

orium:
@jswrenn 

> My only nit is the name—it suggests that count will be the second component of the tuple. Perhaps count_dedup and count_dedup_by?

I think it makes sense to start the function name with the fundamental thing that is going on, which is deduplication.  The count is the extra metadata.  I also like how an alphabetical order of the names of the functions naturally group these functions together (and also that autocompletion on `dedup` will list all these).

@phimuemue

Yeah, it makes sense to try to re-use code.  I've re-wrote it in terms of `CoalesceCore`, which I generalize slightly.  If I try to change `DedupBy` to accommodate a dedup with count it will either look like `CoalesceCore` or it feels a bit forced.  Also, the `fold` specialization for `DedupBy` doesn't fit very well with a dedup with count.

phimuemue:
Kudos for going with `CoalesceCore`. Imho it looks better.

Imho, we should still strive to unify the implementations further, but your observation

> If I try to change `DedupBy` to accommodate a dedup with count it will either look like `CoalesceCore` or it feels a bit forced. Also, the `fold` specialization for `DedupBy` doesn't fit very well with a dedup with count.

made me think why we even have `CoalesceCore` as an extra struct in the first place. I think we could even try to unify the implementations of `Coalesce` and `Dedup`(`By`), simplifying the implementations quite a bit. This would essentially mean that `Dedup`(`By`) is just a special case of `Coalesce` (which - intuitively - seems reasonable to me), and (I think) `Dedup`(`By`)`WithCount` could then be specializations, too.

Doing so, however, may be a bit too much for this ticket.

orium:
Yes, I think that's  perfectly possible.  I can look into it, but, as you suggest, I prefer to do that in a later PR.

phimuemue:
> 
> 
> Yes, I think that's perfectly possible. I can look into it, but, as you suggest, I prefer to do that in a later PR.

I sketched an implementation for this in https://github.com/rust-itertools/itertools/pull/440. As stated there, I will rebase it if this PR is accepted.

src/adaptors/mod.rs
jswrenn: This location of `DedupByWithCount` is fine. The `adaptors/mod.rs` file is absolutely getting over-sized, though. A restructuring is overdue, but this PR doesn't need to do it.

jswrenn:
Looks good to me! Thanks!

bors r+

PR #427, opened 2020-03-11
WIP: Trying to make sure tests only run on nightly

meltinglava:


meltinglava:
Going to need to figure out how to run travis runs locally. This initial commit is definatly wrong

jswrenn:
It's possibly worth revisiting this someday, but I'm going to close this. We moved to github actions, so the PR would need to be re-done anyways.

Issue #425, opened 2020-03-10
travis pipeline is slow


meltinglava:
The pipline takes significant ammount of time. Think there is a better way of doing this.

- [x] 1. ~~run `cargo bench --no-run` insdead of `cargo bench` as they take a while and I suspect results are not looked at. Feel like benchmarking results are better run locally. Checking that they build would always be nice.~~ merged in #426 

- [ ] 2. Building on all platforms, while just running test, check, (fmt, clippy)... on one toolchain (stable or nightly)

- [x] 3. Add caching for builds and compiler. ref: https://docs.travis-ci.com/user/languages/rust/#dependency-management. Ref #429

bluss:
I think it's time to migrate from travis-ci.org - they are closing down. Github Actions should be an ok replacement?

jswrenn:
Agree completely.

PR #429, opened 2020-03-15
add caching to travis

meltinglava:
This is so that the pipeline does not download and compile all the dependencies all the time. Its recompiles them when a new compiler is used.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ matrix:
     - rust: stable
     - rust: beta
     - rust: nightly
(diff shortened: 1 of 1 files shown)


meltinglava:
Ref part 3 of #425

.travis.yml
```diff
@@ -6,6 +6,7 @@ matrix:
     - rust: stable
     - rust: beta
     - rust: nightly
+cashe: cargo # https://docs.travis-ci.com/user/languages/rust/#dependency-management
```
jswrenn: Should this be `cache`?

meltinglava: Yep. Typo on my part.

PR #432, opened 2020-03-25
Add Itertools::zip_with

Eolu:
a simple implementation of zip_with, quite surprised it was missing. I'm still learning the basics of Rust so I would appreciate any nit picked as far as anything missing or incorrect in the implementation or documentation.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -134,6 +134,7 @@ pub mod structs {
     pub use crate::zip_eq_impl::ZipEq;
     pub use crate::zip_longest::ZipLongest;
     pub use crate::ziptuple::Zip;
+    pub use crate::zip_with::ZipWith;
 }
 
 /// Traits helpful for using certain `Itertools` methods in generic contexts.
@@ -207,6 +208,7 @@ mod with_position;
 mod zip_eq_impl;
 mod zip_longest;
 mod ziptuple;
+mod zip_with;
 
 #[macro_export]
 /// Create an iterator over the “cartesian product” of iterators.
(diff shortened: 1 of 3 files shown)


scottmcm:
I think this was probably omitted because instead of 
```rust
a.zip_with(b, |x, y| x + y)
```
you can just do
```rust
a.zip(b).map(|(x, y)| x + y)
```

So I don't know if this addition is worth having.

phimuemue:
Apparently, there already [have been thoughts](https://github.com/rust-lang/rust/issues/46654) about this.

If we decide to keep `zip_with` in itertools: Should we at least just implement it in terms of `Map` and `Zip` (so that `zip_with(self, other, zipper)` would return a `Map<Zip<Self, Other>, Zipper>` (or the like)) so that we gain all the specializations/optimizations done by `Map` and `Zip`?

Eolu:
So my argument back here is that zip creates a tuple, so we can't do 
```
use std::cmp::max;
a.zip(b).map(max)
```
But this does allow code to be written in the form
```
a.zip_with(b, max)
```

ChiefMilesEdgeworth:
This could use a couple tests, just to make sure it behaves as expected :)

jswrenn:
@Eolu thanks for the contribution! I'm sorry for the slow response; this has taken me a while to mull over.

If I'm understanding correctly, the motivation is to avoid having to write a closure to unpack arguments when iterating over tuples.

If that's the case, then this isn't _just_ an inconvenience limited to just zipping-then-mapping (which `zip_with` covers). What about zipping-then-filtering? Or zipping-then-filter-mapping? Or, really, _any_ case where you're applying a function over tuples.

It would be surprising to me to handle the _particular_ case of zipping-then-mapping (and _only_ in the case of applying the function to pairs), but not any of these other cases.

The prospect of handling all of these situations like `zip_with` (i.e., creating a dedicated method for each situation) is bleak, and the maintenance burden would be significant (as @phimuemue noted, we'd need to ensure that performance specializations implemented for the constituent operations are present on the new composite operations).

If the issue is that calling functions on tuples of arguments is unergonomic, it's probably best to address that issue in a way that isn't specific to iterators and isn't fundamentally limited to arity = 2. For example:
```
#![feature(fn_traits)]
#![feature(unboxed_closures)]

fn apply<Arguments, Return>(
    mut f: impl Fn<Arguments, Output = Return>,
) -> impl Fn(Arguments) -> Return
{
    move |arguments| f.call(arguments)
}

fn add_test() {
    use core::ops::Add;

    let l = 0..10;
    let r = 10..20;
    let _ = l.zip(r).map(apply(Add::add));
}
```

Eolu:
I see. Looking at it like this is interesting, as working with functional languages in the past has left me with the impression that zip was itself the `|a, b| (a, b)` specialization of zipWith, with zip -> map being a kind of workaround get back the fundamental functionality of zipWith in cases where a tuple wasn't desirable. So seeing a zip implementation available without zipWith seemed like a strange oversight. That said, maybe given the current design of things the response you just gave is the better way to handle this.

jswrenn:
> I see. Looking at it like this is interesting, as working with functional languages in the past has left me with the impression that zip was itself the |a, b| (a, b) specialization of zipWith, with zip -> map being a kind of workaround get back the fundamental functionality of zipWith in cases where a tuple wasn't desirable.

This is a perspective I hadn't considered! Unfortunately, because of the [aforementioned](https://github.com/rust-itertools/itertools/pull/432#issuecomment-617360581) concerns, I'm going to close this for the time being.

PR #435, opened 2020-04-11
Add multipeek method

fenhl:
This adds a `multipeek` method to the `Itertools` trait as a more convenient way to use the `multipeek` function.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2625,6 +2625,28 @@ pub trait Itertools : Iterator {
             None => Err(ExactlyOneError::new((None, None), self)),
         }
     }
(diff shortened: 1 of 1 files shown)


phimuemue:
I wonder if we have a "standard" way of dealing with free functions vs methods.

I think we could accept this PR, but it may be worth looking into bringing all the free functions vs methods in line (e.g. by always implementing the method and having the free function call it).

There may be obstacles regarding this idea, though.

jswrenn:
Huh, I'm surprised that there isn't parity between the free functions and methods.

bors r+

PR #436, opened 2020-04-15
Add `chunk_while` iterator

pwoolcoc:
This is an iterator that is a mix of `std::slice::chunks` and `fold`.
Instead of chunking the source up into predictable lengths though, it
allows the user to inspect each element of the source, along with a
state, and decide when the chunk is big enough.

### DIFF
--- a/src/groupbylazy.rs
+++ b/src/groupbylazy.rs
@@ -1,5 +1,6 @@
 use std::cell::{Cell, RefCell};
 use std::vec;
+use std::iter::Peekable;
 
 /// A trait to unify FnMut for GroupBy with the chunk key in IntoChunks
 trait KeyFunction<A> {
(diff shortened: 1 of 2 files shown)


pwoolcoc:
Sorry if I should have taken this to the stdlib first, but given 

> if your feature involves heap allocation, such as storing elements in a Vec<T>, then it can't be accepted into libcore, and you should propose it for itertools directly instead.

I figured I would just bring it here, since it uses a Vec internally

pwoolcoc:
two things I'm unhappy with:

1. `type Item = Vec<T>`. I wish that could be `type Item = &'a [T]`, but then I run into the problem of tying `'a` to self in `fn next(&mut self, ...)`. If anyone has a way I can do this, I'm all ears.
2. `B: Clone`. I'm not sure if there's a way to do this or not, but every time it starts accumulating a new chunk, it needs to reset the state. Again, if there's a better way, I'd love to hear about it

phimuemue:
Hi there!

Regarding `type Item=Vec<T>`: Do you think we could we make this a "lazy" chunking operation? (iirc such as `GroupBy`.)

pwoolcoc:
@phimuemue thanks, I'm looking into that right now and I'll update here when I can

Philippe-Cholet:
@pwoolcoc Hi, do you want to update your work on `chunk_while` here?

Philippe-Cholet:
I would prompty reopen if needed.

PR #437, opened 2020-05-06
Simpl is empty

phimuemue:
Taking inspiration from https://github.com/rust-itertools/itertools/pull/303#discussion_r420883592, I think we could exploit that `pop`/`pop_front` return `Option`.

Note: https://godbolt.org/z/WVTD7m seems to indicate that the compiler is doing a good job at optimizing either version.

### DIFF
--- a/src/multipeek_impl.rs
+++ b/src/multipeek_impl.rs
@@ -82,11 +82,7 @@ impl<I> Iterator for MultiPeek<I>
 
     fn next(&mut self) -> Option<I::Item> {
         self.index = 0;
-        if self.buf.is_empty() {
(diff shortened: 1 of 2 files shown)


jswrenn:
Oh, sweet! Thanks!

bors r+

---

@phimuemue, do you have any objections to adding you as a member to the itertools repo? Your code reviews are absolutely invaluable, and they should carry the same weight  as reviews from from bluss or me. I think it'd be appropriate if you had the `[member]` badge next to your username when you comment.

phimuemue:
@jswrenn Thank you for the offer. If it's still available, I'd gladly accept it.

PR #438, opened 2020-05-11
impl debug trait for diff enum

Zaba505:
I added a fairly simple yet informative `fmt::Debug` implementation to the `Diff` enum. The reasoning behind this is to simplify tests where you just want to say:
```rust
// Diff some expected iter with the actual outputted one
let diff = itertools::diff_with(ex, out, |a, b| a == b);

// Now, all you care about is that there is no difference
// If there is, the index of the diff will be logged by assert_eq!
// since diff now impls fmt::Debug
assert_eq!(None, diff);
```

### DIFF
--- a/src/diff.rs
+++ b/src/diff.rs
@@ -13,8 +13,9 @@ use crate::structs::PutBack;
 /// `Diff` represents the way in which the elements yielded by the iterator `I` differ to some
 /// iterator `J`.
 pub enum Diff<I, J>
-    where I: Iterator,
-          J: Iterator
+where
+    I: Iterator,
+    J: Iterator,
 {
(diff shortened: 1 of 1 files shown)


Zaba505:
I just realized my sample scenario won't work even with the `fmt::Debug` impl since `assert_eq!` requires `Eq/PartialEq` to be implemented. If this PR is accepted, then I will open another providing a simple implementation of `PartialEq`.

jswrenn:
I'd prefer a `#[derive(Debug)]` impl—the property that you can generally copy a debug output and paste it into source code is really useful!

> If this PR is accepted, then I will open another providing a simple implementation of PartialEq.

Iterators usually don't provide a `PartialEq` implementation, so this wouldn't be very helpful.

PR #440, opened 2020-05-18
Implement DedupBy in terms of Coalesce

phimuemue:
During the discussion of https://github.com/rust-itertools/itertools/pull/423#issuecomment-629713830, we found that `DedupBy`/`DedupByWithCount` could actually be implemented in terms of `Coalesce`. This simplifies the implementation quite a bit and lets derived adaptors inherit specializations.

* Implement `fold` on `Coalesce` so that all the adaptors derived from it inherit the specialization (currently, only `DedupBy` has this specialization).
* Introduce `CoalescePredicate` (similar to `DedupPredicate`) to allow for different customizations of `Coalesce`.
* Introduce parametrizable `CoalesceBy`: Base for `Coalesce`, `DedupBy`, `DedupByWithCount`.
* Implement `DedupBy`/`DedupByWithCount` in terms of `CoalesceBy` (using particular `impl`s).
* At last, the indirection through `CoalesceCore` is not needed anymore.

### DIFF
--- a/src/adaptors/mod.rs
+++ b/src/adaptors/mod.rs
@@ -649,7 +649,9 @@ impl<I, T> CoalesceCore<I, T>
 ///
 /// See [`.coalesce()`](../trait.Itertools.html#method.coalesce) for more information.
 #[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
-pub struct Coalesce<I, F>
+pub type Coalesce<I, F> = CoalesceBy<I, F>;
+
+pub struct CoalesceBy<I, F>
     where I: Iterator
 {
     iter: CoalesceCore<I, I::Item>,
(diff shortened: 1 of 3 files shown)


src/adaptors/mod.rs
```diff
@@ -606,107 +605,121 @@ impl<I, J, F> Iterator for MergeBy<I, J, F>
     }
 }
 
-#[derive(Clone, Debug)]
-pub struct CoalesceCore<I, T>
+/// An iterator adaptor that may join together adjacent elements.
+///
+/// See [`.coalesce()`](../trait.Itertools.html#method.coalesce) for more information.
+#[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
+pub struct Coalesce<I, F, T>
```
jswrenn, line 612: Unfortunately, adding an extra type parameter to `Coalesce` is a breaking change.

phimuemue: That's a very good objection. I already tried to keep it on 2 parameters, but ran into problems since `Coalesce` has to know the result of `F` and I had problems specifying that.

Could I just rename `Coalesce` to `CoalesceBy` (keeping the type parameters) and define `type Coalesce<I, F> = CoalesceBy<I, F, I::Item>` (and the others alike)?

jswrenn: Yep!

jswrenn:
Thanks!

bors r+

PR #441, opened 2020-05-19
Simplify phantom data type parameter

phimuemue:
Simplifies the type parameter for a `PhantomData`.


jswrenn:
bors r+

(btw, I think you can `bors r+` things, too!)

PR #444, opened 2020-05-30
Remove unused macro variant

phimuemue:
I'd like to try bors with that one.

### DIFF
--- a/src/tuple_impl.rs
+++ b/src/tuple_impl.rs
@@ -245,7 +245,6 @@ pub trait TupleCollect: Sized {
 }
 
 macro_rules! impl_tuple_collect {
-    () => ();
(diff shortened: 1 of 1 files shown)


jswrenn:
[Criterion strikes again!](https://travis-ci.org/github/rust-itertools/itertools/jobs/692920021) (Or, rather, a dependency of it.)

Perhaps we could add a `bench` default feature, and only pull `criterion` if that feature is present. Then, we modify our CI to use `--no-default-features` for the MSRV build.

phimuemue:
Ah, I see.

I think your Idea is good, even though I have no experience in crate features. And while skimming https://doc.rust-lang.org/cargo/reference/features.html#rules, I read 

> Dev-dependencies cannot be optional.

I am not sure if this affects us.

Issue #388, opened 2019-12-05
Add pairwise()


Kinrany:
RxJS and python's itertools have `pairwise`. It's an operator that yields every consecutive pair. For example, `[1, 2, 3, 4]` will become `[(1, 2), (2, 3), (3, 4)`.

RxJS: https://rxjs.dev/api/operators/pairwise \
Python: https://docs.python.org/3/library/itertools.html (not linkable, please search for "pairwise")

I've already implemented it for my own use. I'd be happy to contribute a RP.

jswrenn:
This sounds similar (equivalent?) to [`Itertools::tuples`](https://docs.rs/itertools/0.8.2/itertools/trait.Itertools.html#method.tuples). Does that fit your use case?

Kinrany:
@jswrenn no, they're similar but different:

```
tuples_2: [1, 2, 3, 4] -> [(1, 2), (3, 4)]
pairwise: [1, 2, 3, 4] -> [(1, 2), (2, 3), (3, 4)]
```

jswrenn:
[`Itertools::tuple_windows`](https://docs.rs/itertools/0.8.2/itertools/trait.Itertools.html#method.tuple_windows)?

Kinrany:
Indeed, thanks :)

Kinrany:
I wonder if it would make sense to add pairwise to the docs, so that it's searchable?

robinmoussu:
@Kinrany I think it does. I had the exact same issue, and I wasn't able to find it using the search engine of docs.rs, duckduckgo and google. If I didn't searched in the closed issue, I would have opened one. I created a PR, feel free to modify it ;)

George-Ogden:
I know this was a while ago, but is it worth adding a pairwise method as a convenience function? It would also make the code more readable for people skimming existing Rust code.

Kinrany:
As an alias for `tuple_windows` perhaps? That would make it searchable in docs.

George-Ogden:
I think it would be nice if it could somehow restrict to a 2-tuple/pair, so that you can't write:
```rust
for (a, b, c) in my_list.pairwise() {
   ...
}
```

George-Ogden:
@Kinrany I've opened #1111, do you think I should make any changes?

phimuemue:
Closing as per https://github.com/rust-itertools/itertools/pull/1111#issuecomment-4779199918

**edit:** Thanks @Kinrany for bringing doc-alias to my attention.

[...]

PR #449, opened 2020-06-11
Improve `tuple_windows()` doc (add a reference to `pairwise`)

robinmoussu:
As shown in #388, tuple_windows can be hard to find if you come from other
environment (python, RxJs, …) since people may expect to find a `pairwise()`
method instead.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -561,6 +561,8 @@ pub trait Itertools : Iterator {
     /// ```
     /// use itertools::Itertools;
     /// let mut v = Vec::new();
+    ///
+    /// // pairwise iteration
     /// for (a, b) in (1..5).tuple_windows() {
     ///     v.push((a, b));
     /// }


jswrenn:
The wording makes it sound a little like rust-itertools has a `pairwise` method. Could you clarify the wording? I'm not exactly sure what the right wording is, since python itertools also technically lacks a `pairwise` method.

Perhaps instead just add a comment to the existing example in the `tuple_windows` doc comment; e.g.:
```
/// use itertools::Itertools;
/// let mut v = Vec::new();
///
/// // pairwise iteration:
/// for (a, b) in (1..5).tuple_windows() {
///     v.push((a, b));
/// }
```

robinmoussu:
I agree, and updated the PR accordingly.

> since python itertools also technically lacks a pairwise method.

That's right, I forgot it was in `more_itertools`.

PR #452, opened 2020-06-21
add try_iterate: iterate function until it returns None

yihuang:


scottmcm:
Given that [`unfold`](https://docs.rs/itertools/0.9.0/itertools/fn.unfold.html) already exists, it's not clear to me that this is worth having.

yihuang:
> Given that [`unfold`](https://docs.rs/itertools/0.9.0/itertools/fn.unfold.html) already exists, it's not clear to me that this is worth having.

I didn't know the `unfold`, I guess one can basically implement the `try_iterate` with `unfold` like this:
```rust
pub fn try_iterate<St, F>(initial_value: St, mut f: F) -> impl Iterator<Item = St>
where
    St: Clone,
    F: FnMut(&St) -> Option<St>,
{
    iter::once(initial_value.clone()).chain(itertools::unfold(initial_value, move |v| match f(&*v) {
        Some(r) => {
            *v = r.clone();
            Some(r)
        }
        None => None,
    }))
}
```

The only caveat is "St: Clone`.

phimuemue:
Hi! There is already [`successors`](https://doc.rust-lang.org/std/iter/fn.successors.html), which looks very much like your proposed function. Thus, I suggest to close this PR.

yihuang:
> Hi! There is already [`successors`](https://doc.rust-lang.org/std/iter/fn.successors.html), which looks very much like your proposed function. Thus, I suggest to close this PR.

Thanks for the notice, that's exactly what I wanted.

PR #454, opened 2020-06-24
Cleanup some clippy lints

phimuemue:
I cleaned up some things I got from clippy. I only changed what I'd consider straightforward.

I think it would eventually be beneficial to resolve the remaining lints, so that it is easier to identify real problems spotted by clippy. Do we have a policy regarding this?

Off-topic: I would have merged this right away, but I did not want to introduce conflicts with existing PRs. Is there a way to check if my commits would introduce conflicts?

### DIFF
--- a/src/adaptors/mod.rs
+++ b/src/adaptors/mod.rs
@@ -113,22 +113,11 @@ impl<I, J> Iterator for InterleaveShortest<I, J>
 
     #[inline]
     fn next(&mut self) -> Option<Self::Item> {
-        match self.phase {
-            false => match self.it0.next() {
-                None => None,
-                e => {
-                    self.phase = true;
-                    e
-                }
-            },
-            true => match self.it1.next() {
-                None => None,
-                e => {
-                    self.phase = false;
-                    e
-                }
-            },
+        let e = if self.phase { self.it1.next() } else { self.it0.next() };
+        if e.is_some() {
+            self.phase = !self.phase;
         }
+        e
     }
 
     #[inline]
--- a/src/adaptors/mod.rs
+++ b/src/adaptors/mod.rs
@@ -412,7 +412,7 @@ impl<B, F, I> Iterator for Batching<I, F>
 /// then skipping forward *n-1* elements.
 ///
 /// See [`.step()`](../trait.Itertools.html#method.step) for more information.
-#[deprecated(note="Use std .step_by() instead", since="0.8")]
+#[deprecated(note="Use std .step_by() instead", since="0.8.0")]
 #[allow(deprecated)]
 #[derive(Clone, Debug)]
 #[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
--- a/src/cons_tuples_impl.rs
+++ b/src/cons_tuples_impl.rs
@@ -57,8 +57,8 @@ impl<I, J> Clone for ConsTuples<I, J>
 
 /// Create an iterator that maps for example iterators of
 /// `((A, B), C)` to `(A, B, C)`.
-pub fn cons_tuples<I, J>(iterable: I) -> ConsTuples<I, J>
-    where I: Iterator<Item=J>
+pub fn cons_tuples<I, J>(iterable: I) -> ConsTuples<I::IntoIter, J>
+    where I: IntoIterator<Item=J>
 {
     ConsTuples { iter: iterable.into_iter() }
 }
--- a/src/group_map.rs
+++ b/src/group_map.rs
@@ -15,7 +15,7 @@ pub fn into_group_map<I, K, V>(iter: I) -> HashMap<K, Vec<V>>
     let mut lookup = HashMap::new();
 
     for (key, val) in iter {
-        lookup.entry(key).or_insert(Vec::new()).push(val);
+        lookup.entry(key).or_insert_with(Vec::new).push(val);
     }
 
     lookup
@@ -30,5 +30,3 @@ pub fn into_group_map_by<I, K, V>(iter: I, f: impl Fn(&V) -> K) -> HashMap<K, Ve
         iter.map(|v| (f(&v), v))
     )
 }
-
-
--- a/src/tuple_impl.rs
+++ b/src/tuple_impl.rs
@@ -212,7 +212,7 @@ pub fn circular_tuple_windows<I, T>(iter: I) -> CircularTupleWindows<I, T>
     let iter = tuple_windows(iter.cycle()).take(len);
 
     CircularTupleWindows {
-        iter: iter,
+        iter,
         phantom_data: PhantomData{}
     }
 }
--- a/src/tuple_impl.rs
+++ b/src/tuple_impl.rs
@@ -57,7 +57,7 @@ impl<T> Iterator for TupleBuffer<T>
 
     fn size_hint(&self) -> (usize, Option<usize>) {
         let buffer = &self.buf.as_ref()[self.cur..];
-        let len = if buffer.len() == 0 {
+        let len = if buffer.is_empty() {
             0
         } else {
             buffer.iter()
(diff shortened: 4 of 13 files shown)


jswrenn:
> I think it would eventually be beneficial to resolve the remaining lints, so that it is easier to identify real problems spotted by clippy. Do we have a policy regarding this?

Go for it!

> Off-topic: I would have merged this right away, but I did not want to introduce conflicts with existing PRs. Is there a way to check if my commits would introduce conflicts?

I don't know of any way to quickly judge whether a PR introduces merge conflicts with other PRs. Before merging big PRs, I usually scan through the list of other PRs and see if there's any low-hanging fruit ready to be merged.

I'm currently on my honeymoon and COVID's made my work weird (I do education research), so I'm short on time. If you have the desire and time to knock down the list of pending PRs, feel free to go for it!

phimuemue:
> I'm currently on my honeymoon

Hey, congratulations! Nice to hear this - I hope you're having a great time.

> and COVID's made my work weird (I do education research), so I'm short on time.

Wish you all the best for your work.

jplatte:
Bump.

phimuemue:
> Bump.

@jplatte Have you thoroughly reviewed this? (I'd prefer a second pair of eyes on this, in particular the match bool => if/else?)

If yes, and if @jswrenn has no objections, I'd merge this.

src/tuple_impl.rs
```diff
@@ -57,12 +57,12 @@ impl<T> Iterator for TupleBuffer<T>
...
     fn size_hint(&self) -> (usize, Option<usize>) {
         let buffer = &self.buf.as_ref()[self.cur..];
-        let len = if buffer.len() == 0 {
+        let len = if buffer.is_empty() {
             0
         } else {
             buffer.iter()
                   .position(|x| x.is_none())
-                  .unwrap_or(buffer.len())
+                  .unwrap_or_else(|| buffer.len())
```
jswrenn, line 65: Calling`.len` is *really* cheap. Adding the indirection of a thunk will make more work for the compiler, with probably little (to no) gain in performance.

src/cons_tuples_impl.rs
```diff
@@ -57,8 +57,8 @@ impl<I, J> Clone for ConsTuples<I, J>
 
 /// Create an iterator that maps for example iterators of
 /// `((A, B), C)` to `(A, B, C)`.
-pub fn cons_tuples<I, J>(iterable: I) -> ConsTuples<I, J>
-    where I: Iterator<Item=J>
+pub fn cons_tuples<I, J>(iterable: I) -> ConsTuples<I::IntoIter, J>
+    where I: IntoIterator<Item=J>
```
jplatte, line 61: This seems extremely unlikely to break existing code, but I'm sure one could write code that breaks from this change. Not sure whether that's important.

jplatte:
> Have you thoroughly reviewed this?

I have now :)

The bool matching change looks good, there's one tiny nit above, but I think this can be merged.

src/cons_tuples_impl.rs
phimuemue: I thought this change only makes `cons_tuples` more general as per [`impl<I: Iterator> IntoIterator for I`](https://doc.rust-lang.org/std/iter/trait.IntoIterator.html#impl-IntoIterator-25).

src/tuple_impl.rs
phimuemue: Fair point. When encountering such situations in own projects, I'm inclined to go with clippy, because when this warning does not show up anymore, I can focus more easily on real issues in subsequent clippy runs.

If you still prefer that I should remove it, please let confirm this, and I'll revert it.

src/cons_tuples_impl.rs
jplatte: Yeah, that's what I was thinking of too. Pretty much everything that affects public API is a breaking change for inference in some way.

scottmcm: I think it doesn't.  The usual inference breaks:
- Adding a second implementation of a trait with a generic (`AsRef` and `PartialEq` are common offenders)
- Changing from a concrete type to a trait (`&` to `impl Borrow`, particularly)

In this case there's the blanket impl, so if you had enough information to pick an Iterator before, that's enough to pick an IntoIterator (because it's the same type), AFAICT.

src/tuple_impl.rs
jplatte: > I'm inclined to go with clippy, because when this warning does not show up anymore, I can focus more easily on real issues in subsequent clippy runs

I think if this change is reverted, the surrounding functions should have `#[allow(clippy::lint_name)]`.

src/cons_tuples_impl.rs
jswrenn: Yeah, this is harmless. Minimal example that compiles fine regardless of the bound on `Foo`:
```rust
mod itertools {
    pub trait IntoIterator {}
    
    pub trait Iterator {}
    
    impl<T: Iterator> IntoIterator for T {}
    
    pub fn foo<I>()
    where
        I: /*Iterator*/ IntoIterator
    {}
}

mod dependent_crate {
    pub use super::itertools::*;
    
    pub fn bar<I>()
    where
        I: Iterator
    {
        foo::<I>()
    }
}
```

src/tuple_impl.rs
jswrenn: I'm fine with `unwrap_or_else`, I think. This `len()` is coming from `ExactSizeIterator`, so the call isn't *necessarily* cheap.

src/group_map.rs
```diff
@@ -15,7 +15,7 @@ pub fn into_group_map<I, K, V>(iter: I) -> HashMap<K, Vec<V>>
     let mut lookup = HashMap::new();
 
     for (key, val) in iter {
-        lookup.entry(key).or_insert(Vec::new()).push(val);
+        lookup.entry(key).or_insert_with(Vec::new).push(val);
```
jswrenn, line 18: `Vec::new` is *extremely* cheap, and shouldn't be thunked. This almost makes me want to file a clippy bug...

src/adaptors/mod.rs
```diff
@@ -113,21 +113,22 @@ impl<I, J> Iterator for InterleaveShortest<I, J>
...
                 None => None,
                 e => {
-                    self.phase = false;
+                    self.phase = true;
                     e
                 }
-            },
+            }
         }
     }
```
jswrenn, line 133: I think this method could be written much more succinctly as:
```rust
    #[inline]
        let it = if self.phase { self.it1 } else { self.it0 };
        let e = it.next();

        if e.is_some() {
            self.phase = !self.phase;
        }

        e
    }
```

phimuemue: I thought the same, but `it1` and `it0` have different types.

src/group_map.rs
scottmcm: Hmm, can clippy tell whether it's `const fn`?  Would be nice to disable that link if the RHS is a const.

phimuemue: @jswrenn I agree that we should avoid unnecessary work for the compiler where possible. However, thunking a `Vec::new` has (significant?) run-time advantages [in debug mode](https://godbolt.org/z/jfjnv9)...  maybe nice to have.

[Release mode](https://godbolt.org/z/TqYvTE) brings them closer together (possibly equivalent in run-time).

src/adaptors/mod.rs
jswrenn: D'oh. How about this, then:

```
    #[inline]
        let e = if self.phase { self.it1.next() } else { self.it0.next() };

        if e.is_some() {
            self.phase = !self.phase;
        }

        e
    }
```

phimuemue: Nice. I took this.

jswrenn:
This looks good to me. Ready to merge, @phimuemue?

PR #455, opened 2020-06-25
Unify documentation

### LINKED ISSUE #448: Interleave doc should explains how it is different from `std::iter::zip`
When reading the documentation of [interleave](https://docs.rs/itertools/0.9.0/itertools/fn.interleave.html) I wasn't able to understand what is the difference with [`std::iter::zip()`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.zip), and why/when should one be preferred to the other.

robinmoussu:
Fix #448


jswrenn:
Thanks!

bors r+

PR #456, opened 2020-07-02
Add Itertools::filter_common.

joshwd36:
This came about due to a situation where I needed to remove different combinations of items from an iterator, with the catch that I only wanted to remove the exact values, so duplicates were not all removed, meaning I couldn't simply filter. I'll admit this is fairly niche, and there are still a few outstanding questions I have:

- I'm fairly unsure of the terminology used here. I've tried to search around for similar concepts but I couldn't find anything that wasn't more analagous to a simple filter. If anyone has any other suggestions I'd be very appreciative. Even thinking about it now it should probably be something along the lines of filter_exclusive.
- It seems logical that when removing items from the filter list it would be more efficient to remove the last item, rather than the first if duplicates are present, but I haven't done any benchmarking and I'm not sure if it's worth the additional complexity, although admittedly it shouldn't be too bad.
- I don't think there's another adaptor that takes a Vec as an argument to the method, and while it would be more efficient in the case where the filter list is constructed as a vector to take it as an argument, rather than converting to an iterator and back, I realise there isn't a precedent for this.

### DIFF
--- a/src/filter_common.rs
+++ b/src/filter_common.rs
@@ -17,6 +17,7 @@ impl<I, K> FilterCommon<I, K>
         I: Iterator,
         K: PartialEq<I::Item>,
 {
+    /// Remove the first instance of an item from the filter list, returning true if an item was removed.
     fn remove_item(&mut self, item: &I::Item) -> bool {
         if let Some(index) = self.filter_list.iter().position(|x| x == item) {
             self.filter_list.remove(index);
@@ -47,7 +48,10 @@ impl<I, K> Iterator for FilterCommon<I, K>
 
     fn next(&mut self) -> Option<Self::Item> {
         loop {
+            // Return None if we've reached the end of the iterator.
             let item = self.iterator.next()?;
+            // If the element from the iterator doesn't match an element in the filter list, return true.
+            // Otherwise get the next element.
             if !self.remove_item(&item) {
                 return Some(item);
             }
--- a/src/filter_common.rs
+++ b/src/filter_common.rs
@@ -1,0 +1,56 @@
+/// Filters items from the original iterator that exist in the filter list.
+///
+/// Iterator element type is `I::Item`.
+///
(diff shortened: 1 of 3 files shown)


phimuemue:
Hi there. I have some problems understanding the motivation here.

* Couldn't all this be implemented via `filter` (which accepts an `FnMut` that could do the vector stuff inernally (`Vec` already has unstable `remove_item`))? I am not sure that `filter_common` is such a major usability improvement to a plain old `filter`. Or am I missing something?
* The other problem I see is that hardcoding a `Vec` into this iterator is probably in most cases not as efficient as it could be. (Since many things can be used in a hash-like or sorted data structure.)

joshwd36:
> 
> 
> Hi there. I have some problems understanding the motivation here.
> 
>     * Couldn't a user just call `filter` (which accepts an `FnMut` that could do the vector stuff inernally (`Vec` already has unstable `remove_item`))? I am not sure that `filter_common` is such a major usability improvement to a plain old `filter`. Or am I missing something?
> 
>     * The other problem I see is that hardcoding a `Vec` into this iterator is probably in most cases not as efficient as it could be. (Since many things can be used in a hash-like or sorted data structure.)
> 
>     * Side note: In case it gets accepted, I think we should avoid the indirection via `remove_item` (and just inline it).

I think that's all very fair, and the point about hardcoding a Vec makes me think that it probably would be reasonable to close it. The only thing that does make it difficult to manually implement this kind of function is that remove_item is deprecated and will be removed, so it would always have to be repeated by the user.

phimuemue:
I see that the user probably would have to implement `remove_item` themselves, but I would say this functionality is not iterator-specific and could be needed in other circumstances, too. So I guess it could/should be solved independently of an iterator package.

phimuemue:
> it probably would be reasonable to close it

@joshwd36 Do you have objections against closing this?

joshwd36:
> 
> 
> > it probably would be reasonable to close it
> 
> @joshwd36 Do you have objections against closing this?

No, that's fine

PR #459, opened 2020-07-13
Clarify documentation for `merge_by`

phimuemue:
I think simply stating the elements is easier than going through `map`.


jswrenn:
I agree!

bors r+

Issue #413, opened 2020-02-19
Maintenance


jplatte:
Hi! It seems to me that itertools is quite low on maintenance power – a fairly big amount of PRs have the waiting-on-review tag, some are untagged and some issues also haven't received any comments.

Is there anything I can do to help?

Veetaha:
@bluss, @phimuemue, @jswrenn could you clarify this?

jswrenn:
When we receive a PR for a new method, we need to resolve:
- [ ] Without the addition, how would the operation be expressed with Itertools and the standard library? Is it worth adding to Itertools?
- [ ] Is the name intuitive?
- [ ] If it's an adaptor: ensure that constructing the adaptor on an infinite iterator will terminate.
- [ ] Will the implementation ever panic? Can this be avoided? Is this documented?
- [ ] Does it behave sensibly on edge-cases? Is its behavior documented?
- [ ] Does it have illustrative doctests?
- [ ] Can the implementation be simplified, or consolidated with other parts of Itertools?

If the contribution is an optimized specialization of an `Iterator` method (such as `nth` or `fold`), the specialization needs to be very thoroughly tested (preferably with quickcheck tests) to ensure it behaves identically (_especially_ on edge cases, and with when it panics).

Reviews that resolve any of the above are very helpful!

The bar for inclusion in itertools is much lower than it is for the standard library, but we'd still like to know that additions will be helpful to more than just the person submitting it. If there's a PR proposing something you need, leave a note describing your use-case. This feedback is helpful for prioritizing PRs, too.

jplatte:
Hm… I think I'm more concerned about things that aren't new features missing reviews. Of the 23 PRs that are currently open, only 8 are labeled 'waiting on review'. There's also lots of feature requests that have no answers from a maintainer. I usually ask whether a PR is welcome before writing the code. Should I not do that here? Should I just ping somebody if I don't get an answer after a while?

PR #461, opened 2020-07-14
Clean up specializations tests

phimuemue:
In response to this comment https://github.com/rust-itertools/itertools/issues/413#issuecomment-657670781:

> If the contribution is an optimized specialization of an `Iterator` method (such as `nth` or `fold`), the specialization needs to be very thoroughly tested (preferably with quickcheck tests) to ensure it behaves identically (_especially_ on edge cases, and with when it panics).

I thought about having a simple way to test specializations: If we had *one* test function testing all specializations, we would have an easy way to quickcheck specializations for our iterators.

We already had quite a bit in place, so I took it and streamlined it a bit.

* `fold` is tested by remembering all elements provided to the closure (instead of only XORing elements).
* cleaned up some things (unused specialization of `size_hint`, some redundant lifetimes, superfluous arguments)
* test all specializations in one function (instead of only some specializations in two functions)
* remove manual special case tests in favor of quickcheck'd ones

If we decide to do it like this, we can, in a future commit, just do the following for all iterators (possibly via a macro) to test specializations uniformly (and hopefully reliably):

```
quickcheck! {
    fn test_specializations_qc(params_for_iterator) -> () {
        test_specializations(&make_iterator_to_be_tested(params_for_iterator));
    }
}
```

This way, we would also test methods that are not even specialized, but for simplicity, I'd be willing to accept this. If we see that it slows things down considerably, we could still devise a simple solution that would allow to specify the methods to be tested.

### DIFF
--- a/tests/specializations.rs
+++ b/tests/specializations.rs
@@ -18,7 +18,6 @@ where
 fn check_specialized<'a, V, IterItem, Iter, F>(iterator: &Iter, mapper: F)
 where
     V: Eq + Debug,
-    IterItem: 'a,
     Iter: Iterator<Item = IterItem> + Clone + 'a,
     F: Fn(Box<dyn Iterator<Item = IterItem> + 'a>) -> V,
 {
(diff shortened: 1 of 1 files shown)


jswrenn:
This is slick!

bors r+

PR #463, opened 2020-07-19
Implement intersperse in terms of intersperse_with

phimuemue:
This de-copy-pastes Intersperse/IntersperseWith, similar to what we do in other iterator adaptors.


jswrenn:
This looks good to me!

bors r+

PR #464, opened 2020-07-19
Unify convenience map functions

phimuemue:
In response to https://github.com/rust-itertools/itertools/pull/283: I think we should employ a uniform architecture for our `map`-specializations (as they already diverged). This PR is meant to unify the implementations so we essentially only need to parametrize the mapping function:

* Generalize `MapInto` to `MapSpecialCase` (which can be used as basis for all our map convenience functions) and define `MapInto` in terms of `MapSpecialCase`
* Specialize `collect` (was specialized on `MapOk`, so I guess we want to keep it for now)
* Define `MapOk` in terms of `MapSpecialCase`
* Simple tests for specialized `size_hint`, `fold`, `collect`
* As at this point probably all existing PRs involving `map_xyz` are conflicting anyway, move them to an own module (similar to `coalesce`)

### DIFF
--- a/src/adaptors/map.rs
+++ b/src/adaptors/map.rs
@@ -1,0 +1,120 @@
+use std::iter::FromIterator;
+use std::marker::PhantomData;
+
+#[derive(Clone)]
+#[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
+pub struct MapSpecialCase<I, F> {
+    iter: I,
+    f: F,
+}
+
+pub trait MapSpecialCaseFn<T> {
+    type Out;
+    fn call(&mut self, t: T) -> Self::Out;
+}
+
+impl<I, R> Iterator for MapSpecialCase<I, R>
+where
+    I: Iterator,
+    R: MapSpecialCaseFn<I::Item>,
+{
+    type Item = R::Out;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.iter.next().map(|i| self.f.call(i))
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        self.iter.size_hint()
+    }
+
+    fn fold<Acc, Fold>(self, init: Acc, mut fold_f: Fold) -> Acc
+    where
+        Fold: FnMut(Acc, Self::Item) -> Acc,
+    {
+        let mut f = self.f;
+        self.iter.fold(init, move |acc, v| fold_f(acc, f.call(v)))
+    }
+
+    fn collect<C>(self) -> C
+    where
+        C: FromIterator<Self::Item>,
+    {
+        let mut f = self.f;
+        self.iter.map(move |v| f.call(v)).collect()
+    }
+}
+
+impl<I, R> DoubleEndedIterator for MapSpecialCase<I, R>
+where
+    I: DoubleEndedIterator,
+    R: MapSpecialCaseFn<I::Item>,
+{
+    fn next_back(&mut self) -> Option<Self::Item> {
+        self.iter.next_back().map(|i| self.f.call(i))
+    }
+}
+
+impl<I, R> ExactSizeIterator for MapSpecialCase<I, R>
+where
+    I: ExactSizeIterator,
+    R: MapSpecialCaseFn<I::Item>,
+{
+}
+
+/// An iterator adapter to apply a transformation within a nested `Result::Ok`.
+///
+/// See [`.map_ok()`](../trait.Itertools.html#method.map_ok) for more information.
+pub type MapOk<I, F> = MapSpecialCase<I, MapSpecialCaseFnOk<F>>;
+
+/// See [`MapOk`](struct.MapOk.html).
+#[deprecated(note = "Use MapOk instead", since = "0.10")]
+pub type MapResults<I, F> = MapOk<I, F>;
+
+impl<F, T, U, E> MapSpecialCaseFn<Result<T, E>> for MapSpecialCaseFnOk<F>
+where
+    F: FnMut(T) -> U,
+{
+    type Out = Result<U, E>;
+    fn call(&mut self, t: Result<T, E>) -> Self::Out {
+        t.map(|v| self.0(v))
+    }
+}
+
+#[derive(Clone)]
+pub struct MapSpecialCaseFnOk<F>(F);
+
+/// Create a new `MapOk` iterator.
+pub fn map_ok<I, F, T, U, E>(iter: I, f: F) -> MapOk<I, F>
+where
+    I: Iterator<Item = Result<T, E>>,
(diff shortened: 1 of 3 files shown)


src/adaptors/map.rs
```diff
@@ -0,0 +1,120 @@
+use std::iter::FromIterator;
+use std::marker::PhantomData;
+
+#[derive(Clone)]
+#[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
+pub struct MapSpecialCase<I, F> {
```
pthariensflame, line 6:
```suggestion
pub struct MapSpecialCase<I, F: ?Sized> {
```

This way unsized “special functions” can be used, including trait objects.

src/adaptors/map.rs
```diff
@@ -0,0 +1,120 @@
...
+
+pub trait MapSpecialCaseFn<T> {
+    type Out;
+    fn call(&mut self, t: T) -> Self::Out;
+}
+
+impl<I, R> Iterator for MapSpecialCase<I, R>
+where
+    I: Iterator,
+    R: MapSpecialCaseFn<I::Item>,
```
pthariensflame, line 19:
```suggestion
    R: ?Sized + MapSpecialCaseFn<I::Item>,
```

src/adaptors/map.rs
```diff
@@ -0,0 +1,120 @@
...
+{
+    fn next_back(&mut self) -> Option<Self::Item> {
+        self.iter.next_back().map(|i| self.f.call(i))
+    }
+}
+
+impl<I, R> ExactSizeIterator for MapSpecialCase<I, R>
+where
+    I: ExactSizeIterator,
+    R: MapSpecialCaseFn<I::Item>,
```
pthariensflame, line 61:
```suggestion
    R: ?Sized + MapSpecialCaseFn<I::Item>,
```

src/adaptors/map.rs
```diff
@@ -0,0 +1,120 @@
...
+    {
+        let mut f = self.f;
+        self.iter.map(move |v| f.call(v)).collect()
+    }
+}
+
+impl<I, R> DoubleEndedIterator for MapSpecialCase<I, R>
+where
+    I: DoubleEndedIterator,
+    R: MapSpecialCaseFn<I::Item>,
```
pthariensflame, line 51:
```suggestion
    R: ?Sized + MapSpecialCaseFn<I::Item>,
```

src/adaptors/map.rs
```diff
@@ -0,0 +1,120 @@
...
+    F: FnMut(T) -> U,
+{
+    type Out = Result<U, E>;
+    fn call(&mut self, t: Result<T, E>) -> Self::Out {
+        t.map(|v| self.0(v))
+    }
+}
+
+#[derive(Clone)]
+pub struct MapSpecialCaseFnOk<F>(F);
```
pthariensflame, line 85:
```suggestion
pub struct MapSpecialCaseFnOk<F: ?Sized>(F);
```

src/adaptors/map.rs
```diff
@@ -0,0 +1,120 @@
...
+/// See [`.map_ok()`](../trait.Itertools.html#method.map_ok) for more information.
+pub type MapOk<I, F> = MapSpecialCase<I, MapSpecialCaseFnOk<F>>;
+
+/// See [`MapOk`](struct.MapOk.html).
+#[deprecated(note = "Use MapOk instead", since = "0.10")]
+pub type MapResults<I, F> = MapOk<I, F>;
+
+impl<F, T, U, E> MapSpecialCaseFn<Result<T, E>> for MapSpecialCaseFnOk<F>
+where
+    F: FnMut(T) -> U,
```
pthariensflame, line 76:
```suggestion
    F: ?Sized + FnMut(T) -> U,
```

src/adaptors/map.rs
```diff
@@ -0,0 +1,120 @@
...
+}
+
+#[derive(Clone)]
+pub struct MapSpecialCaseFnOk<F>(F);
+
+/// Create a new `MapOk` iterator.
+pub fn map_ok<I, F, T, U, E>(iter: I, f: F) -> MapOk<I, F>
+where
+    I: Iterator<Item = Result<T, E>>,
+    F: FnMut(T) -> U,
```
pthariensflame, line 91:
```suggestion
    F: ?Sized + FnMut(T) -> U,
```

src/adaptors/mod.rs
```diff
@@ -5,8 +5,10 @@
 //! except according to those terms.
 
 mod coalesce;
+mod map;
 mod multi_product;
 pub use self::coalesce::*;
+pub use self::map::*;
```
jswrenn, line 11: Does `MapSpecialCase` need to be part of itertools's public API? Instead of glob importing everything from `map`, perhaps just import the exact items we want to be part of our public API.

phimuemue: Good idea.

src/adaptors/map.rs
phimuemue: Thanks for this nice idea. In general, I am leaning to incorporate changes that generalize functionality.

However, I'd like to postpone this one and - if we decide to accept `?Sized` functions - do it uniformly throughout itertools. I think my changes did not change behavior in this regard, so I hope postponing is ok.

@pthariensflame As I'm no expert in this area: Could you showcase a situation that requires `F: ?Sized`?

pthariensflame: Sure!  If we have an unsized trait object `dyn Fn(X) -> Y`, then we get an unsized `MapSpecialCaseFnOk<dyn Fn(X) -> Y>`, which is then your example of an unsized `F`.

jswrenn: > However, I'd like to postpone this one and - if we decide to accept ?Sized functions - do it uniformly throughout itertools. I think my changes did not change behavior in this regard, so I hope postponing is ok.

Fine by me!

jswrenn: @pthariensflame I'm curious about the actual use-case `dyn Fn` might appear in. I've never needed to dynamically dispatched function objects before.

pthariensflame: Yeah, I'm fine with waiting too.  This isn't essential functionality, just mildly helpful at best.

As far as `dyn Fn`, I can't point to any specific code, but it might come up when abstracting over dynamically swappable actions.  For example:  imagine a pipeline of scriptable blocks that the user can specify in a GUI, and it gets compiled down to functions that invoke the scripts and munge the values flowing through as necessary.  The whole pipeline is then composed and boxed up as a `dyn Fn` so that it can be altered on the fly as the application runs.

jswrenn:
This looks good to me!

bors r+

Issue #457, opened 2020-07-02
Kotlin like lazy Grouping API


SkiFire13:
I'm opening this to request/discuss a possible Grouping API like Kotlin's `Grouping`/`groupingBy` API. See Kotlin's [Grouping reference](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-grouping/) and its [source code](https://github.com/JetBrains/kotlin/blob/master/libraries/stdlib/src/kotlin/collections/Grouping.kt).

It would provide an easy way to perform efficient group-and-fold operations without allocating additional `Vec`s (unlike `into_group_map`), while also maintaning unique keys (unlike `group_by`). The disadvantage is that it can only support `fold`-like operations.

The implementation should be pretty simple, I've already written a pretty minimal working sample that includes:

- a `fold` function, like a normal fold except it associate each key with the result of folding each element that maps to that key;
- a `fold_first` function, like `fold` but without an initial value, instead the first element acts as the first value;
- a `count` function that associate each key with the number of elements that maps to that key;
- a `aggregate` function that provides the basic functionality we can use to write all the other functions

https://gist.github.com/SkiFire13/a0010ff658d905dcbc1e1f5f6ae910e0

The downsides of my implementation are:

- It can only produce an `HashMap`. This could be solved with a `Map` trait but we're still waiting for GATs for that. It is required because I don't only `extend` the map, but also `remove` elements from it;
- It doesn't provide a way to specify the hasher (it could but then when/if we will switch to a `Map` trait then it would break more things);
- `fold`'s initial value must be `Clone` but I can't see any way around that;
- It could be confused with `group_by` although Kotlin has both `groupBy` and `groupingBy` and I haven't seen someone complaining;
- It pretty much reuses Kotlin's docs and tests but I haven't checked which license the Kotlin stdlib uses;
- It misses code examples from the docs;
- `fold_first` is consistent with the stdlib corrispondent but inconsistent with Itertools' `fold1`.

I'm also thinking of other possible extensions with functions like:
- other flavors of `fold`;
- `max`/`min`/`maxmin` and friends;
- `sum` but I found it would be pretty hard to use the `Sum` trait since it requires a whole iterator and `Grouping` can only fold.

phimuemue:
Hi there! How easy would it be to express these operations in terms of simpler iterator operations?

[...]

Issue #309, opened 2018-09-21
Increase C# group_by


allphfa:
Add a group_by that does not duplicate key.

Philippe-Cholet:
@allphfa Hi, I'm not familiar with C# so I'm wondering if #465 solved your issue (in which case I would close this). If not, provide more intel would be appreciated.

Issue #358, opened 2019-07-21
Should minmax require Ord instead of PartialOrd


phimuemue:
I think requiring only `PartialOrd` instead of `Ord` is not optimal for `minmax`:

```
    // Concern 1: different behaviour than min/max
    let vec = [std::f64::NAN, 1.];
    //vec.iter().min(); // does not compile (for good reasons imho)
    //vec.iter().max(); // does not compile (for good reasons imho)
    vec.iter().minmax(); // compiles, which is surprising
    
    // Concern 2: Order of elements is relevant
    println!("{:?}", vec![std::f64::NAN, 1.].iter().minmax()); // (NaN, 1.0)
    println!("{:?}", vec![1., std::f64::NAN].iter().minmax()); // (1.0, NaN)
```

The [above code](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=afef02f940fd145c3340e4569e710dc8) highlights my concerns:

* It is inconsistent with `min`/`max`.
* The order of the elements plays too important a role. This is particularly problematic for e.g. iterators visiting iterating elements in an arbitrary order (such as [`HashMap::keys`](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.keys)).

The documentation clearly states that 

> The keys can be floats but no particular result is guaranteed if a key is NaN.

but the problem is present not only for floats, but for anything that does not fulfill `Ord`.

Wouldn't it be clearer to require `Ord`, thereby leaving the responsibility of dealing with `PartialOrd` to the callers? I know this is a breaking change

Avi-D-coder:
I would advocate, just adding another note in the docs .

bluss:
The decision to use PartialOrd is concious, so that the caller can be judicious themselves. I think using `Ord` mostly has theoretical benefits, can you maybe explain what concrete problem you see with the current situation?

phimuemue:
To be honest, I did not follow the discussion that lead to having `PartialOrd`, so I may be missing important arguments here.

However, I am a bit wary when it comes to problematic situations that could be prevented by the compiler. (I think silently failing `float` comparisons are problematic sometimes.) I remember debugging C++ code resulting from `NaN` comparisons, and thus being glad that Rust informed me that I could not simply use `min` on `float`s, leaving me the possibility to choose `min_by` and decide what to do for `NaN` myself.

My argument would be similar for `minmax`.

[...]

PR #465, opened 2020-07-22
Add into_grouping_map for efficient group-and-fold operations

SkiFire13:
Adds two functions on the `Itertools` trait, `into_grouping_map` and `into_grouping_map_by`. 

`into_grouping_map` expects an iter of `(K, V)` where the `K` will be used as key and `V` as value. 
`into_grouping_map_by` expect only an iter of `V` as values, the keys will be calculated using the provided functions. 

Both of them return a `GroupingMap`, which is just a wrapper on an iterator. Since it introduces a lot of related methods I thought it would be better to separate them from the `Itertools` trait. This also prevents duplicating every method for the `_by` version.

All of these functions have in common the fact they perform efficient group-and-fold operations without allocating temporary vecs like you would normally do if you used `into_group_map` + `into_iter` + `map` + `collect::<HashMap<_, _>>()`.

Here's the possible issues I can see, I would like to hear some feedback before trying to fix any of them:
- name: I initially though of `grouping_by` which is the java/kotlin equivalent, but later changed it to `into_grouping_map` to match the already existing `into_group_map` and to differentiate from `group_by`;
- `fold_first`: the equivalent function in the `Itertools` trait is `fold1` but there's an unstable stdlib function that does the same thing and is called `fold_first`. I decided to be consistent with the stdlib;
- `minmax` return type is the already existing `MinMaxResult`, but the `NoElements` variant is never returned. I didn't want to duplicate that struct thinking it could cause confusion (and if I did what name could I have chosen?);
- `sum` and `product`: They dont' use the `Sum` and `Product` traits but instead require `V: Add<V, Output=V>` and `V: Mul<V, Output=V>`. They're pretty much a wrapper around `fold_first`. I don't really know if they're meaningful or if I should just remove them;
- almost every closure takes as one of the parameters a reference to the key. It bloats them a bit, but could be useful if computing the key is a relatively expensive operation.
- no `scan` function. Even though it is an iterator adapter I could sort of "collect" it into a `Vec` (more like extend) but I don't really see an use for this;
- ~~no integration tests for the `_by` and `_by_key` versions of `min`, `max` and `minmax`. To be fair I was a bit lazy, but I also couldn't find any integration test for the normal `minmax_by` and `minmax_by_key` so I though it was fine;~~ added;
- no benchmark (I don't think it is required but probably would be cool to compare this with `into_group_map`;
- I didn't want to touch the rest of the library, but I guess we could implement `into_group_map` in terms of `into_grouping_map`?

Related issues: #457, #309
Related PR: #406 (in particular relates to the `into_group_map_by_fold` function that was proposed but later removed)

### DIFF
--- a/src/grouping_map.rs
+++ b/src/grouping_map.rs
@@ -380,12 +380,32 @@ impl<I, K, V> GroupingMap<I>
         self.min_by(|v1, v2| f(&v1).cmp(&f(&v2)))
     }
 
+    /// Groups elements from the `GroupingMap` source by key and find the maximum and minimum of
+    /// each group.
+    /// 
+    /// See [.minmax_by()](../trait.Itertools.html#method.minmax) for the non-grouping version.
+    /// 
+    /// Differences from the non grouping version:
+    /// - It never produces a `MinMaxResult::NoElements`
+    /// - It doesn't have any speedup
+    /// 
+    /// Returns a `HashMap` associating the key of each group with the minimum and maximum of that group.
     pub fn minmax(self) -> HashMap<K, MinMaxResult<V>>
         where V: Ord,
     {
         self.minmax_by(V::cmp)
     }
 
+    /// Groups elements from the `GroupingMap` source by key and find the maximum and minimum of
+    /// each group with respect to the specified comparison function.
+    /// 
+    /// See [.minmax()](../trait.Itertools.html#method.minmax) for the non-grouping version.
+    /// 
+    /// Differences from the non grouping version:
+    /// - It never produces a `MinMaxResult::NoElements`
+    /// - It doesn't have any speedup
+    /// 
+    /// Returns a `HashMap` associating the key of each group with the minimum and maximum of that group.
     pub fn minmax_by<F>(self, mut compare: F) -> HashMap<K, MinMaxResult<V>>
         where F: FnMut(&V, &V) -> Ordering,
     {
@@ -413,6 +433,16 @@ impl<I, K, V> GroupingMap<I>
         })
     }
 
+    /// Groups elements from the `GroupingMap` source by key and find the elements of each group
+    /// that gives the minimum and maximum from the specified function.
+    /// 
+    /// See [.minmax()](../trait.Itertools.html#method.minmax) for the non-grouping version.
+    /// 
+    /// Differences from the non grouping version:
+    /// - It never produces a `MinMaxResult::NoElements`
+    /// - It doesn't have any speedup
+    /// 
+    /// Returns a `HashMap` associating the key of each group with the minimum and maximum of that group.
     pub fn minmax_by_key<F, CK>(self, mut f: F) -> HashMap<K, MinMaxResult<V>>
         where F: FnMut(&V) -> CK,
               CK: Ord,
--- a/src/grouping_map.rs
+++ b/src/grouping_map.rs
@@ -37,10 +37,8 @@ pub fn new<I, K, V>(iter: I) -> GroupingMap<I>
 }
 
 /// `GroupingMapBy` is an intermediate struct for efficient group-and-fold operations.
-/// It groups elements by the key returned by the specified function and at the same
-/// time fold each group using some aggregating operation.
 /// 
-/// No method on this type performs temporary allocations.
+/// See [`GroupingMap`](./struct.GroupingMap.html) for more informations.
 #[must_use = "GroupingMapBy is lazy and do nothing unless consumed"]
 pub type GroupingMapBy<I, F> = GroupingMap<MapForGrouping<I, F>>;
 
--- a/src/grouping_map.rs
+++ b/src/grouping_map.rs
@@ -30,6 +30,22 @@ where
     I: Iterator<Item = (K, V)>,
     K: Hash + Eq,
 {
+    /// Groups elements from the `GroupingMap` source by key and applies `operation` to the elements
+    /// of each group sequentially, passing the key, the previously accumulated value
+    /// and the current element as arguments, and stores the results in an `HashMap`.
+    ///
+    /// The `operation` function is invoked on each element with the following parameters:
+    ///  - a reference to the key of the group this element belongs to;
+    ///  - the current value of the accumulator of the group or `None` if it's the first element
+    ///    encountered in the group;
+    ///  - the element from the source being aggregated;
+    /// If `operation` returns `Some(element)` then the accumulator is updated with `element`,
+    /// otherwise the previous accumulation is discarded.
+    ///
+    /// Return a `HashMap` associating the key of each group with the result of aggregation of the group elements.
+    ///
+    /// This is the generic way to perform any operations on a `Grouping`.
+    /// It's suggested to use it only to implement custom operations when the already provided ones are not enough.
     pub fn aggregate<FO, R>(self, mut operation: FO) -> HashMap<K, R>
     where
         FO: FnMut(Option<R>, &K, V) -> Option<R>,
--- a/src/grouping_map.rs
+++ b/src/grouping_map.rs
@@ -63,6 +63,7 @@ where
     /// assert_eq!(lookup[&1], 14);
     /// assert!(!lookup.contains_key(&2));
     /// assert_eq!(lookup[&3], 7);
+    /// assert_eq!(lookup.len(), 3);
     /// ```
     pub fn aggregate<FO, R>(self, mut operation: FO) -> HashMap<K, R>
     where
--- a/src/grouping_map.rs
+++ b/src/grouping_map.rs
@@ -380,25 +380,25 @@ impl<I, K, V> GroupingMap<I>
         self.min_by(|v1, v2| f(&v1).cmp(&f(&v2)))
     }
 
-    pub fn minmax(self) -> HashMap<K, V>
+    pub fn minmax(self) -> HashMap<K, MinMaxResult<V>>
         where V: Ord,
     {
         self.minmax_by(V::cmp)
     }
 
-    pub fn minmax_by<F>(self, mut compare: F) -> HashMap<K, V>
+    pub fn minmax_by<F>(self, mut compare: F) -> HashMap<K, MinMaxResult<V>>
         where F: FnMut(&V, &V) -> Ordering,
     {
         self.aggregate(|acc, _, val| {
-            match acc {
-                MinMaxResult::OneElement(e) => {
+            Some(match acc {
+                Some(MinMaxResult::OneElement(e)) => {
                     if compare(&val, &e) == Ordering::Less {
                         MinMaxResult::MinMax(val, e)
                     } else {
                         MinMaxResult::MinMax(e, val)
                     }
                 },
-                MinMaxResult::MinMax(min, max) => {
+                Some(MinMaxResult::MinMax(min, max)) => {
                     if compare(&val, &min) == Ordering::Less {
                         MinMaxResult::MinMax(val, max)
                     } else if compare(&val, &min) == Ordering::Greater {
@@ -406,13 +406,14 @@ impl<I, K, V> GroupingMap<I>
                     } else {
                         MinMaxResult::MinMax(min, max)
                     }
-                }
-                MinMaxResult::NoElements => unreachable!()
-            }
+                },
+                None => MinMaxResult::OneElement(val),
+                Some(MinMaxResult::NoElements) => unreachable!()
+            })
         })
     }
 
-    pub fn minmax_by_key<F, CK>(self, mut f: F) -> HashMap<K, V>
+    pub fn minmax_by_key<F, CK>(self, mut f: F) -> HashMap<K, MinMaxResult<V>>
         where F: FnMut(&V) -> CK,
               CK: Ord,
     {
--- a/src/grouping_map.rs
+++ b/src/grouping_map.rs
@@ -62,6 +62,18 @@ where
         destination_map
     }
 
+    /// Groups elements from the `GroupingMap` source by key and applies `operation` to the elements
+    /// of each group sequentially, passing the previously accumulated value, a reference to the key
+    /// and the current element as arguments, and stores the results in a new map.
+    ///
+    /// `init` is the value from which will be cloned the initial value of each accumulator.
+    ///
+    /// `operation` is a function that is invoked on each element with the following parameters:
+    ///  - the current value of the accumulator of the group;
+    ///  - a reference to the key of the group this element belongs to;
+    ///  - the element from the source being accumulated.
+    ///
+    /// Return a `HashMap` associating the key of each group with the result of folding the group elements.
     pub fn fold<FO, R>(self, init: R, mut operation: FO) -> HashMap<K, R>
     where
         R: Clone,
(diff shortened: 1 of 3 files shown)


src/grouping_map.rs
```diff
@@ -461,7 +461,7 @@ impl<I, K, V> GroupingMap<I>
                 Some(MinMaxResult::MinMax(min, max)) => {
                     if compare(key, &val, &min) == Ordering::Less {
                         MinMaxResult::MinMax(val, max)
-                    } else if compare(key, &val, &max) == Ordering::Greater {
+                    } else if compare(key, &val, &max) != Ordering::Less {
```
jswrenn: This seems like an odd change. Doesn't this render the `else` branch completely unreachable?

SkiFire13: The two `if`s are different. The first compares `val` with `min`, the second with `max`.
The final `else` should be executed when `val` is in between `min` and `max`.

The commit should change the behaviour when `val` is equal to `max`.
Before it was executed the `else` branch meaning `max` was keeped, now the second `if` branch is executed and `val` becomes the new maximum.

This was done to be consistent with `std::iter::max` that states that "If several elements are equally maximum, the last element is returned.". Since `val` comes after `max` it should be considered the maximum.

I think I should add some tests and maybe write this behaviour in the docs to make it clearer.

src/grouping_map.rs
```diff
@@ -0,0 +1,562 @@
...
+                        MinMaxResult::MinMax(min, val)
+                    } else {
+                        MinMaxResult::MinMax(min, max)
+                    }
+                }
+                None => MinMaxResult::OneElement(val),
+                Some(MinMaxResult::NoElements) => unreachable!(),
+            })
+        })
```
phimuemue, line 480: I think all this should behave essentially like `Itertools::minmax`. Should we extract the core from `Itertools::minmax` and call it from here to ensure that it does the same?

SkiFire13: It could be done, but I don't think we can gain much. If anything a refactor could probably complicate both implementations.

The problem is that here I can only work with one element at a time, while `minmax::minmax_impl` works with an iterator. This implicate that:
- I don't have to treat at all the case where there are no elements (`minmax::minmax_impl` does that once at the start);
- I always have to check for the case where the accumulator has only 1 element (`minmax::minmax_impl` does that once at the start);
- I can't treat the next 2 elements at the same time (half of `minmax::minmax_impl`'s code treat that case);
- I can't cache the current maximum and minimum keys and so I don't care (`minmax::minmax_impl` can and does care);

I also directly use `MinMaxResult` because it has to be a valid value for the final result while `minmax::minmax_impl` works with `min` and `max` variables because it is sure it's in the case of `MinMaxResult::MinMax` and doesn't want to match at every iteration.

So yes, they behave the same but how they're pretty different on how they reach the result. 

If you have any idea on how you would do this I'm welcome to suggestions.

src/grouping_map.rs
```diff
@@ -0,0 +1,562 @@
...
+    pub fn aggregate<FO, R>(self, mut operation: FO) -> HashMap<K, R>
+        where FO: FnMut(Option<R>, &K, V) -> Option<R>,
+    {
+        let mut destination_map = HashMap::new();
+
+        for (key, val) in self.iter {
+            let acc = destination_map.remove(&key);
+            if let Some(op_res) = operation(acc, &key, val) {
+                destination_map.insert(key, op_res);
+            }
```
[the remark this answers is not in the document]
phimuemue: Using `FnMut(Option<R>, ...)` forces the caller to deal with `Option`, i.e. the caller has to deal with `Some`/`None`, essentially checking the same thing that is already checked in `aggregate` in terms of existence in `destination_map`. It may be ok, but we could avoid these checks in the callers by having two functions (one for the `None`-case, one for the `Some`-case).

src/grouping_map.rs
phimuemue: Fair enough.

src/grouping_map.rs
```diff
@@ -0,0 +1,562 @@
...
+
+impl<K, V, I, F> Iterator for MapForGrouping<I, F>
+    where I: Iterator<Item = V>,
+          K: Hash + Eq,
+          F: FnMut(&V) -> K,
+{
+    type Item = (K, V);
+    fn next(&mut self) -> Option<Self::Item> {
+        let val = self.0.next()?;
+        Some(((self.1)(&val), val))
```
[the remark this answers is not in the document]
SkiFire13: Yes, I didn't realise that could be written with a `map`.

phimuemue:
>     * `minmax` return type is the already existing `MinMaxResult`, but the `NoElements` variant is never returned. I didn't want to duplicate that struct thinking it could cause confusion (and if I did what name could I have chosen?);

I think you came to a sensible decision, even if it means that users have to consider the impossible `NoElements` if they match on the result.

I think that *ideally* `minmax` would return `Option<MinMaxResult>`, and `MinMaxResult` would not have the `NoElements` variant. This would clearly be a breaking change; I am not sure what the best option is here. @jswrenn @bluss Do we have precedence for breaking changes? Could we deprecate `MinMaxResult` in favor of a new `MinMax` (which has no empty-case)?

>     * no integration tests for the `_by` and `_by_key` versions of `min`, `max` and `minmax`. 

In general, I think we should quickcheck-test that the new functions are equivalent to `into_group_map` with the equivalents done on the vectors manually.

>     * I didn't want to touch the rest of the library, but I guess we could implement `into_group_map` in terms of `into_grouping_map`?

If we decide to accept this, I would second this.

src/grouping_map.rs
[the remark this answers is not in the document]
SkiFire13: > Do we really need to `remove` and `insert` back into `destination_map`? Couldn't we facilitate the `entry`-API?

`remove` and `insert` are needed if we want to pass an owned accumulator to the closure, like the normal `Iterator::fold`. The alternative is passing a mutable reference but this is more restricting for the end user.
Maybe I should add a second `aggregate` function that provides mutable access instead? This would allow a more efficient implementation in some cases, like in `GroupingMap::collect` (currently it duplicates `aggregate`'s code).

> Using `FnMut(Option<R>, ...)` forces the caller to deal with `Option`, i.e. the caller has to deal with `Some`/`None`, essentially checking the same thing that is already checked in `aggregate` in terms of existence in `destination_map`. It may be ok, but we could avoid these checks in the callers by having two functions (one for the `None`-case, one for the `Some`-case).

Wouldn't this just move the check from the closure to the `aggregate` function?
Performance wise it should remain the same.
But what if the caller wants to mutate an external variable in both the case the map contains the value or not? It can't do that with 2 closures.

SkiFire13:
> I think that _ideally_ `minmax` would return `Option<MinMaxResult>`, and `MinMaxResult` would not have the `NoElements` variant. This would clearly be a breaking change; I am not sure what the best option is here. @jswrenn @bluss Do we have precedence for breaking changes? Could we deprecate `MinMaxResult` in favor of a new `MinMax` (which has no empty-case)?

Another idea I had was to add a new `MinMax` and provide methods to convert `MinMaxResult<T>` into `Option<MinMax<T>>` and `MinMax<T>` into `MinMaxResult<T>`. This would remove the need for the extra match but would introduce a bit of duplication. Bonus point: `MinMax` wouldn't need an `into_option`, it can be directly converted to a tuple.

> In general, I think we should quickcheck-test that the new functions are equivalent to `into_group_map` with the equivalents done on the vectors manually.

Nice idea, I'm definitely gonna add this.

SkiFire13:
I was wondering in the issues section and notice #358 which made me realize `minmax` requires just `PartialOrd`. I want to point out that all the implementations of `min`, `max` and `minmax` for `GroupingMap` in this PR require `Ord`, just like the stdlib does for `Iterator::{min, max}`. This is inconsistent with the normal `minmax` but consistent within itself, and won't need a breaking change if normal `minmax` will ever be fixed.

phimuemue:
https://github.com/rust-itertools/itertools/issues/467 brought up this: Is the `key==value` situation (instead of having explicit key-value-pairs) prevalent enough to either make it the default, or at least support it in a convenient manner?

(As of now, we either have to specify a `(K,V)`-iterator or provide a function to compute the key.)

jswrenn:
Sorry I haven't yet had time to thoroughly review all this.  :(

---

> Do we have precedence for breaking changes? Could we deprecate MinMaxResult in favor of a new MinMax (which has no empty-case)?

@phimuemue Itertools has deprecated/changed/removed plenty of things in [the past](https://github.com/rust-itertools/itertools/blob/master/CHANGELOG.md). I'm fine with making breaking changes, especially if also we take the opportunity to finally remove some long-deprecated methods that conflict with `Iterator`. My impression is that itertools types are seldom in the public API of crates, so version bumps tend to rarely cause issues for end-users.

---

> Adds two functions on the Itertools trait, into_grouping_map and into_grouping_map_by.
> 
> into_grouping_map expects an iter of (K, V) where the K will be used as key and V as value.
> into_grouping_map_by expect only an iter of V as values, the keys will be calculated using the provided functions.
> 
> Both of them return a GroupingMap, which is just a wrapper on an iterator. Since it introduces a lot of related methods I thought it would be better to separate them from the Itertools trait. This also prevents duplicating every method for the _by version.

Possible alternative: defining a `GroupingMap` trait (or `PairItertools` trait?) for these methods. I don't yet have an opinion either way, but I'd like to make sure we consider this alternative.

sollyucko:
Maybe change the `GroupingMap` methods to be more clear about the fact that they operate on the values for each key separately (perhaps `..._each`)?

SkiFire13:
> #467 brought up this: Is the `key==value` situation (instead of having explicit key-value-pairs) prevalent enough to either make it the default, or at least support it in a convenient manner?
> 
> (As of now, we either have to specify a `(K,V)`-iterator or provide a function to compute the key.)

I don't think this is really the case. It's more like the iterator items are the keys and the values are ingnored (they could be `()`). The best solution to that issue using this API would be `.map(|v| (v,())).into_grouping_map().count()`. It's a bit ugly, mainly because the `map` is required for non-`Copy` items. Maybe I could add a shortcut for that `.map(|v| (v,())).into_grouping_map()`? I'm not sure, it feels weird. I feel like a separate function just for count would be better. In fact I think this API shines better with general folding operations. `count` (and `collect`)  is a more specialized operation that can have its own optimizations if it was a separate function.

Little side note: the real case `key == value` is much more complex because of ownership. It would probably require tradeoffs in the API or an entirely separate API which I don't think would be worth it.

> Possible alternative: defining a GroupingMap trait (or PairItertools trait?) for these methods. I don't yet have an opinion either way, but I'd like to make sure we consider this alternative.

Coupled with @sollyucko proposals it would be possible, but I see 2 problems:

- we're back to using `map` as replacement for `into_grouping_map_by`.
- even with more clearer method names it would feel confusing. Having a function like `into_grouping_map_by` that tells me we're entering a separate context is more explicit and less confusing. (At least to me, I'm probably biased)

Ps: I'm currently on vacation without my pc so I can't write any code. I'll think about all of this better when I'll get back at home.

src/grouping_map.rs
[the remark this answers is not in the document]
SkiFire13: I was thinking about this, in particular since issue #467. I made a couple of basic benchmarks and the diffeference in speed is not negligible. For example on a simple `collect` of 1_000_000 `i32` grouped by their reminder by 10 or 10_000, the entry version performed 2x better! But at what cost? 

- I had to add another closure for the init.
- The `init` closure can't take a reference to the `key` as parameter because `Entry::or_insert_with_key` is unstable. Not sure but there may be a way to do it anyway.
- I had to remove the `key` parameter from the other closure as well
- I had to pass a mutable reference to the accumulator instead of an owned value because the `Entry` api gives me only that.

The problem is that the current approach move a lot the accumulator and makes two lookup in the map, one for `remove` and one for `insert`. Moving the accumulator is required if we don't want to restrict the API, so I wondered if I could use the `OccupiedEntry` and `VacantEntry` API to make only one lookup, but unfortunately currently it's impossible. If `OccupiedEntry` had a way to remove the entry, returning the value, but also returning the just freed `VacantEntry` it would be perfect, but there's no such API. I tried to imitate it using `unsafe` and `MaybeUninit` and this is the result: 

https://gist.github.com/SkiFire13/12a61302e35ffe52c0337e0959fb15f6

On the previous example it runs 1.6x faster than the current approach but 1.3 slower than the normal `entry`. On another example (count the number of occurences of a bunch of `i32`s) it performed 2.5x the current approach and the same as the normal entry. This could actually be used to unify the implementation with #468, however I really don't like that I had to use `unsafe` for this. There's also the problem that `MaybeUninit` wasn't stable in rust 1.32.0

Or maybe should I just add an `aggregate_mut`/`aggregate_update` (has anyone better names?) that provides a restricted but faster API using `entry`?

jswrenn:
@SkiFire13 sorry for the delay on this. Do you feel like it's ready to merge? If so, I'll merge it!

SkiFire13:
There's still https://github.com/rust-itertools/itertools/pull/465#discussion_r472137646 which is a performance issue. I haven't decided yet whether:

- Keep it as it is. Pros: simple, safe. Cons: slower than it could be

- Add an `aggregate` variant which takes a `Option<&mut R>` instead of `Option<R>`. Pros: fast, safe. Cons: complex user interface, more limited than it could be

- Use `unsafe` ([example](https://gist.github.com/SkiFire13/12a61302e35ffe52c0337e0959fb15f6)). Pros: simple interface, fast. Cons: potentially unsound (someone should definitely double-check if I messed up somewhere.

Note that if we go for the 1st or 3rd options we could replace them with [`OccupiedEntry::replace_entry_with`](https://docs.rs/hashbrown/0.9.1/hashbrown/hash_map/struct.OccupiedEntry.html#method.replace_entry_with) from hashbrown when it will land on stable (currently it hasn't been proposed yet, not even in nightly, so it will be a long time until it's available in the MSVR)

So the final question here is: do you want to discard performance, complexity in user interface or no-`unsafe`? I feel like this is a pretty big question that needs to be discussed together before merging this PR.

jswrenn:
Of those three options, I'm most inclined to keep things as-is and merge. Its performance issues can be eventually be resolved with `replace_entry_with `.

A fourth option: add an `R: Default` bound and create a temporary value to swap into the entry; e.g.:
```rust
    pub fn aggregate<FO, R>(self, mut operation: FO) -> HashMap<K, R>
    where
        R: Default,
        FO: FnMut(Option<R>, &K, V) -> Option<R>,
    {
        let mut destination_map = HashMap::new();

        for (key, val) in self.iter {
            let entry = destination_map.entry(key);
            match entry {
                Entry::Occupied(mut entry) => {
                    let acc = entry.insert(R::default());
                    if let Some(op_res) = operation(Some(acc), entry.key(), val) {
                        entry.insert(op_res);
                    }
                },
                Entry::Vacant(entry) => {
                    if let Some(op_res) = operation(None, entry.key(), val) {
                        entry.insert(op_res);
                    }
                },
            }
        }

        destination_map
    }
```

SkiFire13:
I feel like adding a `R: Default` bound would be too limiting. For example you wouldn't be able to use `fold_first` if the `Item` is a reference.

jswrenn:
Great, let's merge this as-is, then.

bors r+

Issue #467, opened 2020-08-10
New function: Count number of occurences of each item in an iterator


JarredAllen:
It'd be nice if there could be a function in the `Itertools` trait which counts the number of times each element appears in the iterator and returns a `HashMap` of that information (maybe named `count_each`? I'm not too sure on the name).

I recently had to do this in code that I was writing and was surprised that I couldn't find this functionality in here. I could add this myself pretty easily, if this is wanted (not sure how that process works as I'm new to this project), as I've just written it in my own code.

phimuemue:
If we decide to merge https://github.com/rust-itertools/itertools/pull/465, this would be solved by `into_grouping_map(...).count()`.

JarredAllen:
I fail to see how `into_grouping_map` is applicable here. I want something like the following behavior:
```rust
let values = [0, 0, 0, 1, 1, 2];
let counts = values.into_iter().count_each();
assert_eq!(3, counts[0]);
assert_eq!(2, counts[1]);
assert_eq!(1, counts[2]);
```
The `into_grouping_map` function can't be called here, because the items of the iterator aren't `(K, V)` tuples.

phimuemue:
You may have to use `into_grouping_map_by` (or something similar)... In any case, I think we should not both introduce a `count_each` *and* have the `into_grouping_map` infrastructure - but instead rely on and re-use one of the alternatives.

JarredAllen:
I'd say that your inability to figure out how to use the `GroupingMap` to implement this functionality is evidence that a method should exist to just do this in the `Itertools` trait. If you, a contributor to this project, can't figure out how to correctly implement the solution (or even which method to call), then I fail to see how regular users of the API could be expected to figure it out for themselves.

It would be easy enough to write this `count_each` function to be a thin wrapper around the functionality implemented in the `GroupingMap`, so that it reuses the other one and there's no code duplication.

phimuemue:
That's an interesting take. Actually, when I saw the first proposal of `into_grouping_map`, I suggested to flatten the method call and go with something closer to your idea.

So, just to make that clear: I do not advocate one or the other alternative, but I really believe that having both variants in there does *not really* offer an advantage.

Still, we should possibly investigate whether we can make `into_grouping_map` more convenient to use for the (possibly prevalent) simple case where the iterator element is used as the key.

[...]

PR #468, opened 2020-08-12
Create counts method and tests for it

JarredAllen:
Creates the function described in #467 (I went with the name `counts`).

It currently uses `for_each` and makes a `HashMap` manually because the PR for the `into_grouping_map` function is still being worked on, but this function could easily be changed to call it once that PR is merged, if desired.

### DIFF
--- a/tests/test_std.rs
+++ b/tests/test_std.rs
@@ -913,16 +913,3 @@ fn tree_fold1() {
         assert_eq!(actual, expected);
     }
 }
-
-#[test]
-fn counts() {
-    let a: [usize; 0] = [];
-    assert_eq!(0, a.iter().counts().len());
-    let b = [1, 1, 1, 2, 2, 3];
-    let b_counts = b.iter().counts();
-    assert_eq!(3, b_counts.len());
-    assert_eq!(Some(&3), b_counts.get(&1));
-    assert_eq!(Some(&2), b_counts.get(&2));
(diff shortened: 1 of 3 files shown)


tests/test_std.rs
```diff
@@ -913,3 +913,16 @@ fn tree_fold1() {
...
+    let a: [usize; 0] = [];
+    assert_eq!(0, a.iter().counts().len());
+    let b = [1, 1, 1, 2, 2, 3];
+    let b_counts = b.iter().counts();
+    assert_eq!(3, b_counts.len());
+    assert_eq!(Some(&3), b_counts.get(&1));
+    assert_eq!(Some(&2), b_counts.get(&2));
+    assert_eq!(Some(&1), b_counts.get(&3));
+    assert_eq!(None, b_counts.get(&4));
+}
```
SkiFire13, line 928: Wouldn't a quicktest be better? You can probably adapt the test I wrote in my PR for my `count` method:

https://github.com/rust-itertools/itertools/pull/465/files#diff-934f6ed12102bba0b289ee2855cb9bd1R1306-R1324

JarredAllen: Yes, it would, and I've now included that. Thanks for the suggestion.

jswrenn:
Looks good to me! Thanks for this!

bors r+

PR #471, opened 2020-08-21
more no_std support

zendurix:
I have implemented most functions to be able to use in `no_std` with `alloc` (by using optional feature `use_alloc`). Also I have included optional crate `hashbrown` (it is used while using `use_alloc`) for no_std `Hashmap`.
 Everything builds and tests are passing in both default and no default with feature `use_alloc`.
There are many changes due to cargo fmt, if this is a problem I can return them to previous formatting, however this will require some manual work.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ test = false
 
 [dependencies]
 either = { version = "1.0", default-features = false }
+hashbrown = { version = "0.8", optional = true}
 
 [dev-dependencies]
 rand = "0.7"
@@ -38,6 +39,7 @@ version = "0.2"
 
 [features]
 default = ["use_std"]
+use_alloc = ["hashbrown"]
 use_std = []
 
 [profile]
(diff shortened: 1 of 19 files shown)


zendurix:
I have fix the cargo fmt problem, only relevant changes are now left.
Please do cargo fmt after merge to avoid this next time :p

It passed no std check
https://github.com/mystor/cargo-no-std-check

Cargo.toml
```diff
@@ -24,6 +24,7 @@ test = false
 
 [dependencies]
 either = { version = "1.0", default-features = false }
+hashbrown = { version = "0.8", optional = true}
```
jswrenn, line 27: Hashbrown doesn't seem to have a minimum-supported-Rust-version policy—Itertools does. We therefore can't just drop-in hashbrown as an alternative to `std::collections::HashMap`.

jswrenn:
I've opened #474, which implements a minimal `use_alloc` feature.

I'm happy to revisit porting `HashMap` support to `use_alloc` in the future, but I'd like to take a carefully considered, incremental approach to doing so.

PR #473, opened 2020-09-01
Add helper method for taking the k smallest elements in an iterator

nbraud:


nbraud:
Many thanks to @Selicre and whoever else in @rustfurs provided feedback.

nbraud:
I started adding another test, but I'll keep figuring out how to generate arbitrarish iterators for another day.

src/lib.rs
```diff
@@ -185,6 +185,8 @@ mod group_map;
 mod groupbylazy;
 mod intersperse;
 #[cfg(feature = "use_std")]
+mod k_smollest;
```
eddyb, line 188: Hmm, I wonder if `itertools` will grow an `alloc` feature, as that's all a lot of these need (AFAIK).

src/lib.rs
```diff
@@ -2268,6 +2270,40 @@ pub trait Itertools : Iterator {
...
+    /// itertools::assert_equal(five_smallest, 0..5);
+    /// ```
+    #[cfg(feature = "use_std")]
+    fn k_smallest(self, k: usize) -> VecIntoIter<Self::Item>
+        where Self: Sized,
+              Self::Item: Ord
+    {
+        crate::k_smollest::k_smollest(self, k)
+            .into_sorted_vec()
+            .into_iter()
```
eddyb, line 2304: Would be cool to use `into_iter_sorted`, too bad that's unstable.

src/lib.rs
```diff
@@ -2268,6 +2270,40 @@ pub trait Itertools : Iterator {
...
+    /// let numbers = vec![6, 9, 1, 14, 0, 4, 8, 7, 11, 2, 10, 3, 13, 12, 5];
+    ///
+    /// let five_smallest = numbers
+    ///     .into_iter()
+    ///     .k_smallest(5);
+    ///
+    /// itertools::assert_equal(five_smallest, 0..5);
+    /// ```
+    #[cfg(feature = "use_std")]
+    fn k_smallest(self, k: usize) -> VecIntoIter<Self::Item>
```
eddyb, line 2298: Not sure what's `itertools` policy but maybe if the return type was newtype'd as `KSmallestIter` or something, it would allow changing it into the future to something more efficient?

src/lib.rs
nbraud: I agree that an `alloc` feature would be nice, as it looks like (almost?) all of those only need that (rather than all of `std`)

src/lib.rs
nbraud: That's a good point, I'll address that tomorrow  <3

src/k_smollest.rs
```diff
@@ -0,0 +1,17 @@
...
+pub(crate) fn k_smollest<T: Ord, I: Iterator<Item = T>>(mut iter: I, k: usize) -> BinaryHeap<T> {
+    if k == 0 { return BinaryHeap::new(); }
+
+    let mut heap = iter.by_ref().take(k).collect::<BinaryHeap<_>>();
+
+    for i in iter {
+        debug_assert_eq!(heap.len(), k);
+        // Guaranteed not-None, since we keep exactly k>0 elements in the heap.
+        let mut lorgest = heap.peek_mut().unwrap();
+        if *lorgest > i { *lorgest = i; }
```
eddyb, line 13: This might need some comments, I had no idea `peek_mut` did the equivalent of `remove`+`insert`, really cool API!

tests/test_std.rs
```diff
@@ -353,6 +356,37 @@ fn sorted_by() {
     it::assert_equal(v, vec![4, 3, 2, 1, 0]);
 }
 
+quickcheck! {
+    fn k_smallest_range(n: u64, m: u64, k: u64) -> () {
+        // Generate a random permutation of n..n+m
+        let i = {
+            let mut v: Vec<u64> = (n..n.saturating_add(m)).collect();
```
nbraud, line 363: I should probably be putting a reasonable bound on `m`, as we do allocate `8 * (m + min(m, k))` bytes in this test...

src/lib.rs
scottmcm: Well that sounds like something I can push for: https://github.com/rust-lang/rust/pull/76234

scottmcm:
I was thinking about this some more, and since the logic in here is so cool, I was wondering if it would make sense to expose it in more ways.  Notably, it seems like if there were a type for the logic, it could implement `Extend` so that once you have the smallest k of some set, you can `.extend(more_elements)` and it'd be the combined smallest k of both sets.

Dunno if that's something that'd be appropriate for `itertools`, though...

src/lib.rs
```diff
@@ -2268,6 +2270,43 @@ pub trait Itertools : Iterator {
...
     }
 
+    /// Sort the k smallest elements into a new iterator, in ascending order.
+    ///
+    /// **Note:** This consumes the entire iterator, and returns the result
+    /// as a new iterator that owns its elements.  If the input contains
+    /// less than k elements, the result is equivalent to `self.sorted()`.
+    ///
+    /// This is guaranteed to use `k * sizeof(Self::Item) + O(1)` memory
+    /// and `O(n log k)` time, with `n` the number of elements in the input.
```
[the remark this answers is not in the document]
nbraud: @scottmcm `O(n)` time is impossible in the general case.  Otherwise, `it.k_smallest(n)` would sort a sequence of length `n` in linear time.

Re: `BinaryHeap::from_iter(it).into_sorted_iter().take(k)`, wouldn't this have strictly-worse performance characteristics, regarding both time and space?  I need to think some more about it once caffeinated.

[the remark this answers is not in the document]
scottmcm: It's [absolutely possible](https://en.wikipedia.org/wiki/Selection_algorithm#Unordered_partial_sorting) if this only needs to find the k smallest elements and not also sort those elements -- the fact that this is _also_ sorting those elements wasn't obvious from the title.  (And I would say, the extra `k log k` of sorting them is unfortunate if someone just wants the smallest k and _doesn't_ need them sorted.  It would be nice to be able to turn them back into a `vec::IntoIter` for the people who would be fine with heap order, like `.k_smallest(10).sum()`.)

So there are plenty of options here:

Algorithm | space | time
-----------|-----|-----------
nth element (unsorted) | n | n
partial sort | n | n + k log k
full heap | n | n + k log n
k-size heap | k | n log k

(Which does emphasize that `BinaryHeap::into_iter_sorted` should only be used for not-known-ahead-of-time `k`.)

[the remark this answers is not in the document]
nbraud: > It's absolutely possible if this only needs to find the k smallest elements and not also sort those

Agreed; that just seemed like a worse API to be exposing, as:
- an undefined output order, which happens to be often sorted or almost-sorted (like heap order), can be confusing for the user (I myself thought for a moment that `BinaryHeap::into_vec()` sorted the result, and was surprised by that, as the examples I was testing with happened to produce a sorted result) ;
- sorting the end result doesn't change the asymptotic complexity, for any heap-based implementation.

However, if we expose something like a fixed-size heap — and I agree it's a good idea, if only for the `.extend()` type of use-cases you suggested — users who do not require a sorted result could use that directly.

> So there are plenty of options here:

You seem to be including partition-based algorithms — AFAIK, the only way to get better time complexity than O(n log k) — and those [do not work directly for iterators]; it's always possible to first collect everything into a `Vec`, use a partition-based algorithm, then truncate the `Vec` down to size `k`, but I would expect this to be much slower in practice.

[do not work directly for iterators]: https://en.wikipedia.org/wiki/Selection_algorithm#Online_selection_algorithm

nbraud:
@scottmcm Do you think I should move making a fixed-size heap wrapper to a separate PR, and make this one depend on it?

scottmcm:
@nbraud I don't know -- there's not clear precedent of exposing non-iterator types of itertools.  So maybe it's best to just leave it as you have it until an itertools maintainer comment.

src/k_smollest.rs
```diff
@@ -0,0 +1,18 @@
+use std::collections::BinaryHeap;
+use std::cmp::Ord;
+
+pub(crate) fn k_smollest<T: Ord, I: Iterator<Item = T>>(mut iter: I, k: usize) -> BinaryHeap<T> {
```
jswrenn, line 4: `smollest`

heh. I like it.

src/lib.rs
nbraud: I forgot to mention, but that doesn't seem feasible while keeping the guarantee that `.collect::<Vec<_>>` is “free” (in-place, O(1) time).  The stdlib's vec iterator can do it, but does so using specialisation (which isn't stabilised yet)

src/k_smollest.rs
```diff
@@ -0,0 +1,18 @@
...
+    if k == 0 { return BinaryHeap::new(); }
+
+    let mut heap = iter.by_ref().take(k).collect::<BinaryHeap<_>>();
+
+    for i in iter {
+        debug_assert_eq!(heap.len(), k);
+        // Guaranteed not-None, since we keep exactly k>0 elements in the heap.
+        let mut lorgest = heap.peek_mut().unwrap();
+        // Equivalent to heap.push(min(i, heap.pop())) but more efficient.
+        if *lorgest > i { *lorgest = i; }
```
SkiFire13, line 14: With the current stdlib implementation this is slower (and by a noticeable amount) than what it should be. See https://github.com/rust-lang/rust/pull/75974 for details. For now it would be more efficient to do the following:

```suggestion
        // Equivalent to heap.push(min(i, heap.pop())) but more efficient.
        if *heap.peek().unwrap() > i { *heap.peek_mut().unwrap() = i; }
```

SkiFire13 (changes requested):

jswrenn (changes requested):

src/k_smollest.rs
phimuemue: I do not want to be a spoilsport, but I am wary with "`smollest`", as it e.g. disallows to `grep` for `smallest`, making it harder to find. Or is it a common term (e.g. know from other (programming) languages) familiar to most programmers?

jswrenn: Hah, no this is definitely just cute and not a term of art. But alas, @phimuemue has a point.

@nbraud, I'm *so* sorry, but could you name this `k_smallest`?

src/k_smollest.rs
nbraud: Thanks for the catch; rust-lang/rust#75974 was merged, but it still seems worth working around the suboptimality in older Rust versions.

src/k_smollest.rs
nbraud: Done :)

src/lib.rs
nbraud: Done :)

nbraud:
Ping?  I addressed all feedback a week ago (but I'm not sure Github notifies on new pushes)

jswrenn (approved):
Thanks for this! Sorry for the delay.

bors r+

PR #475, opened 2020-09-04
Trait impls for tuples of arity <= 12

### LINKED ISSUE #398: add implementation of 5-tuple for tuple_combinations?
I was using tuple_combinations for iterating over various card combinations and ran into the 4-tuple limitation when trying to iterate over all possible 5-card combinations of a deck of 52 cards.  Any chance you'd add a five-tuple implementation or should I use another method?

### LINKED ISSUE #428: HomogeneousTuple / TupleCollect for N > 4 ?
It would be useful to have HomogeneousTuple / TupleCollect for N larger than 4.
E.g. I often use it with `splitn` to avoid allocations, like:
```rust
s.splitn(N, pattern).collect_tuple()
```
instead of 
```rust
if let [/* ... */] = pattern.splitn(N, pattern).collect_vec().as_slice() {
	Some(/* ... */)
} else {
	None
}
```
And often my N is higher than 4.

jswrenn:
I stopped at 12, because that's where libcore stops.

fixes #428
fixes #398

### DIFF
--- a/src/tuple_impl.rs
+++ b/src/tuple_impl.rs
@@ -280,23 +280,14 @@ macro_rules! impl_tuple_collect {
                 return None;
             }
 
-            #[allow(unused_assignments)]
             fn collect_from_iter_no_buf<I>(iter: I) -> Option<Self>
                 where I: IntoIterator<Item = $A>
             {
                 let mut iter = iter.into_iter();
-                loop {
-                    $(
-                        let $Y = if let Some($Y) = iter.next() {
-                            $Y
-                        } else {
-                            break;
-                        };
-                    )*
-                    return Some(($($Y),*,))
-                }
 
-                return None;
+                Some(($(
+                    { let $Y = iter.next()?; $Y },
+                )*))
             }
 
             fn num_items() -> usize {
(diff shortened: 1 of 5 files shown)


src/tuple_impl.rs
```diff
@@ -280,23 +280,14 @@ macro_rules! impl_tuple_collect {
...
-                            break;
-                        };
-                    )*
-                    return Some(($($Y),*,))
-                }
 
-                return None;
+                Some(($(
+                    { let $Y = iter.next()?; $Y },
+                )*))
```
jswrenn, line 290: I *think* this modification is semantically equivalent, but I'd love if someone else could check it over before I merge.

jswrenn:
Seems like a random CI failure?

bors try

phimuemue:
bors try

src/tuple_impl.rs
phimuemue: Looks good to me. Godbolt even seems to generate very similar code in `--release`.

off-topic: I do not have much time to spare at the moment, so I won't be very responsive.

PR #481, opened 2020-09-21
Add fold_mut alternative to fold

mlodato517:
**This PR**
Adds a `fold_mut` alternative to `fold`

**Why?**
1. Sometimes the closure in a `fold` requires an awkward final line to return the accumulator as mentioned [here](https://github.com/rust-lang/rust-clippy/issues/6053).
2. Sometimes `fold_mut` can be faster. And sometimes it can be slower! See the benchmarks [here](https://github.com/rust-lang/rust/pull/76746#discussion_r490572461). TL;DR, if the accumulator is at least as big as a `&mut` to the accumulator, `fold` can be slower than `fold_mut` as moving the larger accumulator is more work than moving the `&mut`.

**Background**
A lot of the background is in [this issue](https://github.com/rust-lang/rust/issues/76725) and [this PR](https://github.com/rust-lang/rust/pull/76746). Based on the discussion there, it seemed that this wasn't warranted for the standard library so I figured I'd pitch it here!

**Benchmarks**
I ran the benchmarks like:
```
cargo bench "fold sum accumulator" \
  && cargo bench "fold vec accumulator" \
  && cargo bench "fold chained iterator with num accumulator" \
  && cargo bench "fold chained iterator with vec accumulator" \
  && open target/criterion/report/index.html
```
(not sure if there's a shorter syntax) and saw:

What | Benchmark
-|-
folding an `i64` |num
folding a `vec` | vec
folding a [`chain`ed iterator](https://github.com/rust-lang/rust/pull/76746#discussion_r490422807) into an `i64` | i64_chain
folding a `chain`ed iterator into a `vec` | vec_chain

I could also add a benchmark with like an `i8` or something to show how much faster `fold` is when folding into something much smaller than a reference if we want that! I just had a hard time coming up with an accumulator that wouldn't overflow but would still be super readable. It might be like `(0i8..10).chain(-10i8..0).cycle().take(1_000_000)` or some such.

### DIFF
--- a/benches/fold_mut.rs
+++ b/benches/fold_mut.rs
@@ -1,0 +1,109 @@
+use criterion::{black_box, criterion_group, criterion_main, Criterion};
+use itertools::Itertools;
+
+fn bench_sum(c: &mut Criterion) {
+    let mut group = c.benchmark_group("fold sum accumulator");
+
+    group.bench_function("fold", |b| {
+        b.iter(|| {
+            (0i64..1_000_000)
+                .map(black_box)
+                .fold(0, |sum, n| sum + n)
+        })
+    });
+
+    group.bench_function("fold_mut", |b| {
+        b.iter(|| {
+            (0i64..1_000_000).map(black_box).fold_mut(0, |sum, n| {
+                *sum += n;
+            })
+        })
+    });
+
+    group.finish();
+}
+
+fn bench_vec(c: &mut Criterion) {
+    let mut group = c.benchmark_group("fold vec accumulator");
+
+    group.bench_function("fold", |b| {
+        b.iter(|| {
+            (0i64..1_000_000)
+                .map(black_box)
+                .fold(Vec::new(), |mut v, n| {
+                    v.push(n);
+                    v
+                })
+        })
+    });
+
+    group.bench_function("fold_mut", |b| {
+        b.iter(|| {
+            (0i64..1_000_000)
+                .map(black_box)
+                .fold_mut(Vec::new(), |v, n| {
+                    v.push(n);
+                })
+        })
+    });
+
+    group.finish();
+}
+
+fn bench_num_with_chain(c: &mut Criterion) {
+    let mut group = c.benchmark_group("fold chained iterator with num accumulator");
+
+    group.bench_function("fold", |b| {
+        b.iter(|| {
+            (0i64..1_000_000)
+                .chain(0i64..1_000_000)
+                .map(black_box)
+                .fold(0, |sum, n| sum + n)
+        })
+    });
+
+    group.bench_function("fold_mut", |b| {
+        b.iter(|| {
+            (0i64..1_000_000)
+                .chain(0i64..1_000_000)
+                .map(black_box)
+                .fold_mut(0, |sum, n| {
+                    *sum += n;
+                })
+        })
+    });
+
+    group.finish();
+}
+
+fn bench_vec_with_chain(c: &mut Criterion) {
+    let mut group = c.benchmark_group("fold chained iterator with vec accumulator");
+
+    group.bench_function("fold", |b| {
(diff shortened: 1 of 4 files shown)


benches/fold_mut.rs
```diff
@@ -0,0 +1,109 @@
...
+    group.finish();
+}
+
+fn bench_num_with_chain(c: &mut Criterion) {
+    let mut group = c.benchmark_group("fold chained iterator with num accumulator");
+
+    group.bench_function("fold", |b| {
+        b.iter(|| {
+            (0i64..1_000_000)
+                .chain(0i64..1_000_000)
```
mlodato517, line 59: Added a benchmark with `chain` to show that, because `fold_mut` uses `for_each`, it benefits from `Chain`'s specialization of [`fold`](https://doc.rust-lang.org/src/core/iter/adapters/chain.rs.html#107)

src/lib.rs
```diff
@@ -2831,6 +2831,47 @@ pub trait Itertools : Iterator {
...
+    /// ```
+    ///
+    /// `fold_mut()` may also be more performant in situations where the
+    /// accumulator is "large" as passing it by `&mut` can be cheaper than
+    /// transferring ownership of the accumulator.
+    ///
+    /// # Examples
+    /// ```
+    /// # use itertools::Itertools;
+    /// let evens = [1, 2, 3, 4, 5, 6].iter().fold_mut(Vec::new(), |evens, &num| {
```
mlodato517, line 2855: Is this example too lame? I was going to use the example of counting but ... there's `.counts` right above this so that seemed worse haha

phimuemue:
> Sometimes the closure in a `fold` requires an awkward final line to return the accumulator as mentioned [here](https://github.com/rust-lang/rust-clippy/issues/6053).

This bugged me in some of my projects, too. However, I realized that I then needed e.g. `fold1_mut` and others, as well.

Thus, I solved this in a different way: I introduced `mutate_return` that "lifts" a `Fn(&mut S, A)` to `Fn(S, A)->S` (see [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=8de33e5a442b486e7524c4986e825fbc)) and makes this independent of `fold`.

> Sometimes fold_mut can be faster.

In my implementation I assumed that the compiler is clever enough to optimize it as good as possible in any case - which, reading your PR, seems wrong.

So, my questions:

1. Should we also offer `fold1_mut` and others? (How far should we go, then?)
2. In an ideal world, I think the compiler should take care of the optimizations necessary, so that using `fold` vs. `fold_mut` should result in identical, most efficient code. Is it foreseeable that the compiler becomes that clever? (If so, I'd be reluctant to introduce `fold_mut`.)

mlodato517:
> Should we also offer fold1_mut and others? (How far should we go, then?)

Great question! I have no idea 🤔 I imagine it'd be great to have this variant of all the methods since you [can implement `fold_mut` in terms of `fold`](https://github.com/rust-lang/rust/pull/76746#discussion_r488818732) (although, of course, you get the performance hit back)

> Is it foreseeable that the compiler becomes that clever?

Maybe! There's a little discussion going on [here](https://rust-lang.zulipchat.com/#narrow/stream/122651-general/topic/Update-in-place.20vs.20mut.3F/near/210199371). I do think that, if we could get `fold` to be as fast as `fold_mut`, then that'd be great, but I still wonder if the ergonomics of `fold_mut` might be desirable in some cases (although, maybe we wait for the [clippy lint](https://github.com/rust-lang/rust-clippy/issues/6053) to be discussed?).

mlodato517:
Just checking back in for some quick guidance here - should I be working to implement `fold_mut` for the other `fold` variants or should we close this?

mlodato517:
I think this is sufficiently stale to close :-)

PR #487, opened 2020-10-10
Simplify CombinationsWithReplacement

phimuemue:
* Eliminate `k`: It can be obtained via `indices.len`.
* Eliminate `max_index`: It can be obtained via `pool.len()-1`.


jswrenn (approved):
bors r+

PR #489, opened 2020-10-16
Adding the partition_results method

bachrc:
Here, we add a method partitionning a list of Results. It is useful while chaining risky operations. 

Though, I would have a question. Why can we only use this method with into_iter ? Why can't we use this with only references ? Thanks in advance.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2322,6 +2322,35 @@ pub trait Itertools : Iterator {
         (left, right)
     }
 
+    /// Collects elements from an Iterator of Results, and then returns
+    /// two vecs : one with Oks, one with Errors
+    ///
+    /// ```
+    /// use itertools::{Itertools, Either};
+    ///
+    /// let successes_and_failures = vec![Ok(1), Err(false), Err(true), Ok(2)];
+    ///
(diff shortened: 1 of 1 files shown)


phimuemue:
Hi there! Isn't this essentially `partition_map` (just with `Result` instead of `Either`, and `Vec` instead of the more concrete result types `A` and `B`)? If so, I'd prefer to close this, as it just would duplicate the logic of `partition_map`. (If you *really* need to use it with `Result` instead of `Either`, there should possibly be an easy way to convert one into the other.)

As for the `into_iter` (I hope I get your question right): [It should also work with `iter`](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=3a81b232317aeb613214621e3289bbdf), but not without it, because you must somehow convert get an iterator from the `Vec`: You can only call the function on an *iterator*, not on the `Vec` itself.

phimuemue:
Please tell if closing this was inadequate.

bachrc:
I am so sorry for the late answer, got busy and forgot

I completely agree that this is similar to partition_map. Though, I am confident that abstracting the result partitioning makes the chained business operations more readable. When you iterate over risky methods (that returns a Result), you could need to know if operations went wrong!

PR #490, opened 2020-10-26
Add `unique_lru` and `unique_by_lru`

futpib:
These are just like `unique` and `unique_by` but use [`LruCache`](https://github.com/jeromefroe/lru-rs) instead of `HashMap`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ test = false
 
 [dependencies]
 either = { version = "1.0", default-features = false }
+lru = { version = "0.6.0", optional = true }
 
 [dev-dependencies]
 rand = "0.7"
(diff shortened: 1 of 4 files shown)


phimuemue:
Hi there, thanks for the idea! I definitely see value in this, but: Should we really tie this to `LruCache`?

Couldn't this share the implementation with `UniqueBy`? (In the past, we already had the problem of diverging implementations if two closely-related things were implemented independently.)

And if so: Couldn't we make `UniqueBy` more generic so that it does not only work with `HashMap`? (E.g. by designing a `trait` that is then implemented for `HashMap`, `Vec`, and `LruCache`.) Then, the current `UniqueBy` is a special case implemented in terms of the more generic `UniqueBy`. - Moreover, users could even plug in their own containers if needed.

futpib:
Yes, adding a `unique_by_something` with support for a generic container would be much nicer, unfortunately Rust does not provide (nor has a conventional community-developed) traits for generic containers.

Anyway, after I submitted this PR, I found that I need a similar method for streams, and figured that instead of implementing it there too, I'd rather create an `LruCache` myself and use a `filter_map`, which would do for both streams and iterators.

I'm closing this for now hoping some day we will see general interfaces for both containers and iterators/streams such that `unique_by` could be generalized over container and iterator/stream (or anything that can be `filter_map`ped).

PR #491, opened 2020-10-29
Use `for_each`

phimuemue:
I went through the code and looked for places where we simply traverse generic iterators. For these cases, `for_each` may be more efficient than a plain `for`.


jswrenn (approved):
bors r+

PR #494, opened 2020-11-09
Added unstable variants of sort from std

BudiNverse:
Hello! This PR adds the following methods

- `sorted_unstable`
- `sorted_unstable_by`
- `sorted_unstable_by_key`

The methods mentioned above have the same signature as the normal sorted version. Internally, all of them uses the unstable variants of sort from std.


jswrenn (approved):
Thanks for this!

bors r+

PR #497, opened 2020-11-26
Intra-doc links

### LINKED ISSUE #496: Intra-doc links
These finally landed on stable. Besides looking better in the code, they avoid dead link issues when the docs for a trait (like `Itertools`) is copied and pasted onto every iterator (which will be at a different address and so will break any relative links). Right now the `itertools` crate is the only remaining source of dead links in my project.

digama0:
This commit changes all the documentation to use intra-doc links, which
were stabilized in rust 1.48. This helps to avoid broken links in crates
that depend on this one.

fixes #496

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -5,13 +5,13 @@
 //! Extra iterator adaptors, functions and macros.
 //!
 //! To extend [`Iterator`] with methods in this crate, import
-//! the [`Itertools` trait](./trait.Itertools.html):
+//! the [`Itertools` trait](Itertools):
 //!
 //! ```
 //! use itertools::Itertools;
 //! ```
 //!
-//! Now, new methods like [`interleave`](./trait.Itertools.html#method.interleave)
+//! Now, new methods like [`interleave`](Itertools::interleave)
 //! are available on all iterators:
 //!
 //! ```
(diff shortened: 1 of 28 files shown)


src/lib.rs
```diff
@@ -279,8 +277,6 @@ macro_rules! iproduct {
 /// Prefer this macro `izip!()` over [`multizip`] for the performance benefits
 /// of using the standard library `.zip()`.
 ///
-/// [`multizip`]: fn.multizip.html
-///
```
phimuemue, line 283: Could you explain why this commit removes some lines?

phimuemue (commented):
I am not that much into documentation, but assuming that you transformed the links properly to the new format, I think this is a valuable contribution.

src/lib.rs
digama0: In markdown if you write `[foo]` without defining `[foo]: bar` or writing `[foo](bar)`, the link will go directly to `foo`. That means that if the text of the link happens to already be valid according to intra-doc link rules, there is no need to provide a custom link target (this line would end up reading ``[`multizip`]: multizip`` and rustdoc knows enough to strip the backticks).

jswrenn (approved):
Thanks for this! This *maybe* violates our MSRV policy, but not in a build-breaking way so I'm fine with merging it.

bors r+

jswrenn:
Huh. I'm *really* not sure what to make of this build "failure". Everything appears to have passed.

jswrenn:
bors retry

jswrenn:
bors retry

PR #502, opened 2020-12-03
add .duplicates() and .duplicates_by(..) operations

petrosagg:
Uses a HashMap to detect duplicates in an iterator and emits them only once. The implemention is similar to the `unique()` and `unique_by()` methods but in this case the items are never cloned.

### DIFF
--- a/src/duplicates_impl.rs
+++ b/src/duplicates_impl.rs
@@ -1,0 +1,206 @@
+use std::hash::Hash;
+
+mod private {
+    use std::collections::HashMap;
+    use std::hash::Hash;
+    use std::fmt;
+
+    #[derive(Clone)]
+    #[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
+    pub struct DuplicatesBy<I: Iterator, Key, F> {
+        pub(crate) iter: I,
+        pub(crate) meta: Meta<Key, F>,
+    }
+
+    impl<I, V, F> fmt::Debug for DuplicatesBy<I, V, F>
+    where
+        I: Iterator + fmt::Debug,
+        V: fmt::Debug + Hash + Eq,
+    {
+        debug_fmt_fields!(DuplicatesBy, iter, meta.used);
+    }
+
+    impl<I: Iterator, Key: Eq + Hash, F> DuplicatesBy<I, Key, F> {
+        pub(crate) fn new(iter: I, key_method: F) -> Self {
+            DuplicatesBy {
+                iter,
+                meta: Meta {
+                    used: HashMap::new(),
+                    pending: 0,
+                    key_method,
+                },
+            }
+        }
+    }
+
+    #[derive(Clone)]
+    pub struct Meta<Key, F> {
+        used: HashMap<Key, bool>,
+        pending: usize,
+        key_method: F,
+    }
+
+    impl<Key, F> Meta<Key, F>
+    where
+        Key: Eq + Hash,
+    {
+        /// Takes an item and returns it back to the caller if it's the second time we see it.
+        /// Otherwise the item is consumed and None is returned
+        #[inline(always)]
+        fn filter<I>(&mut self, item: I) -> Option<I>
+        where
+            F: KeyMethod<Key, I>,
+        {
+            let kv = self.key_method.make(item);
+            match self.used.get_mut(kv.key_ref()) {
+                None => {
+                    self.used.insert(kv.key(), false);
+                    self.pending += 1;
+                    None
+                }
+                Some(true) => None,
+                Some(produced) => {
+                    *produced = true;
+                    self.pending -= 1;
+                    Some(kv.value())
+                }
+            }
+        }
+    }
+
+    impl<I, Key, F> Iterator for DuplicatesBy<I, Key, F>
+    where
+        I: Iterator,
+        Key: Eq + Hash,
+        F: KeyMethod<Key, I::Item>,
+    {
+        type Item = I::Item;
+
+        fn next(&mut self) -> Option<Self::Item> {
+            let DuplicatesBy { iter, meta } = self;
+            iter.find_map(|v| meta.filter(v))
+        }
+
+        #[inline]
+        fn size_hint(&self) -> (usize, Option<usize>) {
+            let (_, hi) = self.iter.size_hint();
+            // There are `hi` number of items left in the base iterator. In the best case scenario,
+            // these items are exactly the same as the ones pending (i.e items seen exactly once so
+            // far), plus (hi - pending) / 2 pairs of never seen before items.
+            let hi = hi.map(|hi| {
+                let max_pending = std::cmp::min(self.meta.pending, hi);
+                let max_new = std::cmp::max(hi - self.meta.pending, 0) / 2;
+                max_pending + max_new
+            });
+            // The lower bound is always 0 since we might only get unique items from now on
+            (0, hi)
+        }
+    }
+
+    impl<I, Key, F> DoubleEndedIterator for DuplicatesBy<I, Key, F>
+    where
+        I: DoubleEndedIterator,
+        Key: Eq + Hash,
+        F: KeyMethod<Key, I::Item>,
+    {
+        fn next_back(&mut self) -> Option<Self::Item> {
+            let DuplicatesBy { iter, meta } = self;
+            iter.rev().find_map(|v| meta.filter(v))
+        }
+    }
+
+    /// A keying method for use with `DuplicatesBy`
+    pub trait KeyMethod<K, V> {
+        type Container: KeyXorValue<K, V>;
+
+        fn make(&mut self, value: V) -> Self::Container;
+    }
+
+    /// Apply the identity function to elements before checking them for equality.
+    pub struct ById;
+    impl<V> KeyMethod<V, V> for ById {
+        type Container = JustValue<V>;
+
+        fn make(&mut self, v: V) -> Self::Container {
+            JustValue(v)
+        }
+    }
+
+    /// Apply a user-supplied function to elements before checking them for equality.
+    pub struct ByFn<F>(pub(crate) F);
+    impl<K, V, F> KeyMethod<K, V> for ByFn<F>
+    where
+        F: FnMut(&V) -> K,
+    {
+        type Container = KeyValue<K, V>;
+
+        fn make(&mut self, v: V) -> Self::Container {
+            KeyValue((self.0)(&v), v)
+        }
+    }
+
+    // Implementors of this trait can hold onto a key and a value but only give access to one of them
+    // at a time. This allows the key and the value to be the same value internally
+    pub trait KeyXorValue<K, V> {
+        fn key_ref(&self) -> &K;
+        fn key(self) -> K;
+        fn value(self) -> V;
+    }
+
+    pub struct KeyValue<K, V>(K, V);
+    impl<K, V> KeyXorValue<K, V> for KeyValue<K, V> {
+        fn key_ref(&self) -> &K {
+            &self.0
+        }
+        fn key(self) -> K {
+            self.0
+        }
+        fn value(self) -> V {
+            self.1
+        }
+    }
+
+    pub struct JustValue<V>(V);
+    impl<V> KeyXorValue<V, V> for JustValue<V> {
+        fn key_ref(&self) -> &V {
+            &self.0
+        }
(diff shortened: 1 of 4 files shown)


src/duplicate_impl.rs
```diff
@@ -0,0 +1,160 @@
...
+            }
+        }
+        None
+    }
+
+    #[inline]
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let (low, hi) = self.iter.size_hint();
+        ((low > 0 && self.used.is_empty()) as usize, hi)
+    }
```
phimuemue, line 63: Hi! Is this really correct? Let's assume `iter` is from a `Vec` containing `[1,2,3,4,5,6]` (so there are no duplicates). Then `self.iter.size_hint` is `(6, 6)` and `used` is initially empty, meaning that you'd mistakenly get `(1, 6)` instead of `(0, 6)`, right?

petrosagg: Hi! You're right, I thought the hints would be the same as the Unique iterator. I just pushed the correct logic now and added a quickcheck test

src/lib.rs
```diff
@@ -1138,6 +1142,54 @@ pub trait Itertools : Iterator {
         adaptors::dedup_by_with_count(self, cmp)
     }
 
+    /// Return an iterator adaptor that produces elements that appear more than once during the
+    /// iteration. Duplicates are detected using hash and equality.
+    ///
+    /// The iterator is stable, returning the duplicate items in the order in which they occur in
+    /// the adapted iterator. Each duplicate item is returned exactly once. If an item appears more
+    /// than twice, the second item is the item retained and the rest are discarded.
```
jswrenn, line 1150: Can you clarify the difference between this method and `unique`? Surely, it's more than *just* the `Clone` bound — otherwise we'd just remove the `Clone` bound on `unique`.

Is the difference just that `unique` produces the first encountered duplicate, and `duplicate` produces the last encountered duplicate?

src/lib.rs
```diff
@@ -1138,6 +1142,54 @@ pub trait Itertools : Iterator {
...
+    ///
+    /// ```
+    /// use itertools::Itertools;
+    ///
+    /// let data = vec![10, 20, 30, 20, 40, 10, 50];
+    /// itertools::assert_equal(data.into_iter().duplicate(),
+    ///                         vec![20, 10]);
+    /// ```
+    #[cfg(feature = "use_std")]
+    fn duplicate(self) -> Duplicate<Self>
```
jswrenn, line 1160: I'm a bit iffy on the name. `duplicate`, in the context of a method name, reads as a verb to me, but this isn't a method that *duplicates* items. I'm hoping that pinning down the difference between `unique` and `duplicate` leads to a more fitting name.

src/lib.rs
petrosagg: For sure, the difference is that `unique` produces a value regardless of whether it is a duplicate or not, whereas this method only produces elements that appear at least twice.

i.e `unique` on `[ 1, 2, 3 ]` produces `[1, 2, 3]` whereas `duplicate` produces no values

src/lib.rs
petrosagg: Hm, yeah it could be mistaken for a method that duplicates. Perhaps it could be named `duplicates`? Making it unambiguously a noun, hinting that this will produce the duplicate values?

jswrenn: Actually, given [this comment](https://github.com/rust-itertools/itertools/pull/502#discussion_r559745443), `duplicate` sounds pretty fitting.

How about calling it `duplicates`, instead?

petrosagg:
@jswrenn thanks for taking another look. I just rebased on top of master and I'm happy to resolve the naming as you see fit :)

src/lib.rs
petrosagg: Looks like we arrived at the same conclusion :) I'll rename them to `duplicates` and `duplicates_by`

jswrenn: Don't forget to change the type names, too (`Duplicates` and `DuplicatesBy`)! (Gah. For some reason, *those* still read as verbs in my head...)

petrosagg: For sure, I wouldn't leave some names as `duplicates` and some as `Duplicate`. What am I, some kind of a madman?? :joy:

petrosagg:
@jswrenn should be good to merge now :)

src/duplicates_impl.rs
```diff
@@ -0,0 +1,197 @@
...
+}
+
+/// An iterator adapter to filter out duplicate elements.
+///
+/// See [`.duplicates()`](../trait.Itertools.html#method.duplicates) for more information.
+#[derive(Clone)]
+#[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
+pub struct Duplicates<I: Iterator> {
+    iter: DuplicatesBy<I, I::Item, ()>,
+}
```
jswrenn, line 176: I think this could be simplified as:
```suggestion
type Duplicates<I> = DuplicatesBy<I, I::Item, ()>;
```
@phimuemue, do you have any opinions on this sort of simplification? IIRC, the only real danger in it is if we ever want the implementations of `Duplicates` and `DuplicatesBy` to diverge, it would be a breaking change — but that seems unlikely to me.

petrosagg: In case this helps resolve the question, I followed the prior art from here https://github.com/rust-itertools/itertools/blob/v0.10.0/src/unique_impl.rs#L139-L146

jswrenn: A few recent simplifications we've merged have taken the type alias approach. Let's go with that here, too.

phimuemue: @jswrenn I think your simplification idea is a good one if it is doable.

I actually remember that I wanted to do a similar thing for `Unique`/`UniqueBy`, but maybe I hit some obstacles.

petrosagg: I tried this today and I also hit some obstacles. First, the type alias can't set the `F` generic parameter to `()` because the implementation of `Iterator` on `DuplicatesBy` has the following trait bounds:

```rust
impl<I, V, F> Iterator for DuplicatesBy<I, V, F>
    where I: Iterator,
          V: Eq + Hash,
          F: FnMut(&I::Item) -> V
```

and I can't provide a separate special implementation for the `F=()` case since Rust doesn't have specialization (yet).

The next best thing would be to have `duplicates` return an instance of `DuplicatesBy` with the closure set to the identity function, but this problematic again because the closure is given a shared reference to the item and has to produce an owned value. This would bring back the `Clone` trait bound on the items and be more wasteful in the common case.

It's been a while since I first worked on this but I now have a faint memory of trying to do that from the beginning and hitting the specialization obstacle. 

Unless there is another way I missed I would keep the code as-is to have good better performance if someone just wants duplicates without a key function.

jswrenn: @petrosagg you can unify the implementations like this: https://github.com/jswrenn/rust-itertools/blob/1582e814047560856a767cf280b7617cca6c4736/src/duplicates_impl.rs

petrosagg: @jswrenn that is certainly interesting and educational, I have some reading on HRTB to do :) 

The verbose implementation feels way more approachable to me and in the end it's not that much longer, HRTB are a bit exotic. But feel free to merge whichever version you think is best for itertools.

jswrenn: I'd like to go forward with the unified implementation. Here's why: 

The trickiest part of writing iterators is the methods. The implementations of `next`, `next_back` and any overridden methods tend to be very subtle, and each additional method implemented/modified requires careful testing. For the lifetime of this iterator, we probably _won't_ get any PRs modifying the `Key` abstraction, but we probably _will_ get PRs for optimizing `next`, `next_back` and adding additional overrides. 

Unifying these two similar implementations halves the amount of subtle code that these PRs will have to account for. It'll make those improvements easier to implement, test, and review.

petrosagg: No problem at all. Do you mind giving a short explanation on how it works? I've been trying to understand but I don't get it. 

In particular, in the `ById` case, how can the `key` produced by `self.k.key(&v);` be inserted into the `used` HashMap without lifetime issues? From what I understand a reference to the just produced item is inserted and not the item itself, but then the item surely goes out of scope? What am I missing?

jswrenn: Whoops, you're right: that _doesn't_ work! (Invocations of `duplicates` fail with a compiler error.)

petrosagg: hm, looking at it a bit more I'm not sure the proposed solution works. If I cargo test I get these kind of errors for invocations of `duplicates()`:

```
error[E0277]: the trait bound `for<'i> itertools::duplicates_impl::private::ById: itertools::duplicates_impl::private::Key<'i, i8, i8>` is not satisfied
   --> tests/quick.rs:920:27
    |
287 | fn correct_size_hint<I: Iterator>(mut it: I) -> bool {
    |                         -------- required by this bound in `correct_size_hint`
...
920 |         correct_size_hint(it.duplicates())
    |                           ^^^^^^^^^^^^^^^ the trait `for<'i> itertools::duplicates_impl::private::Key<'i, i8, i8>` is not implemented for `itertools::duplicates_impl::private::ById`
    |
    = help: the following implementations were found:
              <itertools::duplicates_impl::private::ById as itertools::duplicates_impl::private::Key<'i, Item, &'i Item>>
    = note: required because of the requirements on the impl of `Iterator` for `itertools::duplicates_impl::private::DuplicatesBy<Iter<i8>, i8, itertools::duplicates_impl::private::ById>`

error: aborting due to previous error
```

Which matches the intuition I described above, the call tries to use a `for<'i> Key<'i, i8, i8>` but only an implementation of `for<'i> Key<'i, i8, &'i i8>` exists (which in turn is not suitable because we need to store the `Output` into the HashMap. Am I missing something?

jswrenn: Nope! References definitely pose the challenge here.

I wonder if dropping the `Clone` bound is worth the ergonomic cost. E.g., this is the sort of situation in which I would use `duplicates_by` instead of `duplicates`:
```rust
struct Foo {
    bar: String,
    baz: u8,
}

let foos = 
    ('a'..'z').chain('a'..'z').zip(0..)
        .map(|(bar, baz)| Foo {bar: bar.to_string(), baz});

let dups = duplicates_by(foos, |foo| foo.bar); // Compile error!
```
...but that's a compile error.

petrosagg: @jswrenn I cracked the puzzle! I updated the PR which now has a single implementation that is shared not only between `Duplicates` and `DuplicatesBy` but also between the `Iterator` and `DoubleEndedIterator`.

src/duplicates_impl.rs
```diff
@@ -0,0 +1,197 @@
...
+    {
+        type Item = I::Item;
+
+        fn next(&mut self) -> Option<Self::Item> {
+            while let Some(v) = self.iter.next() {
+                if let Some(v) = self.filter(v) {
+                    return Some(v)
+                }
+            }
+            None
```
jswrenn, line 63: In the spirit of #517, it's possible to use internal iteration to replace these while-let loops with `find_map`!

Unfortunately, because rust lacks split borrows, it's **not** possible to _just_ write this:
```suggestion
            self.iter.find_map(|v| self.filter(v))
```

I've prototyped a work-around here: https://github.com/jswrenn/rust-itertools/blob/c7095488caca3db2dcc04b4233e20d57a7f730c9/src/duplicates_impl.rs

jswrenn (changes requested):
Kudos for unifying the implementations! I just have one last edit for you. :-)

src/duplicates_impl.rs
petrosagg: Yeah I hit the same split borrow problem, but once again it looked simpler in my eyes to keep it with the white loop instead of introducing yet another struct just to make a function call work.

In any case, I just pushed your suggestion but it's failing the MSRV tests, do you know why? It's working on my local linux machine and the linux tests on the CI seem to pass :/

petrosagg: Ok, I added a couple more trait bounds and it's working now. I don't know why it failed before

jswrenn (approved):
Thanks for this! This PR turned into a really excellent case-study in unifying similar implementations and using internal iteration, that I think will be really informative elsewhere in itertools.

bors r+

PR #503, opened 2020-12-08
Migrate to Actions-based CI

jswrenn:


jswrenn:
Hm, I was rather hoping that github would run the PR'd action *before* merging it.

Issue #510, opened 2021-01-01
Feature request: partition_result


mingyli:
[The example for `partition_map` ](https://docs.rs/itertools/0.10.0/itertools/trait.Itertools.html#method.partition_map) partitions a sequence of `Result`s by putting all the `Ok` elements in the left and `Err` elements in the right. This is a common enough operation that [OCaml Base offers `partition_result`](https://ocaml.janestreet.com/ocaml-core/latest/doc/base/Base/List/index.html#val-partition_result) which does the same.

PR #511, opened 2021-01-01
Implement partition_result

mingyli:
Addresses #510 

`partition_result` behaves as [OCaml Base's `partition_result`](https://ocaml.janestreet.com/ocaml-core/latest/doc/base/Base/List/index.html#val-partition_result) does:

```rust
use itertools::Itertools;

let successes_and_failures = vec![Ok(1), Err(false), Err(true), Ok(2)];

let (successes, failures): (Vec<_>, Vec<_>) = successes_and_failures
    .into_iter()
    .partition_result();

assert_eq!(successes, [1, 2]);
assert_eq!(failures, [false, true]);
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2494,6 +2494,33 @@ pub trait Itertools : Iterator {
         (left, right)
     }
 
+    /// Partition a sequence of `Result`s into one list of all the `Ok` elements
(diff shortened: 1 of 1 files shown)


jswrenn (approved):
Thank you for this!

bors r+

PR #512, opened 2021-01-01
Add #[track_caller] to iterators that may panic

Inky-developer:
Adds the #[track_caller] attribute to `zip_eq` and `rc_iter`. That leads to more useful error messages in case of a panic.


Inky-developer:
Oops, messed up the minimum supported version. Closing this.

PR #513, opened 2021-01-06
Simplifications for tuple combinations

phimuemue:
This tries to simplify some things around tuple combinations:

* The first 2 commits reuse our `ignore_ident` to avoid redundant macro parameters.
* The remaining commits try to reduce the usage of macro arguments, which possibly helps us in the future to simplify the macros even further.

One question I have: Why do we `Fuse` tuple-combinations for arity >=2, but not for arity 1?

BTW congrats to the 0.10.0 release!

### DIFF
--- a/src/adaptors/mod.rs
+++ b/src/adaptors/mod.rs
@@ -748,7 +748,7 @@ macro_rules! impl_tuple_combination {
                 Self {
                     item: iter.next(),
                     iter: iter.clone(),
-                    c: $P::from(iter),
+                    c: iter.into(),
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


jswrenn (approved):
Thanks for this! 

bors r+

---

> One question I have: Why do we Fuse tuple-combinations for arity >=2, but not for arity 1?

I'm guessing because it's necessary for the correctness of arity >=2 combinations. But, for arity = 1 combinations, it's tuple-combinations can be a no-op. Perhaps a better question is: why is `HasCombination` implemented for 1-ary tuples?

PR #514, opened 2021-01-10
Add Itertools.contains

mmirate:
This is an extremely trivial shorthand for the code-pattern suggested by the docs for [slice::contains](https://doc.rust-lang.org/std/primitive.slice.html#method.contains) in the case when one has a slice of owned data to be queried using borrowed data.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1616,8 +1616,6 @@ pub trait Itertools : Iterator {
     /// iterator will be exhausted.
     ///
     /// ```
-    /// use itertools::Itertools;
-    ///
     /// #[derive(PartialEq, Debug)]
     /// enum Enum { A, B, C, D, E, }
     /// 
@@ -1633,12 +1631,13 @@ pub trait Itertools : Iterator {
     /// // `E` wasn't found, so `iter` is now exhausted
     /// assert_eq!(iter.next(), None);
     /// ```
-    fn contains<Q>(&mut self, query: Q) -> bool
+    fn contains<Q>(&mut self, query: &Q) -> bool
     where
         Self: Sized,
-        Self::Item: PartialEq<Q>,
+        Self::Item: Borrow<Q>,
+        Q: PartialEq,
     {
-        self.any(|x| x == query)
+        self.any(|x| x.borrow() == query)
     }
 
     /// Check whether all elements compare equal.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1608,6 +1608,53 @@ pub trait Itertools : Iterator {
         None
     }
 
+    /// Check whether the iterator contains an item.
+    ///
+    /// If the iterator contains the item prior to its end,
+    /// this method will short-circuit and only partially
+    /// exhaust the iterator.
+    ///
+    /// ```
+    /// use itertools::Itertools;
+    ///
+    /// let data = vec![
+    ///      "foo".to_owned(),
+    ///      "bar".to_owned(),
+    ///      "baz".to_owned(),
+    /// ];
+    /// // this could get expensive:
+    /// assert!(data.contains(&"foo".to_owned()));
+    /// // this, less so:
+    /// assert!(data.iter().contains("foo"));
+    ///
+    /// // now the not-as-motivating tests involving Copy data:
+    ///
+    /// let data = vec![4usize, 5, 1, 3, 0, 2];
+    /// assert!(data.iter().contains(&4));
+    /// assert!(!data.iter().contains(&6));
+    /// for x in (0..50) {
+    ///     assert_eq!(data.contains(&x), data.iter().contains(&x));
+    /// }
+    ///
+    /// let mut it = data.iter();
+    /// assert!(!it.contains(&6));
+    /// assert_eq!(it.next(), None);
+    ///
+    /// let mut it = data.iter();
+    /// assert!(it.contains(&3));
+    /// assert_eq!(it.next(), Some(&0));
+    ///
+    /// let data : Option<usize> = None;
+    /// assert!(!data.into_iter().contains(0));
+    /// ```
+    fn contains<Q>(&mut self, query: Q) -> bool
+    where
+        Self: Sized,
+        Self::Item: PartialEq<Q>,
+    {
+        self.any(|x| x == query)
+    }
+
     /// Check whether all elements compare equal.
     ///
     /// Empty iterators are considered to have equal elements:
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1618,18 +1618,20 @@ pub trait Itertools : Iterator {
     /// iterator will be exhausted.
     ///
     /// ```
+    /// use itertools::Itertools;
+    ///
     /// #[derive(PartialEq, Debug)]
     /// enum Enum { A, B, C, D, E, }
     /// 
     /// let mut iter = vec![Enum::A, Enum::B, Enum::C, Enum::D].into_iter();
     /// 
     /// // search `iter` for `B`
-    /// assert_eq!(iter.contains(Enum::B), true);
+    /// assert_eq!(iter.contains(&Enum::B), true);
     /// // `B` was found, so the iterator now rests at the item after `B` (i.e, `C`).
     /// assert_eq!(iter.next(), Some(Enum::C));
     /// 
     /// // search `iter` for `E`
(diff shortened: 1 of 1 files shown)


phimuemue:
Should we also support `Borrow`? - Like e.g. [`HashSet::get`](https://doc.rust-lang.org/std/collections/struct.HashSet.html#method.get), so that we can test if an iterator over `String`s contains a `&str`.

mmirate:
> Should we also support `Borrow`? - Like e.g. [`HashSet::get`](https://doc.rust-lang.org/std/collections/struct.HashSet.html#method.get), so that we can test if an iterator over `String`s contains a `&str`.

I believe I included an example of this in the doctests.

src/lib.rs
```diff
@@ -1608,6 +1608,53 @@ pub trait Itertools : Iterator {
...
+    ///
+    /// let data = vec![
+    ///      "foo".to_owned(),
+    ///      "bar".to_owned(),
+    ///      "baz".to_owned(),
+    /// ];
+    /// // this could get expensive:
+    /// assert!(data.contains(&"foo".to_owned()));
+    /// // this, less so:
+    /// assert!(data.iter().contains("foo"));
```
jswrenn, line 1628: What makes the former example expensive, and the latter example cheap?

src/lib.rs
```diff
@@ -1608,6 +1608,53 @@ pub trait Itertools : Iterator {
         None
     }
 
+    /// Check whether the iterator contains an item.
+    ///
+    /// If the iterator contains the item prior to its end,
+    /// this method will short-circuit and only partially
+    /// exhaust the iterator.
```
jswrenn, line 1615: Some tweaks for consistency with the documentation of `find`/`contains` methods in the standard library:
```suggestion
    /// Returns `true` if the given item is present in this iterator.
    ///
    /// This method is short-circuiting. If the given item is present in this
    /// iterator, the this method will consume the iterator up-to-and-including
    /// the item. If the given item is not present in this iterator, the
    /// iterator will be exhausted.
```

src/lib.rs
```diff
@@ -1608,6 +1608,53 @@ pub trait Itertools : Iterator {
...
+    /// assert!(!it.contains(&6));
+    /// assert_eq!(it.next(), None);
+    ///
+    /// let mut it = data.iter();
+    /// assert!(it.contains(&3));
+    /// assert_eq!(it.next(), Some(&0));
+    ///
+    /// let data : Option<usize> = None;
+    /// assert!(!data.into_iter().contains(0));
+    /// ```
```
jswrenn, line 1649: This is a lot of unstructured documentation for a simple shorthand. I think this would be sufficient:
```suggestion
    /// ```
    /// use itertools::Itertools;
    ///
    /// #[derive(PartialEq, Debug)]
    /// enum Enum { A, B, C, D, E, }
    /// 
    /// let mut iter = vec![Enum::A, Enum::B, Enum::C, Enum::D].into_iter();
    /// 
    /// // search `iter` for `B`
    /// assert_eq!(iter.contains(Enum::B), true);
    /// // `B` was found, so the iterator now rests at the item after `B` (i.e, `C`).
    /// assert_eq!(iter.next(), Some(Enum::C));
    /// 
    /// // search `iter` for `E`
    /// assert_eq!(iter.contains(Enum::E), false);
    /// // `E` wasn't found, so `iter` is now exhausted
    /// assert_eq!(iter.next(), None);
    /// ```
```

Using a custom type `Enum` has the advantage that if we ever change the trait bounds of this method, we can be sure about whether the change is breaking or not.

jswrenn (changes requested):
Thanks for the contribution!

src/lib.rs
```diff
@@ -1608,6 +1608,53 @@ pub trait Itertools : Iterator {
...
+    fn contains<Q>(&mut self, query: Q) -> bool
+    where
+        Self: Sized,
+        Self::Item: PartialEq<Q>,
+    {
+        self.any(|x| x == query)
+    }
```
jswrenn, line 1657: Concretely, @phimuemue is proposing this, instead ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=7a59790f362e1517873667f4a245679b)):
```suggestion
    /// ```
    /// #[derive(PartialEq, Debug)]
    /// enum Enum { A, B, C, D, E, }
    /// 
    /// let mut iter = vec![Enum::A, Enum::B, Enum::C, Enum::D].into_iter();
    /// 
    /// // search `iter` for `B`
    /// assert_eq!(iter.contains(Enum::B), true);
    /// // `B` was found, so the iterator now rests at the item after `B` (i.e, `C`).
    /// assert_eq!(iter.next(), Some(Enum::C));
    /// 
    /// // search `iter` for `E`
    /// assert_eq!(iter.contains(Enum::E), false);
    /// // `E` wasn't found, so `iter` is now exhausted
    /// assert_eq!(iter.next(), None);
    /// ```
    fn contains<Q>(&mut self, query: &Q) -> bool
    where
        Self: Sized,
        Self::Item: Borrow<Q>,
        Q: PartialEq,
    {
        self.any(|x| x.borrow() == query)
    }
```

This seems like a sensible improvement to me!

src/lib.rs
```diff
@@ -1608,6 +1608,39 @@ pub trait Itertools : Iterator {
...
+    /// // `E` wasn't found, so `iter` is now exhausted
+    /// assert_eq!(iter.next(), None);
+    /// ```
+    fn contains<Q>(&mut self, query: Q) -> bool
+    where
+        Self: Sized,
+        Self::Item: PartialEq<Q>,
+    {
+        self.any(|x| x == query)
+    }
```
jswrenn, line 1642: Concretely, @phimuemue is proposing this, instead ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=7a59790f362e1517873667f4a245679b)):
```suggestion
    /// ```
    /// #[derive(PartialEq, Debug)]
    /// enum Enum { A, B, C, D, E, }
    /// 
    /// let mut iter = vec![Enum::A, Enum::B, Enum::C, Enum::D].into_iter();
    /// 
    /// // search `iter` for `B`
    /// assert_eq!(iter.contains(Enum::B), true);
    /// // `B` was found, so the iterator now rests at the item after `B` (i.e, `C`).
    /// assert_eq!(iter.next(), Some(Enum::C));
    /// 
    /// // search `iter` for `E`
    /// assert_eq!(iter.contains(Enum::E), false);
    /// // `E` wasn't found, so `iter` is now exhausted
    /// assert_eq!(iter.next(), None);
    /// ```
    fn contains<Q>(&mut self, query: &Q) -> bool
    where
        Self: Sized,
        Self::Item: Borrow<Q>,
        Q: PartialEq,
    {
        self.any(|x| x.borrow() == query)
    }
```

This seems like a sensible improvement to me!

src/lib.rs
mmirate: The former example copies the contents of the given &str from wherever it was borrowed into a new heap allocation, and then searches the Vec for it; the latter example does the search without doing the copy-of-contents.

src/lib.rs
```diff
@@ -1608,6 +1608,38 @@ pub trait Itertools : Iterator {
         None
     }
 
+    /// Returns `true` if the given item is present in this iterator.
+    ///
+    /// This method is short-circuiting. If the given item is present in this
+    /// iterator, the this method will consume the iterator up-to-and-including
```
mmirate, line 1614:
```suggestion
    /// iterator, this method will consume the iterator up-to-and-including
```

Typo. Let's see if the mobile app will allow me to correct it in this manner...

src/lib.rs
```diff
@@ -59,6 +59,8 @@ use alloc::{
 
 pub use either::Either;
 
+#[cfg(feature = "use_std")]
+use std::borrow::Borrow;
```
jswrenn, line 63:
```suggestion
use core::borrow:Borrow;
```

jswrenn: ^this is what you need to change to make the CI pass

mmirate:
That seems to be the last of the typos and such. I hate to think how far they would have gone in a dynamic language. 😃

jswrenn (approved):
Thanks for this!

bors r+

mmirate:
My pleasure. :)

PR #516, opened 2021-01-17
Switched numbers to chars so that its clearer

gilescope:
I read the docs in the IDE and I thought the example could be clearer - there were lots of numbers. 
I think this makes it clearer to read as to how the result is structured.


jswrenn (approved):
Thank you so much for this!

bors r+

PR #517, opened 2021-01-20
Use internal iteration in more places

SkiFire13:
I haven't benched it but internal iteration should always be equivalent or faster than external iteration.


jswrenn (approved):
Thanks!

bors r+

PR #518, opened 2021-01-20
Remove unpredicted branch from kmerge::sift_down

SkiFire13:
This is pretty much a port from rust-lang/rust#78857

Compared with the previous implementation, this adds more branches for bound checks which aren't present on the stdlib version, however they should be predicted almost always. The speedup should come from the removal of an unpredictable branch from the loop body, in favor of boolean arithmetic.

The benchmarks seem to agree:

```
before:

test kmerge default ... bench:        6812 ns/iter (+/- 18)
test kmerge tenway ... bench:      223673 ns/iter (+/- 769)

after:

test kmerge default ... bench:        6212 ns/iter (+/- 43)
test kmerge tenway ... bench:      190700 ns/iter (+/- 419)
```


jswrenn (approved):
Thanks!

bors r+

PR #525, opened 2021-02-04
Add a chain! macro.

cassiersg:
This is proposal to add a `chain!` macro which creates an iterator running multiple iterators sequentially.

It basically calls multiple times the standard `.chain()`. This macro is useful to avoid rightwards shift of the code when chaining many iterators.

For the implementation and tests, I took inspiration on the `izip!` macro.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -347,39 +347,68 @@ macro_rules! izip {
 }
 
 #[macro_export]
-/// Create an iterator running multiple iterators sequentially.
+/// [Chain][`chain`] zero or more iterators together into one sequence.
 ///
-/// This is a version of the standard ``.chain()`` that's supporting more than
-/// two iterators. `chain!` takes `IntoIterator` arguments.
-/// Alternatively, this is an alternative to the standard ``.flatten()`` for a
-/// fixed number of iterators of distinct sizes.
+/// The comma-separated arguments must implement [`IntoIterator`].
+/// The final argument may be followed by a trailing comma.
 ///
-/// **Note:** The result of this macro is in the general case an iterator
-/// composed of repeated `.chain()`.
-/// The special case of one arguments produce $a.into_iter().
+/// [`chain`]: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.chain
+/// [`IntoIterator`]: https://doc.rust-lang.org/std/iter/trait.IntoIterator.html
 ///
+/// # Examples
 ///
+/// [`iter::empty`]: https://doc.rust-lang.org/std/iter/fn.empty.html
+///
+/// Empty invocations of `chain!` expand to an invocation of [`iter::empty`]:
 /// ```
-/// # use itertools::chain;
-/// #
-/// # fn main() {
+/// # use std::iter;
+/// use itertools::chain;
 ///
-/// // chain three sequences
-/// let chained: Vec<i32> = chain!(0..=3, 4..6, vec![6, 7]).collect();
-/// assert_eq!(chained, (0..=7).collect::<Vec<i32>>());
-/// # }
+/// let _: iter::Empty<()> = chain!();
+/// let _: iter::Empty<i8> = chain!();
+/// ```
+///
+/// Invocations of `chain!` with one argument expand to [`arg.into_iter()`][`IntoIterator`]:
+/// ```
+/// use std::{ops::Range, slice};
+/// use itertools::chain;
+/// let _: <Range<_> as IntoIterator>::IntoIter = chain!((2..6),); // trailing comma optional!
+/// let _:     <&[_] as IntoIterator>::IntoIter = chain!(&[2, 3, 4]);
+/// ```
+///
+/// Invocations of `chain!` with multiple arguments [`.into_iter()`][`IntoIterator`] each
+/// argument, and then [`chain`] them together:
+/// ```
+/// use std::{iter::*, ops::Range, slice};
+/// use itertools::{assert_equal, chain};
+///
+/// // e.g., this:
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -346,6 +346,40 @@ macro_rules! izip {
...
+/// ```
+macro_rules! chain {
+    ( $first:expr $(, $rest:expr )* $(,)*) => {
+        core::iter::IntoIterator::into_iter($first)
+            $(
+                .chain(
+                    core::iter::IntoIterator::into_iter($rest)
+                )
+            )*
+    };
```
phimuemue, line 380: Should we support the empty case (so that the library users do not need to test if they have more than 0 iterators in their generic code)?

tests/quick.rs
```diff
@@ -526,6 +527,11 @@ quickcheck! {
         correct_size_hint(izip!(filt, b.clone(), c.clone())) &&
             exact_size(izip!(a, b, c))
     }
+
+    fn size_chain_macro(a: Iter<i16, Exact>, b: Iter<i16, Exact>, c: Iter<i16, Exact>) -> bool {
+        correct_size_hint(chain!(a.clone(), b.clone(), c.clone()))
+    }
```
phimuemue, line 533: I hope I don't miss anything, but is `size_hint` special for `chain!`? What about other potentially specialized methods?

phimuemue (commented):
Hi there! I sometimes thought about such a facility, but I never got around to actually writing it. If we decide to keep it, I'd find it nice to support 0 iterators, too.

tests/quick.rs
cassiersg: It is not special. I guess this test shouldn't be there (I juste copied some of the `izip!` tests) since it tests more the properties of the std `chain`: we don't specialize any method in this crate. I don't see how the macro could be wrong while `chain` being correct.

If we were to verify all the methods that are specialized by `Chain`, that would mean writing quite many tests ([here](https://doc.rust-lang.org/nightly/src/core/iter/adapters/chain.rs.html#68) is the list).

src/lib.rs
cassiersg: I guess it doesn't hurt to add a pattern `() => { core::iter::empty() };`.

scottmcm: Definitely good to support `empty` -- even if it's not clearly useful, neither is `chain!(foo)` 😄

cassiersg:
@phimuemue I believe this is ready for a second review.

phimuemue:
Note: I did not test this, but it looks good to me. As far as I can see, one can call `chain(iter1, iter2, ,,,)` (note trailing commas). For me this would be ok, but let's wait for @jswrenn's opinion.

As I do not have that much time to review this, I'd appreciate a second pair eyes on this, to rule out potential pitfalls.

src/lib.rs
```diff
@@ -346,6 +346,43 @@ macro_rules! izip {
...
+/// ```
+/// # use itertools::chain;
+/// #
+/// # fn main() {
+///
+/// // chain three sequences
+/// let chained: Vec<i32> = chain!(0..=3, 4..6, vec![6, 7]).collect();
+/// assert_eq!(chained, (0..=7).collect::<Vec<i32>>());
+/// # }
+/// ```
```
jswrenn, line 371: I think this is a case where a beefier example can take on a lot of the heavy-lifting from the prose; e.g.:
````suggestion
/// [Chain][`chain`] zero or more iterators together into one sequence.
///
/// The comma-separated arguments must implement [`IntoIterator`].
/// The final argument may be followed by a trailing comma.
///
/// [`chain`]: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.chain
/// [`IntoIterator`]: https://doc.rust-lang.org/std/iter/trait.IntoIterator.html
///
/// # Examples
///
/// [`iter::empty`]: https://doc.rust-lang.org/std/iter/fn.empty.html
///
/// Empty invocations of `chain!` expand to an invocation of [`iter::empty`]:
/// ```
/// # use std::iter;
/// use itertools::chain;
///
/// let _: iter::Empty<()> = chain!();
/// let _: iter::Empty<i8> = chain!();
/// ```
///
/// Invocations of `chain!` with one argument expand to [`arg.into_iter()`][`IntoIterator`]:
/// ```
/// use std::{ops::Range, slice};
/// use itertools::chain;
/// let _: <Range<_> as IntoIterator>::IntoIter = chain!((2..6),); // trailing comma optional!
/// let _:     <&[_] as IntoIterator>::IntoIter = chain!(&[2, 3, 4]);
/// ```
///
/// Invocations of `chain!` with multiple arguments [`.into_iter()`][`IntoIterator`] each
/// argument, and then [`chain`] them together:
/// ```
/// use std::{iter::*, ops::Range, slice};
/// use itertools::{assert_equal, chain};
///
/// // e.g., this:
/// let with_macro:  Chain<Chain<Once<_>, Take<Repeat<_>>>, slice::Iter<_>> =
///     chain![once(&0), repeat(&1).take(2), &[2, 3, 5],];
///
/// // ...is equivalant to this:
/// let with_method: Chain<Chain<Once<_>, Take<Repeat<_>>>, slice::Iter<_>> =
///     once(&0)
///         .chain(repeat(&1).take(2))
///         .chain(&[2, 3, 5]);
///
/// assert_equal(with_macro, with_method);
/// ```
````

jswrenn (changes requested):
Thanks for the PR, and sorry for being so slow at reviewing; I've been super tied up lately.

I wonder if it might also make sense to provide a `Chain!` type macro that expands to the type produced by `chain!`, but that doesn't have to be in this PR. I think it might be somewhat tricky to implement, too...

cassiersg:
@jswrenn Thanks for the suggestions !

jswrenn (approved):
Thanks!

bors r+

PR #527, opened 2021-02-07
add .flatten_ok()

### LINKED ISSUE #522: flatten_ok
## Proposal

I propose adding the following (without the `impl Iterator`, just to demonstrate):

```rust
    fn flatten_ok<A, B, E>(self) -> impl Iterator<Item=Result<B, E>>
        where Self: Iterator<Item = Result<A, E>>, A: IntoIterator<Item = B>
    {...}
```

`flatten_ok` operates on an iterator of `Result<A, E>` where `A: IntoIterator<Item = B>`. If the iterator produces `Ok`, the value is flattened. If the iterator produces `Err`, the value is produced exactly as-is.

## Why add this?

Several others have been trying to solve this exact problem:
* https://paulkernfeld.com/2018/11/03/flatten-nested-iterator-result.html
* https://stackoverflow.com/questions/59852161/how-to-handle-result-in-flat-map

I also encountered this problem and was unable to find a combinator for this purpose, which was surprising to me.

## Potential way to implement

It could be implemented this way:
```rust
fn flatten_ok<A, B, E>(self) -> impl Iterator<Item=Result<B, E>>
where Self: Iterator<Item = Result<A, E>>, A: IntoIterator<Item = B>
{
    self.flat_map(|res| Either::from(res)
        .map_right(|ok| ok.into_iter().map(Ok))
        .map_left(Err)
        .map_left(std::iter::once)
    )
}
```

Of course, a struct could be created within `itertools` to implement the `Iterator` so there is no external API dependency on `either`, and so we can still specify a concrete return type, but the above would approximately the implementation used and does work if this signature (with `impl Iterator`) is acceptable.

## Questions

* Is there another and more clear way to implement this that others are missing? If so, please let me know.
* Should this be proposed to `libcore` first?

vadixidav:
Closes #522

### DIFF
--- a/src/flatten_ok.rs
+++ b/src/flatten_ok.rs
@@ -1,3 +1,4 @@
+use crate::size_hint;
 use std::{fmt, iter::FusedIterator};
 
 pub fn flatten_ok<I, T, E>(iter: I) -> FlattenOk<I, T, E>
@@ -5,7 +6,11 @@ where
     I: Iterator<Item = Result<T, E>>,
     T: IntoIterator,
 {
-    FlattenOk { iter, inner: None }
+    FlattenOk {
+        iter,
+        inner_front: None,
+        inner_back: None,
+    }
 }
 
 /// An iterator adaptor that flattens `Result::Ok` values and
@@ -19,7 +24,8 @@ where
     T: IntoIterator,
 {
     iter: I,
-    inner: Option<T::IntoIter>,
+    inner_front: Option<T::IntoIter>,
+    inner_back: Option<T::IntoIter>,
 }
 
 impl<I, T, E> Iterator for FlattenOk<I, T, E>
@@ -31,45 +37,54 @@ where
 
     fn next(&mut self) -> Option<Self::Item> {
         loop {
-            if let Some(inner) = &mut self.inner {
+            // Handle the front inner iterator.
+            if let Some(inner) = &mut self.inner_front {
                 if let Some(item) = inner.next() {
                     return Some(Ok(item));
                 } else {
                     // This is necessary for the iterator to implement `FusedIterator`
                     // with only the orginal iterator being fused.
-                    self.inner = None;
+                    self.inner_front = None;
                 }
             }
 
             match self.iter.next() {
-                Some(Ok(ok)) => self.inner = Some(ok.into_iter()),
+                Some(Ok(ok)) => self.inner_front = Some(ok.into_iter()),
                 Some(Err(e)) => return Some(Err(e)),
-                None => return None,
+                None => {
+                    // Handle the back inner iterator.
+                    if let Some(inner) = &mut self.inner_back {
+                        if let Some(item) = inner.next() {
+                            return Some(Ok(item));
+                        } else {
+                            // This is necessary for the iterator to implement `FusedIterator`
+                            // with only the orginal iterator being fused.
+                            self.inner_back = None;
+                        }
+                    } else {
+                        return None;
+                    }
+                }
             }
         }
     }
 
     fn size_hint(&self) -> (usize, Option<usize>) {
-        if let Some(inner) = &self.inner {
-            // If we have an inner iterator, then its lower bound is our lower bound,
-            // but we still don't know the upper bound.
-            let (inner_lower, inner_higher) = inner.size_hint();
-            let (_, outer_higher) = self.iter.size_hint();
-            if outer_higher == Some(0) {
-                // If there is nothing remaining in the outer iterator, we know the upper bound.
-                (inner_lower, inner_higher)
-            } else {
-                // However, if the outer iterator could have more items in it, we don't
-                // know the upper bound.
-                (inner_lower, None)
-            }
-        } else if self.iter.size_hint() == (0, Some(0)) {
-            // If the outer iterator is empty, we have no items.
-            (0, Some(0))
-        } else {
-            // Otherwise we do not know anything about the number of items.
-            (0, None)
-        }
+        let inner_hint = |inner: &Option<T::IntoIter>| {
+            inner
+                .as_ref()
+                .map(Iterator::size_hint)
+                .unwrap_or((0, Some(0)))
+        };
+        let inner_front = inner_hint(&self.inner_front);
+        let inner_back = inner_hint(&self.inner_back);
+        // The outer iterator `Ok` case could be (0, None) as we don't know its size_hint yet.
+        let outer = match self.iter.size_hint() {
+            (0, Some(0)) => (0, Some(0)),
+            _ => (0, None),
+        };
+
+        size_hint::add(size_hint::add(inner_front, inner_back), outer)
     }
 }
 
@@ -80,7 +95,7 @@ where
     T::IntoIter: Clone,
 {
     #[inline]
-    clone_fields!(iter, inner);
+    clone_fields!(iter, inner_front, inner_back);
 }
 
 impl<I, T, E> fmt::Debug for FlattenOk<I, T, E>
@@ -92,7 +107,8 @@ where
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_struct("FlattenOk")
             .field("iter", &self.iter)
-            .field("inner", &self.inner)
+            .field("inner_front", &self.inner_front)
+            .field("inner_back", &self.inner_back)
             .finish()
     }
 }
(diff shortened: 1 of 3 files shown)


src/flatten_ok.rs
```diff
@@ -0,0 +1,61 @@
...
+
+    fn next(&mut self) -> Option<Self::Item> {
+        loop {
+            if let Some(inner) = &mut self.inner {
+                if let Some(item) = inner.next() {
+                    return Some(Ok(item));
+                } else {
+                    self.inner = None;
+                }
+            } else {
```
SkiFire13, line 38: Do you really need this `else`? The only case where it isn't executed is if `self.inner` is `Some(_)` and `inner.next()` is also `Some(_)`, but in that case you have an early `return`. I would execute the contents of the `else` in any case, both if the `if let` succeeds or not.

SkiFire13:
A couple of other things you could also add:
- Specialized methods of `Iterator`, for example `size_hint` and `fold`
- Some integration tests
- Implementations of `DoubleEndedIterator`, `FusedIterator` and `Debug` to match the std's `FlatMap`
- Since the std's `FlatMap` fuses the outer iterator you might consider that as well for consistency. Note however that this is undocumented behaviour, so this might change in the future.

src/flatten_ok.rs
```diff
@@ -0,0 +1,61 @@
...
+    I: Iterator<Item = Result<T, E>> + Clone,
+    T: IntoIterator,
+    T::IntoIter: Clone,
+{
+    fn clone(&self) -> Self {
+        Self {
+            iter: self.iter.clone(),
+            inner: self.inner.clone(),
+        }
+    }
```
phimuemue, line 60: You could use `clone_fields` for consistency here.

src/flatten_ok.rs
vadixidav: @SkiFire13 Your point is sound. In fact, we shouldn't even need to assign it to `None` in that case as it will just automatically fall through. That might hurt performance in the case where `next()` is called on an empty iterator repeatedly, but I don't think that actually matters. So yes, let me make that change.

SkiFire13: I think setting it to `None` might have a reason if you consider consistency in case `next` is called after the iterator ended. Should it try to advance the last inner iterator? Or the outer iterator? Currently the stdlib fuse both of them, but I guess it would be reasonable if you advance the outer iterator (in that case you need to set the `self.inner` to `None`)

vadixidav:
@SkiFire13 
>     * Specialized methods of `Iterator`, for example `size_hint` and `fold`

I don't think we can implement `size_hint` because even if the inner iterator had thousands of `Ok` elements, there is no guarantee that we will even get a single element.

I could add a specialized implementation of fold, but that could be done in another PR if someone runs into a performance issue with it. I think it should be alright as-is.

>     * Some integration tests

I can add those.

>     * Implementations of `DoubleEndedIterator`, `FusedIterator` and `Debug` to match the std's `FlatMap`

I couldn't think of a way to implement `DoubleEndedIterator` effectively. It would require me to store the last `Ok` and the next `Ok` at the same time. If this is acceptable, I can do it, but it will make the iterator larger and add more logic to retrieve the final elements of the inner iterator.

I can definitely implement `FusedIterator` and `Debug`, so I will do that.

>     * Since the std's `FlatMap` fuses the outer iterator you might consider that as well for consistency. Note however that this is undocumented behaviour, so this might change in the future.

I think I will just allow the user to call `.fuse()` should they need that behavior. I will implement `FusedIterator` if the input iterator is fused, which should make it maximally flexible for the time being.

src/flatten_ok.rs
vadixidav: Thank you. Fixed and will push shortly.

src/flatten_ok.rs
vadixidav: I realize now that I should not change setting the inner to `None` as it might incur a performance penalty when hitting several `Err` values in a row.

SkiFire13:
> I don't think we can implement size_hint because even if the inner iterator had thousands of Ok elements, there is no guarantee that we will even get a single element.

I think you can use the current inner iterator's `size_hint` for the lower bound. Far from perfect, but better than nothing.

> I could add a specialized implementation of fold, but that could be done in another PR if someone runs into a performance issue with it. I think it should be alright as-is.

I think it's good practice to provide specialized implementations of `fold`, especially if `next` contains complex logic because that can prevent LLVM from autovectorizing the loops. Anyway I can add it later.

> I couldn't think of a way to implement DoubleEndedIterator effectively. It would require me to store the last Ok and the next Ok at the same time. If this is acceptable, I can do it, but it will make the iterator larger and add more logic to retrieve the final elements of the inner iterator.

The stdlib does exactly that. I don't think it's a big deal if the iterator is larger since it's supposed to be short lived and probably inlined. The additional check shouldn't matter that much since if you never call `next_back` it's only for the very last element.

vadixidav:
@SkiFire13 Okay, in that case I will add a size_hint impl. It wont help in the case that the iterator hasn't already had `next` called on it at least once, since the inner iterator is `None`, but perhaps if someone partially consumes an iterator and then tries to collect it then it might help to a degree. I will hold off on adding a specialized `fold` impl for now, and that can be added later. I will also implement `DoubleEndedIterator` now based on your comment, since that is something that matters to the API.

vadixidav:
The `size_hint` impl actually was able to glean more information than I originally thought, but only slightly. It should be helpful in some limited cases.

vadixidav:
@SkiFire13 Okay, I believe I am done now. I did not add tests for `size_hint` because I couldn't think of any good way to test that, but if you would like me to do that, I can add some tests for `size_hint`. If there is some fuzzer I can hook it up to for testing `size_hint`, that would be great. I can produce random sequences of iterators that could be validated by the fuzzer to not violate `size_hint`'s contract, if we have that (I only took a cursory look at other tests).

vadixidav:
Okay, I just looked over my code and I think there may be an issue if the inner iterator is not fused. Notably, the call to `next()` on the outer iterator now happens multiple times in the case that you have to traverse the back iterator. I am going to fix that right now.

vadixidav:
I actually made the "fix" which fuses the outer iterator, and then realized, its technically still correct the way it is because if the outer iterator isn't fused, neither is `FlattenOk`, so this is actually fine. I was just confused :sweat_smile:. I am going to leave this as is. Let me know if its good. I will save the commit which fuses the iterator in another branch on my fork, but its a bit ugly.

SkiFire13:
> Okay, I just looked over my code and I think there may be an issue if the outer iterator is not fused. Notably, the call to next() on the outer iterator now happens multiple times in the case that you have to traverse the back iterator. I am going to fix that right now.

I guess that's why the stdlib decided to always fuse the outer iterator too.

> its a bit ugly

You could use the `Fuse` adapter instead of that `bool` flag

vadixidav:
@SkiFire13 That's alright, I think this is good to go if you think so. If you do want it to be fused to merge though let me know, but for now its only fused if the input iterator is fused.

SkiFire13:
For merging you should ask @phimuemue

src/flatten_ok.rs
```diff
@@ -0,0 +1,166 @@
...
+    T: IntoIterator,
+    T::IntoIter: fmt::Debug,
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("FlattenOk")
+            .field("iter", &self.iter)
+            .field("inner_front", &self.inner_front)
+            .field("inner_back", &self.inner_back)
+            .finish()
+    }
```
phimuemue, line 157: Maybe you could use `debug_fmt_fields` for consistency here.

phimuemue (commented):
Hi there. I think there's really progress been made on this. Thank you for taking the time.

Unfortunately, I do not have enough capacity to review pull requests that ample right now.

However, one thing that might simplify review: As far as I can see, this somewhat resembles [`FlattenCompat`](https://github.com/rust-lang/rust/blob/master/library/core/src/iter/adapters/flatten.rs#L233-L431) - as suggested [here](https://github.com/rust-itertools/itertools/issues/522#issuecomment-771964147) (btw thanks @SkiFire13). Maybe we could document this somewhere and even adapt `FlattenCompat`s structure (i.e. naming of fields, `match` vs `if let`) so that a reviewer can basically diff your implementation against `std`'s `FlattenCompat`.

jswrenn (approved):
Thanks for the contribution! (And sorry for taking so long to merge it!)

bors r+

PR #529, opened 2021-02-12
add process_results_ref

xu-cheng:
It is similar to the existing process_results, except it accepts an iterator of `&Result<T, E>` instead of `Result<T, E>`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -135,7 +135,7 @@ pub mod structs {
     pub use crate::peeking_take_while::PeekingTakeWhile;
     #[cfg(feature = "use_alloc")]
     pub use crate::permutations::Permutations;
-    pub use crate::process_results_impl::ProcessResults;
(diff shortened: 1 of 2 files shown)


SkiFire13:
Isn't this the same as `process_results(iterator.map(Result::as_ref), closure)`?

scottmcm:
Yeah, can you elaborate on why you need this over `&Result`?  That's sortof a weird type -- like with `Option`, it's generally better to convert the `&Result<T, E>` to a `Result<&T, &E>` before using it.  See, for example, how `Result::map` doesn't have a `&self` version, expecting you to instead do `.as_ref().map(...)` or similar instead.

xu-cheng:
Thanks for the comments. Yeah, it can be substituted with `process_results(iterator.map(Result::as_ref), closure)`. I mainly added this because I thought it is more convenient to have it. I didn’t realize that using `as_ref` is more idiomatic.

PR #531, opened 2021-02-27
Implement DoubleEndedIterator and FusedIterator for ZipEq.

KarelPeeters:
I wanted to use `zip_eq(..).rev()` and that required a `DoubleEndedIterator` implementation, so I added one. I checked the other "marker" traits and it was pretty easy to add a `FusedIterator` impl as well. The generic trait bounds are a bit too strong now, technically we don't ned both of the inner iterators to be a `FusedIterator`, but I don't think we can do better in the current rust version.

I didn't create an issue for this since it's such a small change, I hope that's okay.

### DIFF
--- a/src/zip_eq_impl.rs
+++ b/src/zip_eq_impl.rs
@@ -1,3 +1,5 @@
+use std::iter::FusedIterator;
+
 use super::size_hint;
 
 /// An iterator which iterates two other iterators simultaneously
@@ -45,7 +47,7 @@ impl<I, J> Iterator for ZipEq<I, J>
             (None, None) => None,
             (Some(a), Some(b)) => Some((a, b)),
             (None, Some(_)) | (Some(_), None) =>
-            panic!("itertools: .zip_eq() reached end of one iterator before the other")
+                panic!("itertools: .zip_eq() reached end of one iterator before the other")
         }
     }
 
@@ -54,7 +56,27 @@ impl<I, J> Iterator for ZipEq<I, J>
     }
 }
 
+impl<I, J> DoubleEndedIterator for ZipEq<I, J>
+    where I: DoubleEndedIterator,
+          J: DoubleEndedIterator {
+    fn next_back(&mut self) -> Option<Self::Item> {
+        match (self.a.next_back(), self.b.next_back()) {
+            (None, None) => None,
+            (Some(a), Some(b)) => Some((a, b)),
+            (None, Some(_)) | (Some(_), None) =>
+                panic!("itertools: .zip_eq() reached start of one iterator before the other")
+        }
+    }
+}
+
 impl<I, J> ExactSizeIterator for ZipEq<I, J>
     where I: ExactSizeIterator,
           J: ExactSizeIterator
+{}
+
+// technically we only need at least one of the iterators to implement FusedIterator, but for that we either need
+// negative trait bounds or maybe specialization
+impl<I, J> FusedIterator for ZipEq<I, J>
+    where I: FusedIterator,
+          J: FusedIterator
 {}
--- a/tests/zip.rs
+++ b/tests/zip.rs
@@ -29,8 +29,8 @@ fn test_zip_longest_size_hint() {
 fn test_double_ended_zip_longest() {
     let xs = [1, 2, 3, 4, 5, 6];
     let ys = [1, 2, 3, 7];
-    let a = xs.iter().map(|&x| x);
-    let b = ys.iter().map(|&x| x);
+    let a = xs.iter().copied();
+    let b = ys.iter().copied();
     let mut it = a.zip_longest(b);
     assert_eq!(it.next(), Some(Both(1, 1)));
     assert_eq!(it.next(), Some(Both(2, 2)));
@@ -45,8 +45,8 @@ fn test_double_ended_zip_longest() {
 fn test_double_ended_zip() {
     let xs = [1, 2, 3, 4, 5, 6];
     let ys = [1, 2, 3, 7];
-    let a = xs.iter().map(|&x| x);
-    let b = ys.iter().map(|&x| x);
+    let a = xs.iter().copied();
+    let b = ys.iter().copied();
     let mut it = multizip((a, b));
     assert_eq!(it.next_back(), Some((4, 7)));
     assert_eq!(it.next_back(), Some((3, 3)));
@@ -74,4 +74,19 @@ fn zip_eq_panic2()
     let b = [1, 2, 3];
 
     zip_eq(&a, &b).count();
+}
+
+#[test]
+fn zip_eq_backwards() {
+    let mut it = zip_eq(0..3, 4..7);
+    assert_eq!(it.next_back(), Some((2, 6)));
+    assert_eq!(it.next_back(), Some((1, 5)));
+    assert_eq!(it.next_back(), Some((0, 4)));
+    assert_eq!(it.next_back(), None);
+}
+
+#[should_panic]
+#[test]
+fn zip_eq_backwards_panic() {
+    zip_eq(&[1, 2], &[1, 2, 3]).rev().count();
 }


KarelPeeters:
I also noticed we're missing custom implementations of `fold`, `try_fold`, `nth`, `advance_by` etc, but I'm not sure if they are worth the added bloat. Let me know if you want me to send a PR for those as well.

scottmcm:
I think there's an important semantic question about what `zip_eq` is *supposed* to mean.

Notably, with this change, going forward or backwards give different elements if the inputs have different lengths.  And if iteration stops early for some reason -- `any`, `take`, etc -- then one might not notice the problem.  For a concrete example, `zip_eq(0..4, 0..8).all(|(a, b)| a == b)` and `zip_eq(0..4, 0..8).rev().all(|(a, b)| a == b)` do different things.

Is that ok? I don't know.  What *should* happen here? I don't know that either.

scottmcm:
As a thought: `zip_eq` says it panics for different lengths, but it doesn't.

Maybe it could make sense to do a check like this?
```rust
let (hint1, hint2) = (a.size_hint(), b.size_hint());
assert!(hint1.0 <= hint2.1.unwrap_or(usize::MAX) && hint2.0 <= hint1.1.unwrap_or(usize::MAX);
```

That would at least catch the easy cases early, without changing the trait requirements.

jswrenn:
> Notably, with this change, going forward or backwards give different elements if the inputs have different lengths.

I think it'd be fine if this PR resolved that issue by bounding the `DoubleEndedIterator` impl with an `ExactSizeIterator` bound. (The standard library's `Zip` does this too.)

---

> As a thought: `zip_eq` says it panics for different lengths, but it doesn't.

[Concretely:](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=d8a950e321a76719dfb0a884c4a93b96)
```rust
#[cfg(test)]
use itertools::zip_eq;

#[test]
#[should_panic]
fn test() {
    let short = 0..1;
    let long  = 0..9;
    let iter = zip_eq(short, long); // no panic :(
}
```

That's neat. I was aware of the behavior, but it hadn't noticed how odd it was that the documentation implied that `zip_eq`, itself, panicked. This is weird, and totally needs to be sign-posted better at minimum.

Here's a thought:
  * define `zip_eq` *with* an `ExactSizeIterator` bound
  * define `zip_lazy_eq` *without* an `ExactSizeIterator` bound

scottmcm:
> (The standard library's Zip does this too.)

That's not perfectly analogous, since different lengths are acceptable with `std`'s `zip`, so it uses ESI to skip the extra elements in the longer one so that iterating backwards gets the correct pairs of elements.

So it would feel kinda weird to me to have the forward direction here panic lazily but the backward direction panic eagerly.

> - define zip_eq with an ExactSizeIterator bound
> - define zip_lazy_eq without an ExactSizeIterator bound

In a way, this is what my size_hint check is doing, just in the same method -- if you pass two ESIs, it'll panic immediately.  But it'll *also* panic immediately for `zip_eq(0..10, (0..8).filter(f))` even though the latter isn't ESI, because it can never be long enough to match the length of the former.

So it might be ok to just carefully document "⚠ This may do weird (but safe) things if you pass non-equal-length iterators ⚠" and treat it as "garbage in garbage out" in those cases.

jswrenn:
> So it would feel kinda weird to me to have the forward direction here panic lazily but the backward direction panic eagerly.

Ah, yeah.

> In a way, this is what my size_hint check is doing, just in the same method -- if you pass two ESIs, it'll panic immediately. But it'll _also_ panic immediately for `zip_eq(0..10, (0..8).filter(f))` even though the latter isn't ESI, because it can never be long enough to match the length of the former.

I think I'm sold on this approach, but I wish I knew more about how people were using `zip_eq` in practice. Why reach for `zip_eq` instead of `zip`? Is the lazy panicking important? If not, it might make more sense to require ESI on `zip_eq` and panic eagerly. (Rayon's `zip_eq` does this.)

scottmcm:
> I wish I knew more about how people were using zip_eq in practice

Yeah, would be nice to know whether things are ESI in practice.

👍 to following rayon.  The more code can flip back and forth between serial and parallel the better.

I guess it could just change to ESI and early panic in a major version bump and see who complains 😆

phimuemue:
> [...] with this change, going forward or backwards give different elements if the inputs have different lengths. [...]
> Is that ok? I don't know. What _should_ happen here? I don't know that either.

I once stumbled upon a similar issue, and PR'd `DoubleEndedIterator` (see https://github.com/rust-lang/rust/pull/72626), which was accepted. (The documentation for it [now explicitly states](https://doc.rust-lang.org/std/iter/trait.DoubleEndedIterator.html#remarks) that the elements of `DoubleEndedIterator` may differ from the `Iterator` ones.)

Not sure, though, if that applies to `panic`s, too.

jswrenn:
Fascinating! Given that, I don't have any objections to merging this PR.

The issue of a panicking `next()`/`next_back()` seems orthogonal to the directionality concerns. I'm going to merge this PR, but I think we should revisit bounding `zip_eq` with `ExactSizeIterator`.

KarelPeeters:
I'm not so sure about this PR any more, some good issues were raised. I didn't even think of the different-size-different-elements problem, beucase it doesn't really matter for my use cases. That's mostly because of the fact that `Iterator` is a bit of an underspecified state machine, the only thing I can find in the docs about this issue is this sentence in the [`DoubleEndedIterator` docs](https://doc.rust-lang.org/std/iter/trait.DoubleEndedIterator.html):

> It is important to note that both back and forth work on the same range, and do not cross: iteration is over when they meet in the middle.

They _work on the same range_, does that also mean they have to work on the same values? Hard to tell... I'm pretty sure almost all of the code out there that accepts a `DoubleEndedIterator` as a parameter to return the same values in both directions, but I don't know how many such code there actually is.

To give some perspective, I'm kind of confused why the default version of zip is the one that silently ignores differently sized iterators. I grepped a few of my projects in `Rust`, `Python` and `Kotlin` and >90% of the `zip` use cases are actually `zip_eq` use cases, either with an assert that checks the sizes right above it or where it would probably be better to add that assert. In my experience most of the time you're iterating over two collections with the same size and you just want to add a cautionary assert. The only use case for normal `zip` I found is when you have an infinite iterator and a limited one.

In that light it doesn't really matter for me whether the assert happens at the start or the end, but of course the start is of course preferable. I really like the `zip_eq` vs `zip_eq_lazy` solution, although of course that's not backwards compatible. It may be worth it though, I didn't find many cases where the participating iterators are not `ExactSizeIterator`. Something like "keep the current `zip_eq` but add `zip_eq_eager` where only th elatter implements `ExactSizeIterator` is another option.

I could be interesting to grep crates that use `itertools::zip` or `itertools::zip_eq` and see what they're using it for, I'll give that a shot tomorrow.

KarelPeeters:
Okay I looked around a bit on GitHub, this search query is the best I could come up with: [`zip_eq language:Rust -simultaneously`](https://github.com/search?o=desc&p=2&q=zip_eq+language%3ARust+-simultaneously&s=indexed&type=Code) (the `-simultaneously` filters out the file that declares `zip_eq`, because filtering by filename is currently broken in GitHub search).

It looks like most uses are with `ExactSizeIterator`s, mostly from `Vec`s or slices. I only came across a few non-ESI use cases and one of them was with a custom iterator that could easily implement it. I think adding an `ESI` bound to `zip_eq` and introducing a new `zip_eq_lazy` would not break a lot of dependants.

Another interesting datapoint is that `rayon` also has a [`zip_eq`](https://docs.rs/rayon/1.5.0/rayon/iter/trait.IndexedParallelIterator.html#method.zip_eq) function, it's only implemented for `IndexedParallelIterator` which includes the requirements for `ESI`. It [panics immediatly](https://docs.rs/rayon/1.5.0/src/rayon/iter/mod.rs.html#2275) when `zip_eq` is called.

PR #532, opened 2021-03-12
Add take_borrowed iterator adaptor

joshwd36:
This adds the `take_borrowed(n)` method to the itertools trait. This produces an iterator over the first `n` elements, similar to `take`. However, it borrows from the original iterator, leaving the remaining elements once the returned iterator has been evaluated. This provides a simple way of splitting an iterator in two.

The only doubt I have about this is what should happen if the returned iterator isn't fully evaluated. Currently, if it is only evaluated k times, then only the first k elements will be removed from the original. It might be a better idea to add a `drop` implementation to skip the remaining elements, or perhaps add that as an option. I'm not sure what would be the most logical in that situation.

### DIFF
--- a/src/adaptors/mod.rs
+++ b/src/adaptors/mod.rs
@@ -1081,3 +1081,49 @@ where
         }
     }
 }
+
+/// An iterator adapter over the first n elements while borrowing the original iterator.
+///
+/// See [`.take_borrowed()`](crate::Itertools::take_borrowed) for more information.
+#[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
+pub struct TakeBorrowed<'a, I: Iterator> {
+    iter: &'a mut I,
+    n: usize,
+}
+
+/// Create a new `TakeBorrowed` iterator.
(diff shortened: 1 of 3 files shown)


phimuemue:
Hi there! I must say you matched the implementation of [`Take`](https://doc.rust-lang.org/src/core/iter/adapters/take.rs.html#15-114) quite nicely.

However, isn't this essentially the same as `iter`[`.by_ref`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.by_ref)`.take`?

joshwd36:
> However, isn't this essentially the same as `iter`[`.by_ref`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.by_ref)`.take`?

Ah yes, you're right. I did think this must be quite a common pattern but couldn't see it described anywhere online, and didn't think to use as_ref. Should I close this now, or would it be worth having this as well, perhaps with the drop implementation?

jswrenn (commented):
I'm tempted to say that `by_ref` is the right tool for the job here (and in every other case where one wants to 'split' an iterator). I agree completely that `by_ref` under-advertised and under-appreciated, but there isn't too much itertools can do about that.

PR #534, opened 2021-04-08
Clarify how `next()` interacts with `peek()` in` MultiPeek`.

b05902132:
While intuitively this is the most reasonable behavior, it's nice to document it.

### DIFF
--- a/src/multipeek_impl.rs
+++ b/src/multipeek_impl.rs
@@ -38,6 +38,7 @@ impl<I: Iterator> MultiPeek<I> {
     /// Works exactly like `.next()` with the only difference that it doesn't
     /// advance itself. `.peek()` can be called multiple times, to peek
     /// further ahead.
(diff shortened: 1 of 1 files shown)


src/multipeek_impl.rs
```diff
@@ -38,6 +38,7 @@ impl<I: Iterator> MultiPeek<I> {
     /// Works exactly like `.next()` with the only difference that it doesn't
     /// advance itself. `.peek()` can be called multiple times, to peek
     /// further ahead.
+    /// When `.next()` is called, reset the peeking “cursor”.
```
phimuemue, line 41: I like the idea of documenting this. However, when I read this, it was not immediately obvious to me what the "cursor" even is. Maybe we could use examples in there?

phimuemue:
Thanks. I'll merge this as it does - imho - not worsen the documentation.

bors r+

PR #535, opened 2021-04-10
Add `find_or_last` method to `Itertools` trait

mankinskin:
[Playground](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=6abc43c9037445f88fb0a86991da01f2)

```rust
fn find_or_last(mut self, predicate: impl FnMut(&Self::Item) -> bool) -> Option<Self::Item>
```
- returns `None` if iterator is empty
- returns `Some(element)` when element matches the predicate
- returns `Some(last_element)` when no element matches the predicate

[Related discussion on the rust user forum](https://users.rust-lang.org/t/find-or-return-last-element-of-iterator/58171)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1738,7 +1738,7 @@ pub trait Itertools : Iterator {
     /// use itertools::Itertools;
     ///
     /// let numbers = [1, 2, 3, 4];
-    /// assert_eq!(numbers.iter().find_or_last(|x| x > 5), Some(4));
+    /// assert_eq!(numbers.iter().find_or_last(|&&x| x > 5), Some(&4));
     /// ```
 	fn find_or_last<P>(mut self, predicate: P) -> Option<Self::Item>
 		where Self: Sized,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1750,6 +1750,29 @@ pub trait Itertools : Iterator {
         self.find_map(|x| if predicate(&x) { Some(x) } else { prev = Some(x); None })
             .or(prev)
     }
+    /// Find the value of the first element satisfying a predicate or return the first element, if any.
+    ///
+    /// The iterator is not advanced past the first element found.
+    ///
+    /// ```
+    /// use itertools::Itertools;
+    ///
+    /// let numbers = [1, 2, 3, 4];
+    /// assert_eq!(numbers.iter().find_or_first(|&&x| x > 5), Some(&1));
+    /// assert_eq!(numbers.iter().find_or_first(|&&x| x > 2), Some(&3));
+    /// assert_eq!(std::iter::empty::<i32>().find_or_first(|&x| x > 5), None);
+    /// ```
+    fn find_or_first<P>(mut self, mut predicate: P) -> Option<Self::Item>
+        where Self: Sized,
+              P: FnMut(&Self::Item) -> bool,
+    {
+        let first = self.next()?;
+        Some(if predicate(&first) {
+            first
+        } else {
+            self.find(|x| predicate(&x)).unwrap_or(first)
+        })
+    }
     /// Returns `true` if the given item is present in this iterator.
     ///
     /// This method is short-circuiting. If the given item is present in this
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,7 +1,6 @@
 #![warn(missing_docs)]
 #![crate_name="itertools"]
 #![cfg_attr(not(feature = "use_std"), no_std)]
-#![feature(control_flow_enum)]
 
 //! Extra iterator adaptors, functions and macros.
 //!
@@ -76,7 +75,6 @@ use std::fmt::Write;
 type VecIntoIter<T> = alloc::vec::IntoIter<T>;
 #[cfg(feature = "use_alloc")]
 use std::iter::FromIterator;
-use std::ops::ControlFlow;
 
 #[macro_use]
 mod impl_macros;
@@ -1747,16 +1745,13 @@ pub trait Itertools : Iterator {
 			  P: FnMut(&Self::Item) -> bool,
 	{
 		#[inline]
-        fn check<T>(mut predicate: impl FnMut(&T) -> bool) -> impl FnMut(Option<T>, T) -> ControlFlow<T, Option<T>> {
+        fn check<T>(mut predicate: impl FnMut(&T) -> bool) -> impl FnMut(Option<T>, T) -> Result<Option<T>, T> {
             move |_, x| {
-                if predicate(&x) { ControlFlow::Break(x) } else { ControlFlow::Continue(Some(x)) }
+                if predicate(&x) { Result::Err(x) } else { Result::Ok(Some(x)) }
             }
         }
 
-        match self.try_fold(None, check(predicate)) {
-			ControlFlow::Continue(x) => x,
-			ControlFlow::Break(x) => Some(x),
-		}
+        self.try_fold(None, check(predicate)).unwrap_or_else(Some)
 	}
     /// Returns `true` if the given item is present in this iterator.
     ///
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1742,18 +1742,13 @@ pub trait Itertools : Iterator {
     /// assert_eq!(numbers.iter().find_or_last(|&&x| x > 2), Some(&3));
     /// assert_eq!(std::iter::empty::<i32>().find_or_last(|&x| x > 5), None);
     /// ```
-    fn find_or_last<P>(mut self, predicate: P) -> Option<Self::Item>
+    fn find_or_last<P>(mut self, mut predicate: P) -> Option<Self::Item>
         where Self: Sized,
               P: FnMut(&Self::Item) -> bool,
     {
-        #[inline]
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1730,7 +1730,31 @@ pub trait Itertools : Iterator {
...
+    /// assert_eq!(numbers.iter().find_or_last(|&&x| x > 5), Some(&4));
+    /// assert_eq!(numbers.iter().find_or_last(|&&x| x > 2), Some(&3));
+    /// assert_eq!(std::iter::empty::<i32>().find_or_last(|&x| x > 5), None);
+    /// ```
+	fn find_or_last<P>(mut self, predicate: P) -> Option<Self::Item>
+		where Self: Sized,
+			  P: FnMut(&Self::Item) -> bool,
+	{
+		#[inline]
+        fn check<T>(mut predicate: impl FnMut(&T) -> bool) -> impl FnMut(Option<T>, T) -> Result<Option<T>, T> {
```
phimuemue, line 1750: I've seen the pattern of `inline`d local functions sometimes, but I do not know if we use it regularly in `Itertools`. Could you possibly shed some light on the benefit of this trick? Does it generate more efficient code?

src/lib.rs
```diff
@@ -1730,7 +1730,31 @@ pub trait Itertools : Iterator {
...
+    /// let numbers = [1, 2, 3, 4];
+    /// assert_eq!(numbers.iter().find_or_last(|&&x| x > 5), Some(&4));
+    /// assert_eq!(numbers.iter().find_or_last(|&&x| x > 2), Some(&3));
+    /// assert_eq!(std::iter::empty::<i32>().find_or_last(|&x| x > 5), None);
+    /// ```
+	fn find_or_last<P>(mut self, predicate: P) -> Option<Self::Item>
+		where Self: Sized,
+			  P: FnMut(&Self::Item) -> bool,
+	{
+		#[inline]
```
phimuemue, line 1749: Totally nit-picking: Indentation.

phimuemue (commented):
Hi there! I know of libraries that support an operation like this, so this might be useful for `Itertools`, too.

One question: Might it be useful for the user to distinguish between  "successful search" and "last element because nothing else succeeded"? (We have a similar (or isn't it?) issue with `exactly_one` where a fail could occur because there are either no elements or more than one element.)

src/lib.rs
scottmcm: What it's trying to do is reduce the number of monomorphizations that happen, and thus be a bit faster to compile.  A closure would monomorphize for every iterator type *and* every predicate type, whereas the local function just monomorphizes for the latter.  That said, I'm not sure how essential that is here.  (It makes more sense inside `try_fold`, where predicates like "always returns `Ok`" are more plausible to see in multiple places times.)

mankinskin: It seems to be that way ([inlining docs](https://nnethercote.github.io/perf-book/inlining.html)). The way I came up with this implementation is by basically combining [Iterator::last](https://doc.rust-lang.org/src/core/iter/traits/iterator.rs.html#273-275) and [Iterator::find](https://doc.rust-lang.org/src/core/iter/traits/iterator.rs.html#2323-2326), where they used `#[inline]`.

src/lib.rs
mankinskin: Sure, I guess I haven't configured my editor (vscode) correctly..

src/lib.rs
mankinskin: @scottmcm [try_fold](https://doc.rust-lang.org/src/core/iter/traits/iterator.rs.html#1973-1986) only uses the provided predicate, I am not sure what you mean with using `#[inline]` inside `try_fold`.

scottmcm:
Pondering: why is _last_ the right option?  Should there also be `find_or_first`?

Would it be fine for people to just do `.with_position().find(|x| x.is_last() || pred(x.as_inner()))`?  (Which admittedly doesn't compile right now since those methods don't exist on `itertools::Position`, but they'd be easy additions.)

Some thoughts on [alternative](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=132d603acd56d5d3f1a7c0b0fa6fee72) implementation strategies:

My first thought was that this should do the `fold1` trick to make the accumulator be a `T` instead of an `Option<T>` to reduce the size of the variable getting passed along:
```rust
pub fn find_or_last<T>(
    mut it: impl Iterator<Item = T>,
    mut pred: impl FnMut(&T) -> bool,
) -> Option<T> {
    let first = it.next()?;
    Some(
        it.try_fold(first, |_, x| if pred(&x) { Err(x) } else { Ok(x) })
            .map_or_else(|x| x, |x| x),
    )
}
```

But the previous accumulator being unused lead to my second thought, namely that if the code doesn't need move access to the accumulator, it probably [shouldn't thread it at all](https://github.com/rust-lang/rust/issues/76725).  That would look something like this:
```rust
pub fn find_or_last<T>(
    mut it: impl Iterator<Item = T>,
    mut pred: impl FnMut(&T) -> bool,
) -> Option<T> {
    let mut prev = None;
    it.try_for_each(|x| if pred(&x) { Err(x) } else { Ok(prev = Some(x)) })
        .err()
        .or(prev)
}
```
Which I find pretty elegant, actually.  (And avoids the extra `.next()` callsite of the previous one, which generally helps binary size.)

src/lib.rs
scottmcm: See https://github.com/rust-lang/rust/pull/62429 for why the ones in `core` are written like that.

mankinskin:
@scottmcm you might want to look at [this reply by steffahn](https://users.rust-lang.org/t/find-or-return-last-element-of-iterator/58171/20), where he suggests the same implementation, with some more refactoring. The result would look like this:
```rust
fn find_or_last<P>(
    mut self,
    predicate: P,
) -> Option<Self::Item>
    where Self: Sized,
          P: FnMut(&Self::Item) -> bool,
{
    let mut prev = None;
    self.find_map(|x| if predicate(&x) { Err(x) } else { Ok(prev = Some(x)) })
        .or(prev)
}
```
if I am not mistaken.

mankinskin:
About the option of using `with_position`, this would probably be less efficient, because wraps each value in an enum and needs to peek for the next element.

But `with_position` could probably be used to satisfy more use-cases, if we got that code example you gave to compile.

About why I picked the last value: In my use-case it really doesn't matter which value is returned, I just intuitively thought taking the last value would be the most efficient. But, as it turns out, getting the last value requires an accumulator, so it is probably less efficient than taking the first element.

mankinskin:
I suppose a `find_or_first` would just look like this?
```rust
fn find_or_first<P>(mut self, mut predicate: P) -> Option<Self::Item>
    where Self: Sized,
          P: FnMut(&Self::Item) -> bool,
{
    let first = self.next()?;
    Some(if predicate(&first) {
        first
    } else {
        self.find(|x| predicate(&x)).unwrap_or(first)
    })
}
```

phimuemue (commented):
Apologies if this review is too brief, I do not have much time at the moment.

As far as I can see, the implementations look ok, although I preferred the `try_fold` approach for some reason... Cannot really substantiate this feeling. Maybe it would lead to similar implementations for `find_or_last` and `find_or_first`? Anyways, my feeling should not be a blocker.

My remark regarding the return type stands: Maybe some users want to differentiate between "nothing found because iterator empty" and "nothing found, so the last (resp. first) element was returned instead". On the other hand, we could include your change and wait for actual complaints from users.

SkiFire13:
> although I preferred the try_fold approach for some reason

Note that custom iterators can't implement `try_fold` because it requires the unstable `Try` trait, however they can implement `find_map`

mankinskin:
@phimuemue yes, the idea is still good, and I don't see how it would hurt anything, except maybe requiring something like an `.into_inner()` call on the output to convert it to an `Option`, in case you don't care about how the value was selected:
```rust
enum FindOrDefaultResult<T> {
    Found(T),
    NotFound(T),
    Empty,
}
impl<T> FindOrDefaultResult<T> {
    pub fn into_inner(self) -> Option<T> {
        match self {
            Self::Found(x) |
            Self::NotFound(x) => Some(x),
            Empty => None,
        }
    }
}
```

Theoretically this could even lead to more kinds of `find_or_default` implementations, where the default somehow depends on the iterator.. for example by using an enum to specify how to select the default from the iterator:
```rust
enum DefaultSelector {
    First,
    Last,
    AtPosition(usize),
    Random(Box<dyn rand::RngCore>),
    ...
}
```
But it is probably better to wait for someone who actually needs something like this to make a proposal, rather than blindly implementing something.

mankinskin:
I am unsure now, should I implement the `FindOrDefaultResult` too, or do we leave it as it is?

phimuemue:
> I am unsure now, should I implement the `FindOrDefaultResult` too, or do we leave it as it is?

I'd leave it as it is and wait for @jswrenn 's opinion. Tbh, I second your idea of not blindly implementing abstractions that may not even be needed.

jswrenn (approved):
Happy to merge this as-is. We can revisit improvements in the future, as needed.

bors r+

Issue #537, opened 2021-04-21
Feature Request: `zip_pad_with_default`


abrassel:
When trying to zip over two iterators of unequal lengths, `zip_longest` is very useful for the case in which one doesn't exist.  
When trying to do zip longest, but the underlying items implement `Default`, it could be useful to have a version of `zip_longest` that pads the shorter iterator with `Item::default()`.

scottmcm:
I don't think a new iterator adapter is the best choice here, since that's a relatively heavy and non-orthogonal addition.

The item type for `zip_longest` is `EitherOrBoth`, though, so maybe add some methods on there to make things easier?

As a sketch,
```
impl<T:Default> impl EitherOrBoth<T> {
    fn or_default(self) -> (T, T) { ... }
}
```

Then `.zip_pad_with_default()` could be done as just `.zip_longest().map(|x| x.or_default())`.

(Dunno what the best name for the method would be.  I'm inspired here by https://docs.rs/itertools/0.9.0/itertools/enum.MinMaxResult.html#method.into_option)

abrassel:
That looks great to me!  I've not contributed to an open source project before - at this point do I submit a pull request?

phimuemue:
> That looks great to me!  I've not contributed to an open source project before - at this point do I submit a pull request?

I think you could give it a try.

jswrenn:
Please do! I'd be happy to merge this.

abrassel:
@jswrenn Done!

PR #538, opened 2021-04-29
Added the `or_default`  method to `EitherOrBoth`

abrassel:
This commit is motivated by the context explained in issue #537 

Allows for doing things like padding out an iterator with a default value while the other one is consumed.

### DIFF
--- a/src/either_or_both.rs
+++ b/src/either_or_both.rs
@@ -163,6 +163,21 @@ impl<A, B> EitherOrBoth<A, B> {
             Right(b) | Both(_, b) => f(b),
         }
     }
+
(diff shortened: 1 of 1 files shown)


phimuemue:
Looks good to me.

Maybe we can collect all the `impl EitherOrBoth` under one single `impl` block afterwards.

abrassel:
> Looks good to me.
> 
> Maybe we can collect all the `impl EitherOrBoth` under one single `impl` block afterwards.

Thanks!  So how would we do that if there are different bounds on the generics in different `impl` blocks?

jswrenn:
Put the bounds on the methods, rather than the block.

abrassel:
Oh that makes sense. I'll add another commit which does that.

PR #539, opened 2021-05-02
Refine intra-doc links

tranzystorek-io:
A few changes to make documenting code with links simpler and more uniform.

Highlights:

- Make some common stdlib traits into links, e.g. `IntoIterator`
- Add links to docs for free functions
- Use module paths uniformly in links (instead of html paths)
- Link to `slice::sort*` methods in docs for sorting functions

### DIFF
--- a/src/combinations.rs
+++ b/src/combinations.rs
@@ -47,9 +47,7 @@ impl<I: Iterator> Combinations<I> {
     pub fn k(&self) -> usize { self.indices.len() }
 
     /// Returns the (current) length of the pool from which combination elements are
-    /// selected. This value can change between invocations of [`next`].
-    ///
-    /// [`next`]: #method.next
+    /// selected. This value can change between invocations of [`next`](Combinations::next).
     #[inline]
     pub fn n(&self) -> usize { self.pool.len() }
 
--- a/src/free.rs
+++ b/src/free.rs
@@ -223,9 +223,7 @@ pub fn join<I>(iterable: I, sep: &str) -> String
 
 /// Sort all iterator elements into a new iterator in ascending order.
 ///
-/// [`IntoIterator`] enabled version of [`iterable.sorted()`][1].
-///
-/// [1]: crate::Itertools::sorted
+/// [`IntoIterator`] enabled version of [`Itertools::sorted`].
 ///
 /// ```
 /// use itertools::sorted;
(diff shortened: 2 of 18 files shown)


src/combinations.rs
```diff
@@ -47,9 +47,7 @@ impl<I: Iterator> Combinations<I> {
     pub fn k(&self) -> usize { self.indices.len() }
 
     /// Returns the (current) length of the pool from which combination elements are
-    /// selected. This value can change between invocations of [`next`].
-    ///
-    /// [`next`]: #method.next
+    /// selected. This value can change between invocations of [`next`](Combinations::next).
```
phimuemue, line 50: Links to `Combinations::next` (not `Iterator::next`). Is this canonical?

src/free.rs
```diff
@@ -223,7 +223,7 @@ pub fn join<I>(iterable: I, sep: &str) -> String
 
 /// Sort all iterator elements into a new iterator in ascending order.
 ///
-/// `IntoIterator` enabled version of [`iterable.sorted()`][1].
+/// [`IntoIterator`] enabled version of [`Itertools::sorted`].
 ///
 /// [1]: crate::Itertools::sorted
```
phimuemue, line 228: Is this still needed?

phimuemue (commented):
Thanks for the cleanups - better docs are always nice! I skimmed through the commit and have some questions. (Not sure that I'll be able to re-visit this, though.)

Do you think you could split this up into smaller commits (e.g. one for `[IntoIterator]` corrections, basically along the lines of your (possibly refinable) bullet list)? I know it puts more burden on your shoulders, but the constant context switching seems to make review quite tedious.

tranzystorek-io:
> Do you think you could split this up into smaller commits

Sure, shouldn't be a problem 😄

src/free.rs
tranzystorek-io: No, missed it, thanks!

src/combinations.rs
tranzystorek-io: I'm not sure, my first thought was that it should point to the `itertools` documentation instead of sending to the general trait method.

tranzystorek-io:
As requested, I split the change into smaller commits that should be more palatable 😄

jswrenn (approved):
bors r+

PR #540, opened 2021-05-02
either_or_both: Fix inverted Left/Right in docstrings

cgwalters:
I stared at this for a bit, confused before deciding the docs
must be wrong.

### DIFF
--- a/src/either_or_both.rs
+++ b/src/either_or_both.rs
@@ -140,7 +140,7 @@ impl<A, B> EitherOrBoth<A, B> {
         }
     }
 
-    /// Apply the function `f` on the value `b` in `Right(b)` or `Both(a, _)` variants if it is
+    /// Apply the function `f` on the value `a` in `Left(a)` or `Both(a, _)` variants if it is
     /// present.
     pub fn left_and_then<F, L>(self, f: F) -> EitherOrBoth<L, B>
     where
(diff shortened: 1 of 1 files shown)


src/either_or_both.rs
```diff
@@ -140,7 +140,7 @@ impl<A, B> EitherOrBoth<A, B> {
         }
     }
 
-    /// Apply the function `f` on the value `b` in `Right(b)` or `Both(a, _)` variants if it is
+    /// Apply the function `f` on the value `b` in `Left(b)` or `Both(a, _)` variants if it is
```
phimuemue, line 143: Shouldn't this be ".. or `Both(b, _)`? And maybe it should use the placeholder `a`?

src/either_or_both.rs
```diff
@@ -153,7 +153,7 @@ impl<A, B> EitherOrBoth<A, B> {
     }
 
     /// Apply the function `f` on the value `a`
-    /// in `Left(a)` or `Both(a, _)` variants if it is present.
+    /// in `Right(a)` or `Both(a, _)` variants if it is present.
```
phimuemue, line 156: Shouldn't this use placeholder `b` and be `Both(_, b)`?

phimuemue (commented):
Thank you for spotting this. Maybe we should try to bring all the doc in line, using `a` for `Left`, `b` for `Right`?

src/either_or_both.rs
cgwalters: Yeah I think we're `Both` `Right`, that'd `b` `a` better fix.  Updated!

phimuemue:
Thank you.

bors r+

PR #548, opened 2021-05-25
Add FusedIterator for which it is commented fused.

aobatact:
`Interleave`, `IntersperseWith`, and `ZipLongest` are said to be fused in doc comment, but it wasn't marked as `FusedIterator`.


jswrenn (approved):
bors r+

PR #552, opened 2021-06-17
Fix subtraction overflow in DuplicatesBy size_hint

arpankapoor:
Sample that panics with `attempt to subtract with overflow` in a `debug` build:

```rust
use itertools::Itertools;

fn main() {
    println!("{:?}", vec!["a", "b", "c", "c"].iter().duplicates().collect_vec());
}
```

### DIFF
--- a/src/duplicates_impl.rs
+++ b/src/duplicates_impl.rs
@@ -89,7 +89,7 @@ mod private {
             // far), plus (hi - pending) / 2 pairs of never seen before items.
             let hi = hi.map(|hi| {
                 let max_pending = std::cmp::min(self.meta.pending, hi);
-                let max_new = hi.checked_sub(self.meta.pending).unwrap_or(0) / 2;
+                let max_new = hi.saturating_sub(self.meta.pending) / 2;
                 max_pending + max_new
             });
             // The lower bound is always 0 since we might only get unique items from now on
--- a/src/duplicates_impl.rs
+++ b/src/duplicates_impl.rs
@@ -89,7 +89,7 @@ mod private {
             // far), plus (hi - pending) / 2 pairs of never seen before items.
             let hi = hi.map(|hi| {
                 let max_pending = std::cmp::min(self.meta.pending, hi);
-                let max_new = std::cmp::max(hi - self.meta.pending, 0) / 2;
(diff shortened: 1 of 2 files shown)


phimuemue:
Thanks for this. Yes, `cmp::max(a-b, 0)` seems strange on `usize`. I do not have much time right now, so just two questions:

* Do you know if there are other occurences of this pattern in the code? (A "no, I do not know" is fine, we should just look ourselves, then.)
* Should we add tests for this case?

arpankapoor:
- I haven't seen this pattern repeated elsewhere on a cursory glance.
- Added a test similar to the above example. Let me know if that seems sufficient.

I've also changed the `checked_sub` call to `saturating_sub` as per [clippy's suggestion](https://rust-lang.github.io/rust-clippy/master/index.html#manual_saturating_arithmetic).

src/duplicates_impl.rs
```diff
@@ -89,7 +89,7 @@ mod private {
             // far), plus (hi - pending) / 2 pairs of never seen before items.
             let hi = hi.map(|hi| {
                 let max_pending = std::cmp::min(self.meta.pending, hi);
-                let max_new = std::cmp::max(hi - self.meta.pending, 0) / 2;
+                let max_new = hi.saturating_sub(self.meta.pending) / 2;
                 max_pending + max_new
```
phimuemue, line 93: Hi there, one thing that bothers me is that `cmp::min` and `saturating_sub` actually answer the same question twice. (Namely: is `hi <= self.meta.pending`?)

I think it would be clearer to do this:

```
if hi <= self.meta.pending {
    // fewer or equally many iter-remaining elements than pending elements
    // => at most, each iter-remaining element is matched
    hi
} else {
    // fewer pending elements than iter-remaining elements
    // => at most:
    //    * each pending element is matched
    //    * the other iter-remaining elements come in pairs
    self.meta.pending + (hi-self.meta.pending)/2
}

```

arpankapoor: updated as per your suggestion

phimuemue:
Thank you.

bors r+

phimuemue:
Merged as this improves the current `size_hint`. Maybe we should use our `size_hint::add` et al instead of plain `+` and others.

PR #555, opened 2021-06-22
Add intersperse and intersperse_with free functions

zertosh:
Aside from automatic parameter conversion using `IntoIterator`, having
`intersperse` and `intersperse_with` as free functions makes it easier for
downstream code to avoid the `unstable_name_collisions` lint because of
lib core's own similarly named `Iterator` methods.
https://github.com/rust-lang/rust/issues/48919

### DIFF
--- a/src/free.rs
+++ b/src/free.rs
@@ -14,8 +14,8 @@ use alloc::{
     string::String,
 };
 
-#[cfg(feature = "use_alloc")]
 use crate::Itertools;
+use crate::intersperse::{Intersperse, IntersperseWith};
 
 pub use crate::adaptors::{
     interleave,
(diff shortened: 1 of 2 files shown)


phimuemue:
Hi there, I agree: As long as we have `intersperse` in `Itertools` and we offer free functions, we should possibly include it in the free functions, too. However, there seems to be a problem (see CI).

zertosh:
@phimuemue, fixed! Sorry, I didn't get CI signal when I first put up the PR.

src/free.rs
```diff
@@ -14,8 +14,8 @@ use alloc::{
     string::String,
 };
 
-#[cfg(feature = "use_alloc")]
```
phimuemue, line 17: I'm no expert on `cfg`-stuff. Could you elaborate why this was removed?

zertosh: `intersperse` does not need `use_alloc` (works in `no_std` mode). This `#[cfg(feature = "use_alloc")]` was likely added here to avoid an `unused_imports` warning during `cargo check --no-default-features`  – since everything else in `free.rs` that depends on `Itertools` requires `use_alloc`. Since now `intersperse` is here, `Itertools` needs to be in scope when no features are set.

src/free.rs
```diff
@@ -35,6 +35,41 @@ pub use crate::merge_join::merge_join_by;
...
+    where I: IntoIterator,
+          <I as IntoIterator>::Item: Clone
+{
+    Itertools::intersperse(iterable.into_iter(), element)
+}
+
+/// Iterate `iterable` with a particular value created by a function inserted
+/// between each element.
+///
+/// [`IntoIterator`] enabled version of [`Iterator::intersperse`].
```
phimuemue, line 57: Documentation should say `intersperse_with`, right?

src/free.rs
```diff
@@ -35,6 +35,41 @@ pub use crate::merge_join::merge_join_by;
...
+/// ```
+/// use itertools::intersperse;
+///
+/// itertools::assert_equal(intersperse((0..3), 8), vec![0, 8, 1, 8, 2]);
+/// ```
+pub fn intersperse<I>(iterable: I, element: I::Item) -> Intersperse<I::IntoIter>
+    where I: IntoIterator,
+          <I as IntoIterator>::Item: Clone
+{
+    Itertools::intersperse(iterable.into_iter(), element)
```
phimuemue, line 51: I think we canonically use `iterable.into_iter().intersperse(element)`. Could you adapt this?

zertosh: I think this will eventually break because of https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.intersperse . This is actually the motivation for providing the free functions - so there's no conflict when the core `intersperse` and `intersperse_with` are stable.

phimuemue:
Thank you.

bors r+

PR #558, opened 2021-07-08
Compare `into_group_map_by`  with `into_grouping_map_by` in its doc (fix typo)

duchainer:
It was confusing to have the doc compare the function with itself, so it seems to be referring to the `into_grouping_map_by`  which doesn't include a key but only the values.

Make me know if I corrected wrong, or if it was good from the start. ^^

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2791,7 +2791,7 @@ pub trait Itertools : Iterator {
 
     /// Return an `Iterator` on a `HashMap`. Keys mapped to `Vec`s of values. The key is specified
     /// in the closure.
-    /// Different to `into_group_map_by` because the key is still present. It is also more general.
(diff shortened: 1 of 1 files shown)


phimuemue:
Hi, thanks for spotting this. Unfortunately, I think the documentation is wonky in a broader sense there, as it e.g. does not return an `Iterator` on a `HashMap`, but a `HashMap` directly.

If we touch this, we should probably review all the docs for `into_group_map[_by]` and `into_grouping_map[_by]`, bring the documentation in line, and clearly work out the differences.

phimuemue:
Closed as per https://github.com/rust-itertools/itertools/commit/599ae8cced7ede91c85a0e507808e8dcbf1acd27.

Issue #547, opened 2021-05-25
Const generics iterator candidates


conradludgate:
Discussed in #546 where I've written an alternative to `tuple_combinations`, `array_combinations` which is an iterator that outputs `[T; N]`. This feature requires const generics which were introduced in Rust 1.51. It was suggested that we can bump the MSRV in a major release (ie itertools 0.11) but it would be sensible to add more features along side.

This crate uses a lot of tuples where arrays make more sense. Tuples are good when you have multiple types (eg cartesian product or zip) but when you have a collection of the same type, a slice/vec/array makes more sense.

Here's a list of functions that seem to qualify using arrays:
* circular_tuple_windows -> circular_array_windows
* tuples -> arrays
* collect_tuple -> collect_array
* next_tuple -> next_array

These all work on a single iterator over the single `I::Item` type, and since these use tuples, the number of elements is known at compile time, making them perfect for using arrays instead of tuples.

scottmcm:
Note that things like `array_windows` for iterators are under discussion for iterators in std, see conversations like https://rust-lang.zulipchat.com/#narrow/stream/219381-t-libs/topic/Iterator.3A.3A.7Bpairwise.2C.20windows.7D/near/224587771

As I understand it `std` didn't want the tuple ones because of the extra trait complications in the interface, but with const generics available the array ones might be able to live there instead of here.

Plecra:
Could this be added and gated on rust 1.51? That'd maintain backwards compatibility without any problem

Philippe-Cholet:
@Plecra We use 1.43.1 at the moment but will move to 1.51 at some point and use const generics, no doubt.
I don't think there is a discussion about when we will do it but I suppose after #755 (well underway) and #844 (not started).

TonalidadeHidrica:
Now that #959 is merged, can we also start to merge https://github.com/rust-itertools/itertools/labels/const-generics PRs?  I really want to see `array_windows` and many other possible array-based features which simplify my code.

jswrenn:
Yes! I'm going to try to get some of these merged in the coming weeks.

johan-overbye:
> I'm going to try to get some of these merged in the coming weeks.

Does that include `circular_array_windows`? Been high on my wishlist for a long time. 😁

On another note, wouldn't the proposed `next_array` be obsoleted by the nightly std [`array_chunks`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.array_chunks)?

Thanks!

PR #560, opened 2021-07-15
Add `next_array` and `collect_array`

orlp:
With this pull request I add two new functions to the `Itertools` trait:

```rust
fn next_array<T, const N: usize>(&mut self) -> Option<[T; N]>
where Self: Sized + Iterator<Item = T>;

fn collect_array<T, const N: usize>(mut self) -> Option<[T; N]>
where Self: Sized + Iterator<Item = T>;
```

These behave exactly like `next_tuple` and `collect_tuple`, however they return arrays instead. Since these functions require `min_const_generics`, I added a tiny build script that checks if Rust's version is 1.51 or higher, and if yes to set the `has_min_const_generics` config variable. This means that `Itertools` does not suddenly require 1.51 or higher, only these two functions do.  

In order to facilitate this I did have to bump the minimum required Rust version to 1.34 from the (documented) 1.32, since Rust 1.32 and 1.33 have trouble parsing the file even if stuff is conditionally compiled. However, this should not result in any (new) breakage, because `Itertools` actually already requires Rust 1.34 for 9+ months, since https://github.com/rust-itertools/itertools/commit/83c0f046c077a71185042e020a04262f388a5157 uses `saturating_pow` which wasn't stabilized until 1.34.

---

As for rationale, I think these functions are useful, especially for pattern matching and parsing. I don't think there's a high probability they get added to the standard library either, so that's why I directly make a pull request here. When/if `TryFromIterator` stabilizes we can simplify the implementation, but even then I believe these functions remain a good addition similarly how `collect_vec` is nice to have despite `.collect::<Vec<_>>` existing.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -209,6 +209,7 @@ mod merge_join;
 mod minmax;
 #[cfg(feature = "use_alloc")]
 mod multipeek_impl;
+mod next_array;
 mod pad_tail;
 #[cfg(feature = "use_alloc")]
 mod peek_nth;
@@ -1968,6 +1969,50 @@ pub trait Itertools: Iterator {
     }
 
     // non-adaptor methods
+    /// Advances the iterator and returns the next items grouped in an array of
+    /// a specific size.
+    ///
+    /// If there are enough elements to be grouped in an array, then the array
+    /// is returned inside `Some`, otherwise `None` is returned.
+    ///
+    /// ```
+    /// use itertools::Itertools;
+    ///
+    /// let mut iter = 1..5;
+    ///
+    /// assert_eq!(Some([1, 2]), iter.next_array());
+    /// ```
+    fn next_array<const N: usize>(&mut self) -> Option<[Self::Item; N]>
+    where
+        Self: Sized,
+    {
+        next_array::next_array(self)
+    }
+
+    /// Collects all items from the iterator into an array of a specific size.
+    ///
+    /// If the number of elements inside the iterator is **exactly** equal to
+    /// the array size, then the array is returned inside `Some`, otherwise
+    /// `None` is returned.
+    ///
+    /// ```
+    /// use itertools::Itertools;
+    ///
+    /// let iter = 1..3;
+    ///
+    /// if let Some([x, y]) = iter.collect_array() {
+    ///     assert_eq!([x, y], [1, 2])
+    /// } else {
+    ///     panic!("Expected two elements")
+    /// }
+    /// ```
+    fn collect_array<const N: usize>(mut self) -> Option<[Self::Item; N]>
+    where
+        Self: Sized,
+    {
+        self.next_array().filter(|_| self.next().is_none())
+    }
+
     /// Advances the iterator and returns the next items grouped in a tuple of
     /// a specific size (up to 12).
     ///
--- a/src/next_array.rs
+++ b/src/next_array.rs
@@ -1,0 +1,269 @@
+use core::mem::{self, MaybeUninit};
+
+/// An array of at most `N` elements.
+struct ArrayBuilder<T, const N: usize> {
+    /// The (possibly uninitialized) elements of the `ArrayBuilder`.
+    ///
+    /// # Safety
+    ///
+    /// The elements of `arr[..len]` are valid `T`s.
+    arr: [MaybeUninit<T>; N],
+
+    /// The number of leading elements of `arr` that are valid `T`s, len <= N.
+    len: usize,
+}
+
+impl<T, const N: usize> ArrayBuilder<T, N> {
+    /// Initializes a new, empty `ArrayBuilder`.
+    pub fn new() -> Self {
+        // SAFETY: The safety invariant of `arr` trivially holds for `len = 0`.
+        Self {
+            arr: [(); N].map(|_| MaybeUninit::uninit()),
+            len: 0,
+        }
+    }
+
+    /// Pushes `value` onto the end of the array.
+    ///
+    /// # Panics
+    ///
+    /// This panics if `self.len >= N`.
+    #[inline(always)]
+    pub fn push(&mut self, value: T) {
+        // PANICS: This will panic if `self.len >= N`.
+        let place = &mut self.arr[self.len];
+        // SAFETY: The safety invariant of `self.arr` applies to elements at
+        // indices `0..self.len` — not to the element at `self.len`. Writing to
+        // the element at index `self.len` therefore does not violate the safety
+        // invariant of `self.arr`. Even if this line panics, we have not
+        // created any intermediate invalid state.
+        *place = MaybeUninit::new(value);
+        // Lemma: `self.len < N`. By invariant, `self.len <= N`. Above, we index
+        // into `self.arr`, which has size `N`, at index `self.len`. If `self.len == N`
+        // at that point, that index would be out-of-bounds, and the index
+        // operation would panic. Thus, `self.len != N`, and since `self.len <= N`,
+        // that means that `self.len < N`.
+        //
+        // PANICS: Since `self.len < N`, and since `N <= usize::MAX`,
+        // `self.len + 1 <= usize::MAX`, and so `self.len += 1` will not
+        // overflow. Overflow is the only panic condition of `+=`.
+        //
+        // SAFETY:
+        // - We are required to uphold the invariant that `self.len <= N`.
+        //   Since, by the preceding lemma, `self.len < N` at this point in the
+        //   code, `self.len += 1` results in `self.len <= N`.
+        // - We are required to uphold the invariant that `self.arr[..self.len]`
+        //   are valid instances of `T`. Since this invariant already held when
+        //   this method was called, and since we only increment `self.len`
+        //   by 1 here, we only need to prove that the element at
+        //   `self.arr[self.len]` (using the value of `self.len` before incrementing)
+        //   is valid. Above, we construct `place` to point to `self.arr[self.len]`,
+        //   and then initialize `*place` to `MaybeUninit::new(value)`, which is
+        //   a valid `T` by construction.
+        self.len += 1;
+    }
+
+    /// Consumes the elements in the `ArrayBuilder` and returns them as an array
+    /// `[T; N]`.
+    ///
+    /// If `self.len() < N`, this returns `None`.
+    pub fn take(&mut self) -> Option<[T; N]> {
+        if self.len == N {
+            // SAFETY: Decreasing the value of `self.len` cannot violate the
+            // safety invariant on `self.arr`.
+            self.len = 0;
+
+            // SAFETY: Since `self.len` is 0, `self.arr` may safely contain
+            // uninitialized elements.
+            let arr = mem::replace(&mut self.arr, [(); N].map(|_| MaybeUninit::uninit()));
+
+            Some(arr.map(|v| {
+                // SAFETY: We know that all elements of `arr` are valid because
+                // we checked that `len == N`.
+                unsafe { v.assume_init() }
+            }))
+        } else {
+            None
+        }
+    }
+}
+
+impl<T, const N: usize> AsMut<[T]> for ArrayBuilder<T, N> {
+    fn as_mut(&mut self) -> &mut [T] {
+        let valid = &mut self.arr[..self.len];
+        // SAFETY: By invariant on `self.arr`, the elements of `self.arr` at
+        // indices `0..self.len` are in a valid state. Since `valid` references
+        // only these elements, the safety precondition of
+        // `slice_assume_init_mut` is satisfied.
+        unsafe { slice_assume_init_mut(valid) }
+    }
+}
+
+impl<T, const N: usize> Drop for ArrayBuilder<T, N> {
+    // We provide a non-trivial `Drop` impl, because the trivial impl would be a
+    // no-op; `MaybeUninit<T>` has no innate awareness of its own validity, and
+    // so it can only forget its contents. By leveraging the safety invariant of
+    // `self.arr`, we do know which elements of `self.arr` are valid, and can
+    // selectively run their destructors.
+    fn drop(&mut self) {
+        // SAFETY:
+        // - by invariant on `&mut [T]`, `self.as_mut()` is:
+        //   - valid for reads and writes
+        //   - properly aligned
+        //   - non-null
+        // - the dropped `T` are valid for dropping; they do not have any
+        //   additional library invariants that we've violated
+        // - no other pointers to `valid` exist (since we're in the context of
+        //   `drop`)
+        unsafe { core::ptr::drop_in_place(self.as_mut()) }
+    }
+}
+
+/// Assuming all the elements are initialized, get a mutable slice to them.
+///
+/// # Safety
+///
+/// The caller guarantees that the elements `T` referenced by `slice` are in a
+/// valid state.
+unsafe fn slice_assume_init_mut<T>(slice: &mut [MaybeUninit<T>]) -> &mut [T] {
+    // SAFETY: Casting `&mut [MaybeUninit<T>]` to `&mut [T]` is sound, because
+    // `MaybeUninit<T>` is guaranteed to have the same size, alignment and ABI
+    // as `T`, and because the caller has guaranteed that `slice` is in the
+    // valid state.
+    unsafe { &mut *(slice as *mut [MaybeUninit<T>] as *mut [T]) }
+}
+
+/// Equivalent to `it.next_array()`.
+pub(crate) fn next_array<I, const N: usize>(it: &mut I) -> Option<[I::Item; N]>
+where
+    I: Iterator,
+{
+    let mut builder = ArrayBuilder::new();
+    for _ in 0..N {
+        builder.push(it.next()?);
+    }
+    builder.take()
+}
+
+#[cfg(test)]
+mod test {
+    use super::ArrayBuilder;
+
+    #[test]
+    fn zero_len_take() {
+        let mut builder = ArrayBuilder::<(), 0>::new();
+        let taken = builder.take();
+        assert_eq!(taken, Some([(); 0]));
+    }
+
+    #[test]
+    #[should_panic]
+    fn zero_len_push() {
+        let mut builder = ArrayBuilder::<(), 0>::new();
+        builder.push(());
+    }
+
+    #[test]
+    fn push_4() {
+        let mut builder = ArrayBuilder::<(), 4>::new();
+        assert_eq!(builder.take(), None);
+
+        builder.push(());
+        assert_eq!(builder.take(), None);
+
+        builder.push(());
+        assert_eq!(builder.take(), None);
+
+        builder.push(());
+        assert_eq!(builder.take(), None);
+
+        builder.push(());
+        assert_eq!(builder.take(), Some([(); 4]));
+    }
+
+    #[test]
+    fn tracked_drop() {
+        use std::panic::{catch_unwind, AssertUnwindSafe};
+        use std::sync::atomic::{AtomicU16, Ordering};
+
+        static DROPPED: AtomicU16 = AtomicU16::new(0);
+
+        #[derive(Debug, PartialEq)]
+        struct TrackedDrop;
+
+        impl Drop for TrackedDrop {
+            fn drop(&mut self) {
+                DROPPED.fetch_add(1, Ordering::Relaxed);
+            }
+        }
+
+        {
+            let builder = ArrayBuilder::<TrackedDrop, 0>::new();
+            assert_eq!(DROPPED.load(Ordering::Relaxed), 0);
+            drop(builder);
+            assert_eq!(DROPPED.load(Ordering::Relaxed), 0);
+        }
+
+        {
+            let mut builder = ArrayBuilder::<TrackedDrop, 2>::new();
+            builder.push(TrackedDrop);
+            assert_eq!(builder.take(), None);
+            assert_eq!(DROPPED.load(Ordering::Relaxed), 0);
+            drop(builder);
+            assert_eq!(DROPPED.swap(0, Ordering::Relaxed), 1);
+        }
+
+        {
+            let mut builder = ArrayBuilder::<TrackedDrop, 2>::new();
+            builder.push(TrackedDrop);
+            builder.push(TrackedDrop);
+            assert!(matches!(builder.take(), Some(_)));
+            assert_eq!(DROPPED.swap(0, Ordering::Relaxed), 2);
+            drop(builder);
+            assert_eq!(DROPPED.load(Ordering::Relaxed), 0);
+        }
+
+        {
+            let mut builder = ArrayBuilder::<TrackedDrop, 2>::new();
+
+            builder.push(TrackedDrop);
+            builder.push(TrackedDrop);
+
+            assert!(catch_unwind(AssertUnwindSafe(|| {
+                builder.push(TrackedDrop);
+            }))
+            .is_err());
+
+            assert_eq!(DROPPED.load(Ordering::Relaxed), 1);
+
+            drop(builder);
+
+            assert_eq!(DROPPED.swap(0, Ordering::Relaxed), 3);
+        }
+
+        {
+            let mut builder = ArrayBuilder::<TrackedDrop, 2>::new();
+
+            builder.push(TrackedDrop);
+            builder.push(TrackedDrop);
+
+            assert!(catch_unwind(AssertUnwindSafe(|| {
+                builder.push(TrackedDrop);
+            }))
+            .is_err());
+
+            assert_eq!(DROPPED.load(Ordering::Relaxed), 1);
+
+            assert!(matches!(builder.take(), Some(_)));
+
+            assert_eq!(DROPPED.load(Ordering::Relaxed), 3);
+
+            builder.push(TrackedDrop);
+            builder.push(TrackedDrop);
+
+            assert!(matches!(builder.take(), Some(_)));
+
+            assert_eq!(DROPPED.swap(0, Ordering::Relaxed), 5);
+        }
+    }
+}
--- a/tests/test_core.rs
+++ b/tests/test_core.rs
@@ -372,3 +372,28 @@ fn product1() {
     assert_eq!(v[1..3].iter().cloned().product1::<i32>(), Some(2));
     assert_eq!(v[1..5].iter().cloned().product1::<i32>(), Some(24));
 }
+
+#[test]
+fn next_array() {
+    let v = [1, 2, 3, 4, 5];
+    let mut iter = v.iter();
+    assert_eq!(iter.next_array(), Some([]));
+    assert_eq!(iter.next_array().map(|[&x, &y]| [x, y]), Some([1, 2]));
+    assert_eq!(iter.next_array().map(|[&x, &y]| [x, y]), Some([3, 4]));
+    assert_eq!(iter.next_array::<2>(), None);
+}
+
+#[test]
+fn collect_array() {
+    let v = [1, 2];
+    let iter = v.iter().cloned();
+    assert_eq!(iter.collect_array(), Some([1, 2]));
+
+    let v = [1];
+    let iter = v.iter().cloned();
+    assert_eq!(iter.collect_array::<2>(), None);
+
+    let v = [1, 2, 3];
+    let iter = v.iter().cloned();
+    assert_eq!(iter.collect_array::<2>(), None);
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Added
 - Added `array_combinations` (#991)
 - Added `k_smallest_relaxed` and variants (#925)
+- Added `next_array` and `collect_array` (#560)
 - Implemented `DoubleEndedIterator` for `FilterOk` (#948)
 - Implemented `DoubleEndedIterator` for `FilterMapOk` (#950)
 


orlp:
A possible enhancement might be to return `A` where `A: FromArray<Self::Item, N>` instead, and adding the `FromArray` trait.

```rust
impl<T, const N: usize> FromArray<T, N> for [T; N] {
    fn from_array(array: [T; N]) -> Self {
        array
    }
}

impl<T, const N: usize> FromArray<T, N> for Box<[T; N]> { /* .. */ }
impl<'a, T: Clone, const N: usize> FromArray<T, N> for Cow<'a, [T; N]> { /* .. */ }
impl<T, const N: usize> FromArray<T, N> for Rc<[T; N]> { /* .. */ }
impl<T, const N: usize> FromArray<T, N> for Arc<[T; N]> { /* .. */ }
impl<T, const N: usize> FromArray<Option<T>, N> for Option<[T; N]> { /* .. */ }
impl<T, E, const N: usize> FromArray<Result<T, E>, N> for Result<[T; N], E> { /* .. */ }
```

In fact, I think this is highly useful because it allows things like

```rust
let ints = line.split_whitespace().map(|n| n.parse());
if let Ok([x, y, z]) = ints.collect_array() {
    ...
}
```

orlp:
So I have a working implementation of the above here: <https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=c009d90c3c1921d86428fa33e305559a>.

It makes this compile:

```rust
fn main() {
    let line = "32 -12 24";
    let nums = line.split_whitespace().map(|n| n.parse::<i32>());
    if let Some(Ok([x, y, z])) = nums.collect_array() {
        println!("x: {} y: {} z: {}", x, y, z);
    }
}
```

It would change the interface to:

```rust
trait ArrayCollectible<T>: Sized {
    fn array_from_iter<I: IntoIterator<Item = T>>(iterable: I) -> Option<Self>;
}

trait Itertools: Iterator {
    fn collect_array<A>(self) -> Option<A>
    where
        Self: Sized,
        A: ArrayCollectible<Self::Item>;
}
```

where
 - `ArrayCollectible<T>` is implemented for `[T; N]`, `Box<[T; N]>`, `Cow<'a, [T; N]>`, `Rc<[T; N]>`, `Arc<[T; N]>`;
 - `ArrayCollectible<Option<T>>` is implemented for `Option<[T; N]>`;
 - `ArrayCollectible<Result<T, E>>` is implemented for `Result<[T; N], E>`.

build.rs
```diff
@@ -0,0 +1,13 @@
...
+    let is_at_least_1_51 = version_check::is_min_version("1.51.0").unwrap_or(false);
+    
+    if !is_at_least_1_34 && !is_nightly {
+        println!("cargo:warning=itertools requires rustc => 1.34.0");
+    }
+
+    if is_at_least_1_51 || is_nightly {
+        println!("cargo:rustc-cfg=has_min_const_generics");
+    }
+}
```
phimuemue, line 13: Usually, I like the idea of having everything automated, but I am  not sure if we should go with a `build.rs` and an additional dependency. My first idea was to use a feature flag (that would probably be off by default) that the user can enable if desired.

src/lib.rs
```diff
@@ -42,7 +42,7 @@
 //!
 //! ## Rust Version
 //!
-//! This version of itertools requires Rust 1.32 or later.
+//! This version of itertools requires Rust 1.34 or later.
```
phimuemue, line 45: If your assessment is correct, we could possibly increment the minimum rust version in a separate commit?

src/lib.rs
```diff
@@ -1721,13 +1776,7 @@ pub trait Itertools : Iterator {
...
               T: traits::HomogeneousTuple
     {
-        match self.next_tuple() {
-            elt @ Some(_) => match self.next() {
-                Some(_) => None,
-                None => elt,
-            },
-            _ => None
-        }
+        self.next_tuple().filter(|_| self.next().is_none())
```
phimuemue, line 1779: Is this really relevant to this PR? If not, could we separate it into another PR?

src/next_array.rs
```diff
@@ -0,0 +1,80 @@
+use core::mem::MaybeUninit;
```
phimuemue, line 1: I think there was some discussion about building arrays:

* https://github.com/rust-itertools/itertools/pull/549#issuecomment-848235283
* https://github.com/rust-itertools/itertools/pull/546
* https://github.com/rust-itertools/itertools/issues/547

phimuemue (commented):
Hi there! Thanks for this. I particularly like that you thought about a way of enabling `const`-generic stuff without raising the minimum required rust version (even if I would imagine something else due to having an aversion against depending on other crates too much).

There has been some discussion recently about basically supporting not only tuples, but also arrays. I just want to make sure that we do not loose input from these discussions when actually settling with your solution:
* https://github.com/rust-itertools/itertools/pull/549
* https://github.com/rust-itertools/itertools/pull/546
* https://github.com/rust-itertools/itertools/issues/547

On top of that, I think there are some changes in there that are not directly related to this issue. If you'd like to have them merged, could you possibly factor them out into separate PRs/commits?

src/lib.rs
orlp: It was mostly to be consistent with the other implementation. As it's just a stylistic change I don't think it's worth a pull request by itself to be honest.

src/lib.rs
orlp: Do you mean a pull request? It's already a separate commit.

build.rs
orlp: > My first idea was to use a feature flag (that would probably be off by default) that the user can enable if desired.

I think feature flags to enable things that are already available in the latest stable Rust and have no further compile time or dependency drawbacks makes no sense.

1. You could end up with dozens of feature flags for minor features, or hold back progress because a feature that would otherwise be merged would now be too minor to accept due to introducing a new feature flag.
2. You end up with useless features that you can't remove without a breaking change as your MSRV goes up.
3. It's un-ergonomic as it adds an extra step for the user.

These drawbacks while it could be done completely and correctly automatically are unacceptable in my opinion. If you are hesitant regarding the `version-check` dependency, I'd just like to note that it's tiny, has no further downstream dependencies, and is already relied on by crates such as `time`, `nom`, `rocket`, `fd-find` among others.

orlp:
@phimuemue Any update on this?

phimuemue:
> @phimuemue Any update on this?

I appreciate your effort, but unfortunately nothing substantial from my side: I changed my mind regarding `version-check` (so we could use it as a `dev-dependency`), but I do not have enough time right now to review and merge PRs that ample.

orlp:
@phimuemue Just for posterity's sake, `version-check` would be a `build-dependency`, not `dev-dependency`.

orlp:
@phimuemue Just checking in what the status is, I feel very strongly about the usefulness of `collect_array`. I miss it very often in `itertools`.

scottmcm:
Note that if you want `collect_array`, you can use https://lib.rs/crates/arrayvec, as the usual way to collect into an array.

I'll also mention [`Iterator::next_chunk`](https://doc.rust-lang.org/nightly/std/iter/trait.Iterator.html#method.next_chunk) (https://github.com/rust-lang/rust/issues/98326) as a nightly API that'll be `next_array`.

Expurple:
This is a very useful feature. Today there was a [thread](https://www.reddit.com/r/rust/comments/1bovowu/collecting_arrays/) on Reddit where the author basically asks if there's a crate that provides `collect_array()`. IMO, `itertools` should be the crate to do it

Philippe-Cholet:
@Expurple
I sure would like to do use const generics and `collect_array` is one of them.
Our MSRV is quite old (1.43.1 currently) while min-const-generics is 1.51 but I do not think it's the main blocker.
The fact is that there is not much available in recent stable Rust yet which is sad. `Iterator::next_chunk` and `core::array::try_from_fn` would be nice to have.
Plus, we currently don't really use `unsafe` ourselves (only in `EitherOrBoth::insert*` with obvious unfaillable patterns). I guess we prefer that the std does the heavy work.

phimuemue:
I sometimes think about adding `arrayvec` as a dependency - and falling back to `std` as soon it's possible. I think it might also solve some other issues (e.g. [`ExactlyOneError` having a manual two-element-arrayvec](https://docs.rs/itertools/latest/src/itertools/exactly_one_err.rs.html#20-26)). Would require Rust 1.51.

Another option I just saw: Crates can offer "nightly-only experimental API" (see https://docs.rs/arrayvec/latest/arrayvec/struct.ArrayVec.html#method.first_chunk for an example) - maybe this would help _some_ users.

I personally would lean towards `arrayvec`. @jswrenn @Philippe-Cholet Opinions?

Philippe-Cholet:
@phimuemue 

> Another option I just saw: Crates can offer "nightly-only experimental API" (see https://docs.rs/arrayvec/latest/arrayvec/struct.ArrayVec.html#method.first_chunk for an example) - maybe this would help _some_ users.

`ArrayVec<T, CAP>` implements `Deref<Target = [T]>` so (nightly-available) slice methods are directly accessible, that seems to be it.

> I sometimes think about adding `arrayvec` as a dependency - and falling back to `std` as soon it's possible. I think it might also solve some other issues (e.g. [`ExactlyOneError` having a manual two-element-arrayvec](https://docs.rs/itertools/latest/src/itertools/exactly_one_err.rs.html#20-26)). Would require Rust 1.51.

I'm definitely not opposed to the idea but the `ExactlyOneError` use case is quite small.
I did not give thoughts before, do you have other examples in mind? (with private usage, in order to fall back to `std` ASAP).

orlp:
@jswrenn Please don't forget that we are discussing this on a PR that already has a working implementation without adding dependencies...

src/next_array.rs
```diff
@@ -0,0 +1,80 @@
...
+/// Equivalent to `it.next_array()`.
+pub fn next_array<I, T, const N: usize>(it: &mut I) -> Option<[T; N]>
+where
+    I: Iterator<Item = T>,
+{
+    let mut builder = ArrayBuilder::new();
+    for el in it.take(N) {
+        unsafe {
+            // SAFETY: the take(N) guarantees we never go out of bounds.
+            builder.push_unchecked(el);
```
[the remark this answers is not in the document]
orlp: That's a nasty one, I think you're right.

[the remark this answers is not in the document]
orlp: There should still be a `take` in there though, when using `try_for_each`.

jswrenn:
@orlp, thanks, I had forgotten that this was a PR and not an issue when I made my reply. Still, we're talking about adding some *extremely* subtle unsafe code to Itertools. I'd like us to take extreme care to avoid accidentally introducing UB.

A PR adding `unsafe` to itertools should:
- rigorously document the safety and panicking conditions of every `unsafe` function it introduces
- prove that every invocation of an unsafe function (even invocations occurring within other unsafe functions) satisfies the safety precondition of that invocation, with citations to official Rust documentation
- rigorously document why any potentially panicking function within an unsafe function does not create invalid state that would cause UB upon panicking unwinds
- intensively test its API with miri

If you can update this PR to do those things, I can see a path forward to merging it.

src/next_array.rs
```diff
@@ -0,0 +1,80 @@
+use core::mem::MaybeUninit;
+
+/// Helper struct to build up an array element by element.
+struct ArrayBuilder<T, const N: usize> {
+    arr: [MaybeUninit<T>; N],
+    i: usize
```
jswrenn, line 6: What is the safety invariant of `i` with relation to `arr`?

src/next_array.rs
```diff
@@ -0,0 +1,80 @@
...
+}
+
+impl<T, const N: usize> ArrayBuilder<T, N> {
+    pub fn new() -> Self {
+        Self { arr: maybe_uninit::uninit_array(), i: 0 }
+    }
+    
+    pub unsafe fn push_unchecked(&mut self, x: T) {
```
jswrenn, line 14: Needs a safety comment, in the format of:
```suggestion
    /// Does XYZ.
    ///
    /// # Safety
    /// 
    /// Callers promises that blah blah blah.
    ///
    /// # Panics
    ///
    /// This method does (or does not) panic.
    pub unsafe fn push_unchecked(&mut self, x: T) {
```

src/next_array.rs
```diff
@@ -0,0 +1,80 @@
...
+        debug_assert!(self.i < N);
+        *self.arr.get_unchecked_mut(self.i) = MaybeUninit::new(x);
```
jswrenn, line 16: Needs a safety comment in the form:
```suggestion
        // SAFETY: By contract on the caller, the safety condition on `get_unchecked_mut` that BLAH BLAH BLAH is satisfied.
        *self.arr.get_unchecked_mut(self.i) = MaybeUninit::new(x);
```

src/next_array.rs
```diff
@@ -0,0 +1,80 @@
...
+        self.i += 1;
+    }
+    
+    pub fn take(mut self) -> Option<[T; N]> {
+        if self.i == N {
+            unsafe {
+                // SAFETY: prevent double drop.
+                self.i = 0;
+                // SAFETY: [MaybeUninit<T>; N] and [T; N] have the same layout.
```
jswrenn, line 25: Could this safety comment site the standard library documentation? While it's true that these two types have the same size and alignment, it's not true that they have the same bit validity.

src/next_array.rs
```diff
@@ -0,0 +1,80 @@
...
+    
+    pub unsafe fn push_unchecked(&mut self, x: T) {
+        debug_assert!(self.i < N);
+        *self.arr.get_unchecked_mut(self.i) = MaybeUninit::new(x);
+        self.i += 1;
+    }
+    
+    pub fn take(mut self) -> Option<[T; N]> {
+        if self.i == N {
+            unsafe {
```
jswrenn, line 22: Could you scope this `unsafe { ... }` block down to just the `ptr::read`?

src/next_array.rs
```diff
@@ -0,0 +1,80 @@
...
+                // SAFETY: prevent double drop.
+                self.i = 0;
+                // SAFETY: [MaybeUninit<T>; N] and [T; N] have the same layout.
+                let init_arr_ptr = &self.arr as *const _ as *const [T; N];
+                Some(core::ptr::read(init_arr_ptr))
```
jswrenn, line 27: This needs a SAFETY comment citing what the preconditions of `ptr::read` are, and proving why they are satisfied.

src/next_array.rs
```diff
@@ -0,0 +1,80 @@
...
+/// Replacements for unstable core methods, copied from stdlib.
+mod maybe_uninit {
+    use core::mem::MaybeUninit;
+
+    pub fn uninit_array<T, const N: usize>() -> [MaybeUninit<T>; N] {
+        // SAFETY: an uninitialized `[MaybeUninit<_>; N]` is valid.
+        unsafe { MaybeUninit::<[MaybeUninit<T>; N]>::uninit().assume_init() }
+    }
+
+    pub unsafe fn assume_init_drop<T>(u: &mut MaybeUninit<T>) {
```
jswrenn, line 74: You don't need to replicate the entire stdlib doc comment here, but could you document the safety preconditions of `assume_init_drop`?

src/next_array.rs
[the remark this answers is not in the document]
jswrenn: Seconding @scottmcm's comment: For our MVP, does `push_unchecked` really need to be `_unchecked`?

build.rs
jswrenn: @Philippe-Cholet, @phimuemue, perhaps it's time we updated our MSRV to 1.51 (which is two years old at this point). 

While I don't mind the version-detection approach, I would like us to adopt it in tandem with changes to our CI that ensure we are testing on all detected versions. I'd also like to perhaps avoid taking the dependency on `rust_version`. This would all be a substantial change, and outside the scope of this PR.

My vote is that we increase our MSRV. We can aways *decrease* it in the future.

jswrenn (changes requested):
Thanks for this PR! I like the `ArrayBuilder` abstraction quite a bit. As I mentioned, this will need additional documentation and testing before it can be merged. See the recent safety comments in my other project, [`zerocopy`](https://github.com/google/zerocopy) for a sense of the ~paranoia~ rigor I'd like these safety comments to take.

orlp:
@jswrenn I will be busy the upcoming week but I'm willing to bring this up to standards after that. If before then you could decide on whether or not to bump the MSRV to 1.51 I could include that in the rewrite.

build.rs
Philippe-Cholet: @jswrenn I sure don't mind increasing the MSRV but I would suggest we release the 0.13.0 first, and then increase the MSRV in 0.14.0 to not require the build script (in which case **orlp** will have enough time to work on this).

jswrenn:
@orlp Think you might have time to revisit this soon? :-)

orlp:
@jswrenn Yes, I do. Have you reached a decision yet on the MSRV issue?

Philippe-Cholet:
I think it's time to set the MSRV to 1.51, I'm pretty sure @jswrenn and @phimuemue will agree.

jswrenn:
Absolutely. I'd even be fine going up to 1.55, which is nearly three years old. In my other crates, I've found that to be the lowest MSRV that users actually need.

Philippe-Cholet:
After quickly going though [release notes](https://github.com/rust-lang/rust/releases?page=4), I may have missed something but I only noted two things 1.55 has over 1.51 that I considered to be of potential use for us:
- [`<[T; N]>::map`](https://doc.rust-lang.org/stable/std/primitive.array.html#method.map)
- [`MaybeUninit::write`](https://doc.rust-lang.org/stable/std/mem/union.MaybeUninit.html#method.write-1)?

phimuemue:
Out of curiosity and slightly off-topic: What's a real reason to *not* update to stable Rust? Does it ever remove support for some platform or raise the system requirements dramatically? Or, put alternatively: Are there situations where someone *could* use cutting-edge `itertools` but *not* stable Rust?

jswrenn:
> Are there situations where someone _could_ use cutting-edge `itertools` but _not_ stable Rust?

Yes: Libraries that depend on itertools, but set a MSRV lower than stable. They are, of course, welcome to use an older, MSRV-compatible version of itertools, but we currently don't backport bugfixes to older versions.

> What's a real reason to not update to stable Rust? Does it ever remove support for some platform or raise the system requirements dramatically?

Rust occasionally does remove support for platforms; e.g.: https://blog.rust-lang.org/2022/08/01/Increasing-glibc-kernel-requirements.html

(The above post suggests that, conservatively, we could increase our MSRV to 1.63 without causing major problem for users. Maybe that's a good target MSRV for now?)

Philippe-Cholet:
After skimming all release notes up to 1.78.0, I noted this:
- 1.57.0: `Iterator::map_while`, `[T; N]::as_[mut_]slice`
- 1.60.0: `MaybeUninit::assume_init_[drop|read]`
- 1.62.0: `bool::then_some`
- 1.63.0: `array::from_fn`
- 1.65.0: `let else`
- 1.71.0: `From<[T; N]>` <--> `(T, ...)` (N in 1..=12)

I'm not particularly waiting anything in there (maybe `array::from_fn`). I'd prefer some nightly stuff like the `Try` trait and `array::try_from_fn`.

Unless we need [`<[T; N]>::map`](https://doc.rust-lang.org/stable/std/primitive.array.html#method.map) (1.55.0), I'm fine with 1.51 for quite some time.
Is there something currently stable that you are waiting for? Maybe something I missed?

orlp:
@Philippe-Cholet I feel [`<[T; N]>::map`](https://doc.rust-lang.org/stable/std/primitive.array.html#method.map) is rather fundamental and a strong reason to prefer 1.55 over 1.51. Similarly with `array::from_fn` for 1.63. The other stuff does not feel very relevant for this crate.

orlp:
I have force-pushed with a new cleaner implementation, based on 1.55 for `[T; N]::map`.

src/next_array.rs
```diff
@@ -0,0 +1,64 @@
...
+            // Take the array, resetting the length back to zero.
+            let arr = core::mem::replace(&mut self.arr, [(); N].map(|_| MaybeUninit::uninit()));
+            self.len = 0;
+
+            // SAFETY: we had len == N, so all elements in arr are initialized.
+            Some(unsafe { arr.map(|v| v.assume_init()) })
+        } else {
+            None
+        }
+    }
```
jswrenn, line 37: Slightly rearranging operations here allows us to preserve `ArrayBuilder`'s invariants and avoid temporarily violating them:
```suggestion
    /// Attempts to convert `self` into an array `[T; N]`.
    ///
    /// If `self.len() < N`, this returns `None`.
    pub fn take(&mut self) -> Option<[T; N]> {
        let len = self.len;
        if len == N {
            // SAFETY: Decreasing the value of `self.len` cannot violate its
            // safety invariant.
            self.len = 0;

            // SAFETY: Since `self.len` is 0, `self.arr` may safely contain
            // uninitialized elements.
            let arr = mem::replace(&mut self.arr, [(); N].map(|_| MaybeUninit::uninit()));

            let arr = arr.map(|elt| {
                // SAFETY: We know that all elements of `arr` are valid because
                // we checked that `len == N`.
                unsafe { elt.assume_init() }
            });

            Some(arr)
        } else {
            None
        }
    }
```

src/next_array.rs
```diff
@@ -0,0 +1,64 @@
...
+    fn drop(&mut self) {
+        unsafe {
+            // SAFETY: arr[..len] is initialized, so must be dropped.
+            // First we create a pointer to this initialized slice, then drop
+            // that slice in-place. The cast from *mut MaybeUninit<T> to *mut T
+            // is always sound by the layout guarantees of MaybeUninit.
+            let ptr_to_first: *mut MaybeUninit<T> = self.arr.as_mut_ptr();
+            let ptr_to_slice = ptr::slice_from_raw_parts_mut(ptr_to_first.cast::<T>(), self.len);
+            ptr::drop_in_place(ptr_to_slice);
+        }
```
jswrenn, line 50: Since our MSRV is now >= 1.60 we can leverage `MaybeUninit::assume_init_drop`. I know this is contrary to @scottmcm's suggestion [here](https://github.com/rust-itertools/itertools/pull/560#discussion_r1543287242), but it entirely avoids working with raw pointers.

```suggestion
        // LEMMA: The elements of `init` reference the valid elements of
        // `self.arr`.
        //
        // PROOF: `slice::split_at_mut(mid)` produces a pair of slices, the
        // first of which contains the elements at the indices `0..mid`. By
        // invariant on `self.arr`, the elements of `self.arr` at indicies
        // `0..self.len` are valid. Assuming that `slice::split_at_mut` is
        // correctly implemented, the slice `init` will only reference the
        // valid elements of `self.arr`.
        let (init, _) = self.arr.split_at_mut(self.len);
        // AXIOM: We assume that `slice::into_iter` and the subsequent
        // `for_each` are implemented correctly, and will yield each and every
        // element of `init`.
        init.into_iter().for_each(|elt| {
            // SAFETY: `elt` references a valid `T`, because by the above AXIOM,
            // `elt` is an element of `init`, and by the above LEMMA, the
            // elements of `init` reference the valid elements of `self.arr`.
            unsafe { elt.assume_init_drop() }
        });
```

src/next_array.rs
```diff
@@ -0,0 +1,64 @@
...
+
+/// Equivalent to `it.next_array()`.
+pub fn next_array<I, T, const N: usize>(it: &mut I) -> Option<[T; N]>
+where
+    I: Iterator<Item = T>,
+{
+    let mut builder = ArrayBuilder::new();
+    for _ in 0..N {
+        builder.push(it.next()?);
+    }
```
jswrenn, line 62: Now that these builder methods are safe, we can go back to calling `.take`, which can have performance benefits in some cases:
```suggestion
    it.take(N).for_each(|elt| builder.push(elt));
```

src/next_array.rs
```diff
@@ -0,0 +1,64 @@
+use core::mem::MaybeUninit;
```
jswrenn, line 1: This lets the later invocation of `core::mem::replace` be a little more brief:
```suggestion
use core::mem::{self, MaybeUninit};
```

tests/test_core.rs
```diff
@@ -372,3 +372,28 @@ fn product1() {
...
+    assert_eq!(iter.collect_array(), Some([1, 2]));
+
+    let v = [1];
+    let iter = v.iter().cloned();
+    assert_eq!(iter.collect_array::<_, 2>(), None);
+
+    let v = [1, 2, 3];
+    let iter = v.iter().cloned();
+    assert_eq!(iter.collect_array::<_, 2>(), None);
+}
```
jswrenn, line 399:
```suggestion
}

```

jswrenn (changes requested):
Thanks! This looks mostly good to me, though I've left some pedantic suggestions.

Can also you bump the MSRV in `Cargo.toml` and do a quick check for newly-resolvable clippy lints: https://github.com/rust-itertools/itertools/blob/ad5cc9645a5c3de2811eed1b2c6643f27f06764a/Cargo.toml#L18-L19

src/next_array.rs
```diff
@@ -0,0 +1,64 @@
...
+        }
+    }
+
+    pub fn push(&mut self, value: T) {
+        // We maintain the invariant here that arr[..len] is initialized.
+        // Indexing with self.len also ensures self.len < N, and thus <= N after
+        // the increment.
+        self.arr[self.len] = MaybeUninit::new(value);
+        self.len += 1;
+    }
```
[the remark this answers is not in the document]
orlp: I really feel this is overly verbose and ultimately reduces readability of the code. In my opinion, sometimes less is more.

src/next_array.rs
orlp: I will reorder those operations, sure.

src/next_array.rs
orlp: This actually has the wrong semantics compared to e.g. `Vec<T>`, as it will leak values if a `Drop` impl panics. `ptr::drop_in_place` will forcibly keep dropping elements in that scenario, causing an abort if another `Drop` panics.

src/next_array.rs
```diff
@@ -0,0 +1,64 @@
+use core::mem::MaybeUninit;
+use core::ptr;
+
+/// Helper struct to build up an array element by element.
+struct ArrayBuilder<T, const N: usize> {
+    arr: [MaybeUninit<T>; N], // Invariant: arr[..len] is initialized.
+    len: usize,               // Invariant: len <= N.
+}
```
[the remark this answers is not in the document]
jswrenn: Good point; I've updated the suggestion accordingly.

src/next_array.rs
orlp: I actually believe `take` here is more likely to negatively affect performance, either way the performance is bad due to the optimizer really not liking this code: https://github.com/rust-lang/rust/issues/126000.

orlp:
@jswrenn I actually wanted to suggest bumping the MSRV to whatever version you decide on in a separate PR we'd merge first rather than tacking it onto this one.

src/next_array.rs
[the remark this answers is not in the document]
jswrenn: I share that opinion, except in the case of documenting unsafe code. This PR is the first substantial addition of unsafe code to itertools, and I'd like us to set our standard for unsafe code very high. I also suspect that `ArrayBuilder` may turn out to be a major internal tool as we expand our support for const-generic adaptors.

src/next_array.rs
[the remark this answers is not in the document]
orlp: Also I must ask, will anyone... ever.. run `cargo doc --document-private-items` to look at the (safety) invariants of the private members of this private type instead of just looking at the code?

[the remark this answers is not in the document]
jswrenn: I use it quite often on the crates I maintain, and there's hardly a downside of writing an extra `/`.

src/next_array.rs
jswrenn: > Also, again I find this overly verbose, ultimately reducing readability of the code

I think it is likely that we have different preferences for how unsafe code is written and documented.

Abstractions like the one we're adding here are a notorious source of soundness issues in the Rust ecosystem (see the RUSTSEC advisories for [`smallvec`](https://rustsec.org/packages/smallvec.html), [`stackvector`](https://rustsec.org/packages/stackvector.html), and [`stack`](https://rustsec.org/packages/stack.html), for a sampler). Itertools has not yet had any soundness issues, and I'd like to keep it that way.

I am deeply reluctant to add *any* unsafe code to `itertools`. If we are to do so, I will require that it meets the standards I set for the other unsafe crates I maintain. The safety comments I'm proposing are not just suggestions; they're prerequisites of merging this PR. 

The argument about a semantic difference is interesting. Leaking is not a soundness issue, and we don't make any promises to match the drop behavior of `Vec` under such conditions. Nonetheless, I'll grant that minimizing leaking is preferable. I'll have to give this one more thought.

src/next_array.rs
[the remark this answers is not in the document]
orlp: @jswrenn Of course I think the standard should be very high. And for that reason I think the comments should be pristine, and not overly verbose blocks of text that just make it harder to review the code as correct. I sincerely mean here that less is more, the shorter unsafe comments are *better* in my opinion.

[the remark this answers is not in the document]
orlp: Think about it like this, lines of code that do things are signal, comments that do not help you understand the code are noise. In this example in your suggestion there are two lines of code, and nine lines of comments.

If you as a reader can reach the same understanding of the code with the same effort from a three-line comment as a nine-line comment (which I believe you can), then that means six lines were unnecessary and thus noise, giving a signal-to-noise ratio of 25%. I really prefer the signal-to-noise ratio to be as high as possible in `unsafe` code.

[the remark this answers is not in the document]
jswrenn: We'll have to agree to disagree; you are unlikely to convince me otherwise.

Making these comments has been an essential step in vetting this PR, and these comments will continue to aid future contributors and reviewers in making and evaluating changes to this code. 

For example, I only noticed the (temporary) [invariant violation issue in `ArrayBuilder::take`](https://github.com/rust-itertools/itertools/pull/560#discussion_r1627861703) because I attempted to write detailed safety comments for this PR. I also see that you've accepted my suggested reordering, but not my suggested comments. Now, the effect achieved by reordering is totally implicit. We could easily regress on that issue.

Philippe-Cholet:
**Breaking news:** the MSRV has been bumped to 1.63.0! 🎉

@scottmcm If I were dreaming of itertools at night, I would have nightmares not having `Iterator::next_chunk` available to us (stable and within our MSRV).
arrayvec/tinyvec was rejected (probable comments above).
This unsafe work sure is a _temporary fix_ for all that. We will rely on the stdlib in the end.

jswrenn:
@Philippe-Cholet, @phimuemue and @scottmcm, could you give this PR a final review? I've:
- revised the safety proofs
- modified the `Drop` impl avoid raw pointers and to have smaller `unsafe` blocks
- modified `push` to avoid unspecified behavior between builds with checked and wraparound arithmetic

src/next_array.rs
```diff
@@ -0,0 +1,144 @@
...
+            len: 0,
+        }
+    }
+
+    /// Pushes `value` onto the end of the array.
+    ///
+    /// # Panics
+    ///
+    /// This panics if `self.len >= N` or if `self.len == usize::MAX`.
+    pub fn push(&mut self, value: T) {
```
rcoh, line 32: I wonder if `try_push` would be a useful API here

src/next_array.rs
```diff
@@ -0,0 +1,144 @@
...
+            }))
+        } else {
+            None
+        }
+    }
+}
+
+impl<T, const N: usize> Drop for ArrayBuilder<T, N> {
+    fn drop(&mut self) {
+        // Select the valid elements of `self.arr`.
```
rcoh, line 81: I might add a comment noting that dropping `Uninit` does nothing (hence the reason for this function existing)

src/next_array.rs
```diff
@@ -0,0 +1,144 @@
...
+        // return value will be empty and `valid` will contain all of the
+        // elements that it did prior to the cast.
+        //
+        // SAFETY: It is sound to cast a `MaybeUninit<T>` that contains a valid
+        // `T` to a `T`. A `MaybeUninit<T>` is guaranteed to have the same size,
+        // alignment, and ABI as `T` [1], and by LEMMA 1, `valid` consists only
+        // of `MaybeUninit<T>` in the initialized state.
+        //
+        // [1]: https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#layout-1
+        let (_, valid, _): (_, &mut [T], _) = unsafe { valid.align_to_mut::<T>() };
```
rcoh, line 110: just curious—why not take the approach above where you use `arr.map { unsafe { v.assume_init() } }` then drop that?

src/next_array.rs
```diff
@@ -0,0 +1,144 @@
...
+    /// If `self.len() < N`, this returns `None`.
+    pub fn take(&mut self) -> Option<[T; N]> {
+        if self.len == N {
+            // SAFETY: Decreasing the value of `self.len` cannot violate the
+            // safety invariant on `self.arr`.
+            self.len = 0;
+
+            // SAFETY: Since `self.len` is 0, `self.arr` may safely contain
+            // uninitialized elements.
+            let arr = mem::replace(&mut self.arr, [(); N].map(|_| MaybeUninit::uninit()));
```
rcoh, line 66: could be simplified replace all of `self` with `ArrayBuilder::new()`?

src/next_array.rs
jswrenn: We cannot do so here because we're operating on a dynamic — not static — number of elements. Here, the only elements initialized are the ones from index `0` to `self.len`. By contrast, in `ArrayBuilder::take`, elements `0` to `N` (a constant) are initialized. `array::map` only operates on fixed-sized arrays, not slices.

src/next_array.rs
jswrenn: I don't believe so. I agree, it would be ideal if we could simply write:
```rust
let Self { arr, len } = mem::replace(self, Self::new());
```
instead. However, because `ArrayBuilder` has a non-trivial `Drop` implementation, we cannot move-destructure it in that manner. Instead, we need to read and write each field individually.

src/next_array.rs
jswrenn: [Done](https://github.com/rust-itertools/itertools/compare/461be14fb94929aa72957814e2603e63dd777cb0..4af66ad8f0af45e517bb98a230961f695ad019b7)

src/next_array.rs
```diff
@@ -0,0 +1,163 @@
...
+        // `0..self.len` are valid. Due to the above write, the element at
+        // `self.len` is now also valid. Consequently, all elements at indicies
+        // `0..(self.len + 1)` are valid, and `self.len` can be safely
+        // incremented without violating `self.arr`'s invariant. It is fine if
+        // this increment panics, as we have not created any intermediate
+        // invalid state.
+        self.len = match self.len.checked_add(1) {
+            Some(sum) => sum,
+            None => panic!("`self.len == usize::MAX`; cannot increment `len`"),
+        };
```
zachs18, line 51: Isn't this panic branch unreachable? Since `self.arr[self.len]` did not panic above, we therefore know that `self.len < N`, meaning that `self.len + 1 <= N` and thus `self.len + 1` cannot overflow.
```suggestion
        // PANICS: This cannot panic, since `self.len < N <= usize::MAX`.
        // SAFETY: By invariant on `self.arr`, all elements at indicies
        // `0..self.len` are valid. Due to the above write, the element at
        // `self.len` is now also valid. Consequently, all elements at indicies
        // `0..(self.len + 1)` are valid, and `self.len` can be safely
        // incremented without violating `self.arr`'s invariant.
        self.len += 1;
```

src/next_array.rs
zachs18: I was going to suggest using `std::ptr::read` instead of `std::mem::replace`ing with `[uninit(); N]`, but in release mode they seem to be optimized to the same thing (I assume the optimizer knows that writing `uninit()` is a no-op), so whichever is more readable is probably better. ([godbolt link](https://rust.godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DEArgoKkl9ZATwDKjdAGFUtEywYgArAA5SjgBk8BkwAOXcAI0xiEABmPwAHVAVCOwYXNw9vROTUgSCQ8JYomPjLTGtbASECJmICDPdPX3LKtJq6ggKwyOi4v3NOxqyWwfruopL%2BgEpLVBNiZHYOM0wAanN0EBA2Fm2AWSYATyi5BmDCAFJYgCFLjQBBe4eEkwiNglMbNYfiYmOAGqYZDXJwAFVIazQDHMa1CIDWZjwAC9MNdsGtLgB2O6PNb4tboJi1BGXLw3Q4nTBnC4EUFg9HXG5wskAEVIzwJa3onkRKVRHMe2NZz2eeBYCVo9Mh0Nh8L5KLRsQxv3%2BRyBINi4MhoXRmJxnIJr3esoIayoDDWIQA7hBpmsALR6oQVKj63EPLlcl20N3Yj1ewOE4lMUnk03utaU07nc4EbYmWOEO36kW3FledmGoMEnkIjSCz1B4XZ1OivFGt7my2vBQICCXABMADYWCYzVYqJCAG5iEzsNZg%2B3%2B0tczsAOiJtTJNwnPLZmNirKjxxjtO2NogvbcmGmTNHBLnjExjbuS7WXH3FfxJaF17Wxura1qAGtMAB9LgNlttjuu%2B1OsqawAPIJFUDCgjODLprqmZ6iO95cngbpHpaYBgNcy6hO6B6BsQmAEAslqhAISoBsWWIiohh6uuOPKLsuGhXkWgb0GadRqgxiIwkwVDrAhLE5h8WwgGBMQgPhTDoBAmzbGJ2xUMQqAsO%2B%2BFUN%2BzYTlOTDTOOojmHae7UTelHMTmQjKZgEAcccRmCbeTz3o%2BFrPkwb7vo2Gm/hs/6OnqoHgZB5LQcysFphiAk5shPm%2BnRx7oZhcI4cZXr4YRxDEaRZkUVRgljrR9GJUxty4dyBFrDZRxcbJOyYHsEmYJKTBLF57YxVQk4hpCM52kyGasuOLBMAkDZYk477Yk4K5UjScYJkmBCGXuJUpRZbDWX8xyDcNo1OBUdWTdxCi8fxOJrPtLDjkwCgKO4H60imwrTHZXIOW9QqNrEM6rOgbLPIMtjID8CIWq1ZqqoCwKgiYsSNpCsPogB/lgWkQU3DDjZ9QjcFAYlEPqsC2yvh%2Bl4rY5Tyfd9Si/XBjwA3gQM3CDDBgz8m0E5qTgY/DmPKkjQEBajWrfQj6bY%2BFXH4xqRNuR%2BvM3Bwsy0JwXi8J43C8KgnAAEpmB28yLPxn08KQ8YcFoz2kC%2B3iNuOGjNi2zaxFwPheLEWKuwAnPonCSGrmia5wvAKCABZmxbpBwLASBoBKdDROQlCxwk8cxOcwAIAQtBHHwdAENEIcQBEAekBEwR1EcnAm2XzDEEcwERNowJmybsdsIIwEMNnJdYBEJjAE4Yi0CHGukFgQ1GOI5u8Pg%2BE2Hg3aYCPWj%2BKowLtssJvBPnyvT/oeARP8dcuFgJefOKVe8IvxARMkmCsnVhjALQwSgNPsxUAYwAKACeCYNaoFGCXxkIIEQYh2BSBAfIJQagS66C4PoJ%2BEk9b7wiCHSAsxUAowECPB0LgwIOnoIvKUS5Yia2vsQPAWB0FQGYGwEAmB8DgVINufsnBGwaEbDwS2Vhm5pAcAwZwrgmh6ECMEHoxQ%2BgIKSCkcCwxPDSNyOBCYvQYgIN4fPaotR6jyL0Bo8CHRxjiMmFIyw2iGjCKyOo8xKjJFqNmAoA2SwJBKxVv7PeWsOBrFMOYIG6dM7ZzWBAXAhASAnmdtMXg4dFaRyQPMAgrwLAUAgMnVOoRWDLB8QQIGXBxweztrwRhoTKFbAQfwUBohxCQPKdAlQ6g97wNINaf4CRL6uI4KrUg6sV6eOAu2RJaxUBuiyX4vAGcs5VQgC4OO9BiDhK4JE02AdLbW0kFiccnt3axFiM2HwUgHZSAQbvP2pA9hcA0AWbpgcODB1Dks9%2B7TGzuJ6UHe5FtZjXxSPYSQQA%3D%3D%3D))

src/lib.rs
```diff
@@ -1914,6 +1915,50 @@ pub trait Itertools: Iterator {
...
+    /// is returned inside `Some`, otherwise `None` is returned.
+    ///
+    /// ```
+    /// use itertools::Itertools;
+    ///
+    /// let mut iter = 1..5;
+    ///
+    /// assert_eq!(Some([1, 2]), iter.next_array());
+    /// ```
+    fn next_array<T, const N: usize>(&mut self) -> Option<[T; N]>
```
Philippe-Cholet, line 1931: `T` is the same as `Self::Item`, _we don't need a generic here!_
Same for `collect_array` and the private `next_array::next_array`.

@jswrenn I don't feel competent enough to review unsafe parts, I'll let you judge.

orlp:
@jswrenn In your refactor you've introduced two bugs, one of which is undefined behavior:

1. The new code only frees the first element and leaks the rest on drop.
2. The new code calls `drop` on an uninitialized value if the `ArrayBuilder` is empty.

Please read my original version carefully, I intentionally used [`ptr::slice_from_raw_parts_mut`](https://github.com/rust-itertools/itertools/blob/70842fc3d2173a35c03459d6077927838daf6f13/src/next_array.rs#L65) to transform the pointer-to-first-element (`*mut T`) returned by [`[T]::as_mut_ptr`](https://doc.rust-lang.org/std/primitive.slice.html#method.as_mut_ptr) into a pointer-to-slice (`*mut [T]`) to then drop the slice in-place.  That's why the variables were called `ptr_to_first` and `ptr_to_slice`.

Your new code only makes a pointer to the first element, and forcibly only drops that element, regardless of whether it is initialized or not.

I can't help but point out that adding 60 lines of comments and various extra steps to what used to be a correct 3 LOC + 4 lines of comments implementation did not help with preventing this bug for the writer, and for me personally as a reader it made the issue harder to spot.

jswrenn:
I appreciate the catch! Given that this passed CI, it looks like you've also uncovered a gap in our test coverage.

Orson, you will not have me in agreement on the commenting issue. I wouldn't have to read your original version so carefully to guess at your intentions if had you had left comments making them explicit. The comments are for the benefit of maintainers, who need to make sure we don't regress on any of these subtleties in perpetuity.

And the comments *did* result in a potential soundness issue being discovered by a reviewer — the original code leaves a live reference to an invalid (dropped) referent.

orlp:
> And the comments did result in a potential soundness issue being discovered by a reviewer — the original code leaves a live reference to an invalid (dropped) referent. In fact, the regression you just discovered [was introduced](https://github.com/rust-itertools/itertools/compare/4af66ad8f0af45e517bb98a230961f695ad019b7..9c33e3dccbdcb105856380a9b362c60cb323f47a) as part of fixing **this potential soundness issue that was present in your original PR**.

That was never part of my original PR, but also due to your refactor. My original code never left a live reference. I implore you to look at [the original code](https://github.com/rust-itertools/itertools/blob/70842fc3d2173a35c03459d6077927838daf6f13/src/next_array.rs#L65) again.

Even if we take that at face value, your purported solution does nothing to resolve it. The following code compiles:

```rust
let mut x = 42;
let mut r = &mut x;
{
    // <pointer stuff>
    let _ = r; // This does not drop r! This is a no-op!
}
*r = 42; // r is still live!
```

orlp:
@jswrenn And it gets worse.

Even if you had used `ptr::slice_from_raw_parts_mut` to turn `valid: &mut [T]` into `*mut [T]` and even if you had indeed used `drop(valid)` to drop `valid`... it **still** would be undefined behavior by Miri's stacked borrows model. The slice from which a pointer is derived must stay alive for that pointer to remain valid under stacked borrows. That is, this is UB under stacked borrows:

```rust
let mut arr = [1, 2, 3];
let r = &mut arr[..];
let p = r.as_mut_ptr();
drop(r);
unsafe { dbg!(&*p) };
```


|  `drop`  as intended | pointer-to-slice as intended | Outcome |
| ------------- | ------------- | - |
| No  | No  | Undefined behavior, uninitialized element dropped |
| No  | Yes  | Code accidentally correct, comments incorrect and perceived potential unsoundness from live `&mut [T]` unadressed |
| Yes  | No  | Undefined behavior, invalidated pointer accessed and uninitialized element dropped |
| Yes  | Yes  | Undefined behavior, invalidated pointer accessed |

I kindly ask you to reflect on why the refactor, despite what I assume is your best intent and care with respect to the safety comments, did not catch either of the above issues nor that their intended combination still would be invalid.

jswrenn:
> I kindly ask you to reflect on why the refactor, despite what I assume is your best intent and care with respect to the safety comments, did not catch either of the above issues nor that their intended combination still would be invalid.

Whatever the reason, I'm sure it's not because I wrote too many comments.

I appreciate your knowledgeability, but, again, this abstraction has to survive idiots like me and whoever else contributes to itertools in the future. Its soundness cannot depend on the brilliance of a lone genius. I'm delighted to take your suggestions, but I *am* going to make the reasoning behind those suggestions explicit with new comments.

jswrenn:
Stripping most of the comments out for your benefit, could you review this implementation (essentially undoing [this](https://github.com/rust-itertools/itertools/compare/4af66ad8f0af45e517bb98a230961f695ad019b7..9c33e3dccbdcb105856380a9b362c60cb323f47a)):

```rust
fn drop(&mut self) {
    // Select the valid elements of `self.arr`.
    let (valid, _) = self.arr.split_at_mut(self.len);

    // Cast `valid` from `&[MaybeUninit<T>]` to `&[T]`
    let (_, valid, _): (_, &mut [T], _) = unsafe { valid.align_to_mut::<T>() };

    // Drop `valid`.
    unsafe {
        ptr::drop_in_place(valid);
    }
}
```

orlp:
@jswrenn While I believe that is sound, I find using `align_to_mut` to assume the slice is initialized to be very strange, and it's certainly an intent-implementation semantic mismatch.

As an alternative, I would propose that we copy [`MaybeUninit::slice_assume_init_mut`](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#method.slice_assume_init_mut) from the standard library with a simple safety comment deferring to the standard library:

```rust
unsafe fn maybeuninit_slice_assume_init_mut<T>(slice: &mut [MaybeUninit<T>]) -> &mut [T] {
    // SAFETY: see the standard library implementation of MaybeUninit::slice_assume_init_mut,
    // this implementation is copied from there as it is not available at our MSRV.
    unsafe { &mut *(slice as *mut [MaybeUninit<T>] as *mut [T]) }
}
```

Furthermore I also don't understand why you use `split_at_mut` and then promptly ignore one half. May I suggest instead the following, also making things more re-usable / composable:

```rust
impl<T, const N: usize> AsMut<[T]> for ArrayBuilder<T, N> {
    fn as_mut(&mut self) -> &mut [T] {
        // SAFETY: self.arr[..self.len] is valid.
        unsafe { maybeuninit_slice_assume_init_mut(&mut self.arr[..self.len]) }
    }
}
```

Then our `drop` is painfully simple:

```rust
impl<T, const N: usize> Drop for ArrayBuilder<T, N> {
    fn drop(&mut self) {
        unsafe { core::ptr::drop_in_place(self.as_mut()) }
    }
}
```

jswrenn:
I quite like that formulation!

src/lib.rs
```diff
@@ -1914,6 +1915,50 @@ pub trait Itertools: Iterator {
...
+    ///
+    /// if let Some([x, y]) = iter.collect_array() {
+    ///     assert_eq!([x, y], [1, 2])
+    /// } else {
+    ///     panic!("Expected two elements")
+    /// }
+    /// ```
+    fn collect_array<T, const N: usize>(mut self) -> Option<[T; N]>
+    where
+        Self: Sized + Iterator<Item = T>,
```
Philippe-Cholet, line 1957: @jswrenn `T` is not needed here too.

Philippe-Cholet:
@jswrenn About CI, I just noticed 😮 that the `needs` list in `all-jobs-succeed` does not include `miri`. Let's add it here.

src/next_array.rs
```diff
@@ -0,0 +1,254 @@
...
+
+    /// Pushes `value` onto the end of the array.
+    ///
+    /// # Panics
+    ///
+    /// This panics if `self.len >= N`.
+    #[inline(always)]
+    pub fn push(&mut self, value: T) {
+        // PANICS: This will panic if `self.len >= N`.
+        let place = &mut self.arr[self.len];
```
joshlf, line 34: Maybe use `.get_mut()` followed by `.unwrap()` or `.expect()` to make the panic condition more explicit?

```suggestion
        // PANICS: This will panic if `self.len >= N`.
        let place = self.arr.get_mut(self.len).unwrap();
```

src/next_array.rs
```diff
@@ -0,0 +1,254 @@
...
+        // invariant of `self.arr`. Even if this line panics, we have not
+        // created any intermediate invalid state.
+        *place = MaybeUninit::new(value);
+        // PANICS: This cannot panic, since `self.len < N <= usize::MAX`.
+        // `0..self.len` are valid. Due to the above write, the element at
+        // `self.len` is now also valid. Consequently, all elements at indicies
+        // `0..(self.len + 1)` are valid, and `self.len` can be safely
+        // incremented without violating `self.arr`'s invariant. It is fine if
+        // this increment panics, as we have not created any intermediate
+        // invalid state.
```
joshlf, line 47: A few concerns with this:
- Parts of this comment are actually a safety comment, not a panics comment
- The safety comment aspects of this are insufficient

Here's a suggested update:

```suggestion
        // Lemma: `self.len < N`. By invariant, `self.len <= N`. Above, we index
        // into `self.arr`, which has size `N`, at index `self.len`. If `self.len == N`
        // at that point, that index would be out-of-bounds, and the index
        // operation would panic. Thus, `self.len != N`, and since `self.len <= N`,
        // that means that `self.len < N`.
        //
        // PANICS: Since `self.len < N`, and since `N <= usize::MAX`,
        // `self.len + 1 <= usize::MAX`, and so `self.len += 1` will not
        // overflow. Overflow is the only panic condition of `+=`.
        //
        // SAFETY:
        // - We are required to uphold the invariant that `self.len <= N`.
        //   Since, by the preceding lemma, `self.len < N` at this point in the
        //   code, `self.len += 1` results in `self.len <= N`.
        // - We are required to uphold the invariant that `self.arr[..self.len]`
        //   are valid instances of `T`. Since this invariant already held when
        //   this method was called, and since we only increment `self.len`
        //   by 1 here, we only need to prove that the element at
        //   `self.arr[self.len]` (using the value of `self.len` before incrementing)
        //   is valid. Above, we construct `place` to point to `self.arr[self.len]`,
        //   and then initialize `*place` to `MaybeUninit::new(value)`, which is
        //   a valid `T` by construction.
```

src/next_array.rs
```diff
@@ -0,0 +1,254 @@
...
+
+impl<T, const N: usize> Drop for ArrayBuilder<T, N> {
+    // We provide a non-trivial `Drop` impl, because the trivial impl would be a
+    // no-op; `MaybeUninit<T>` has no innate awareness of its own validity, and
+    // so it can only forget its contents. By leveraging the safety invariant of
+    // `self.arr`, we do know which elements of `self.arr` are valid, and can
+    // selectively run their destructors.
+    fn drop(&mut self) {
+        // SAFETY:
+        // - by invariant on `&[T]`, `self.as_mut()` is:
```
joshlf, line 95:
```suggestion
        // - by invariant on `&mut [T]`, `self.as_mut()` is:
```

joshlf (changes requested):

src/next_array.rs
jswrenn: Going to stick with the bracket indexing, as that will ensure that the panic message here matches the familiar one used by the standard library for out-of-bounds indexing.

jswrenn (approved):

2moe:
When I first used `collect_array`, I didn't read the documentation carefully. I thought this code would run, but it didn't.

```rust
let actual = (1..=4) // 1, 2, 3, 4
    .collect_array::<2>(); // None

let expected = Some([1, 2]);

assert_eq!(expected, actual);
// assertion `left == right` failed
// left: Some([1, 2])
// right: None
```

I quickly realized that I needed to use `.take(N)` to limit the number of iterations. The following code works, but it doesn't feel as good as the one above.

```rust
let arr = {
    const N: usize = 2;
    (1..=4)
        .take(N)
        .collect_array::<N>()
};
```

Considering that the array size is known at compile time, could we add a method like `collect_array_exact()` or `take_and_collect_array()`?

orlp:
@2moe `next_array` already does what you want.

2moe:
@orlp Thank you for your patient explanation. Have a wonderful day. 
(Perhaps if I had taken a closer look at the documentation, I wouldn't have had the questions above.)

Issue #559, opened 2021-07-13
Consider to add a README.md to crates.io


rusty-snake:
Currently this crate does has no README.md (and no readme key in Cargo.toml) causing it to be rendered without a README on crates.io. Why is that bad? Crates which miss a README on crates.io look (and often are) not well maintained, unfinished and/or experimental. Even a samll README with a titel, a description and one example looks better.

PR #561, opened 2021-07-16
Add crude README to Cargo.toml

phimuemue:
Addressing https://github.com/rust-itertools/itertools/issues/559, this adds our README to Cargo.toml.

Instead of creating a separate one, I suggest simply redirecting to the existing one and update it if needed.


jswrenn:
AFAIK, crates.io _only_ interprets the file at the `readme` key as markdown (and ours is reStructuredText), so this change would render very poorly.

We'd be better off translating the existing README.rst into markdown, I think.

phimuemue:
I converted it to markdown, and I think they look reasonably similar. Here's the comparable files:

* reStructuredText: https://github.com/rust-itertools/itertools/blob/master/README.rst
* markdown: https://github.com/phimuemue/rust-itertools/blob/readme/README.md

README.md
```diff
@@ -1,50 +1,39 @@
...
-
-|build_status|_ |crates|_
+Please read the [API documentation here](https://docs.rs/itertools/)
 
-.. |build_status| image:: https://travis-ci.org/rust-itertools/itertools.svg?branch=master
-.. _build_status: https://travis-ci.org/rust-itertools/itertools
-
-.. |crates| image:: https://meritbadge.herokuapp.com/itertools
-.. _crates: https://crates.io/crates/itertools
+[![build_status](https://travis-ci.org/rust-itertools/itertools.svg?branch=master)](https://travis-ci.org/rust-itertools/itertools)
```
atouchet, line 7: This should probably be updated to use GitHub Actions instead of Travis CI.

phimuemue: @atouchet Could you possibly review these patches and confirm that the new README.md resembles what we had previously in README.rst? If so, I'd merge it after your review.

atouchet: Sorry for missing this earlier. The changes look mostly good to me and I can submit some suggestions as a followup.

PR #563, opened 2021-07-20
Specialize ProcessResults::fold

### LINKED ISSUE #562: process_results is lacking a lot of function specializations
fold/try_fold are missing on the iterator adapter, resulting in a lot of underlying implementations possibly being slower (including in particular collect/into_group_map, etc...).

I believe we should just copy over the standard library's implementation (that's actually used in `FromIterator<Result<T,E>> for Result<Collection, E> where Collection<FromIterator<T>>`):
https://github.com/rust-lang/rust/blob/68511b574ffe019a5cb3e9fa92605f80d39167bc/library/core/src/iter/adapters/mod.rs#L125-L197

SkiFire13:
Partially fixes #562. `try_fold` can't be added right now because it requires the unstable `Try` trait.

### DIFF
--- a/src/process_results_impl.rs
+++ b/src/process_results_impl.rs
@@ -30,6 +30,23 @@ impl<'a, I, T, E> Iterator for ProcessResults<'a, I, E>
     fn size_hint(&self) -> (usize, Option<usize>) {
         (0, self.iter.size_hint().1)
     }
+
+    fn fold<B, F>(mut self, init: B, mut f: F) -> B
+    where
+        Self: Sized,
+        F: FnMut(B, Self::Item) -> B,
+    {
+        let error = self.error;
+        self.iter
+            .try_fold(init, |acc, opt| match opt {
+                Ok(x) => Ok(f(acc, x)),
+                Err(e) => {
+                    *error = Err(e);
+                    Err(acc)
+                }
+            })
+            .unwrap_or_else(|e| e)
+    }
 }
 
 /// “Lift” a function of the values of an iterator so that it can process
--- a/tests/specializations.rs
+++ b/tests/specializations.rs
@@ -98,3 +98,50 @@ quickcheck! {
         test_specializations(&v.into_iter().map_ok(|u| u.checked_add(1)));
     }
 }
+
+quickcheck! {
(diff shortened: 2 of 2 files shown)


phimuemue:
Thanks a lot for this - looks very good to me (as always with your contributions).

And even if I *think* this is correct, I think we do ourselves a favor if we have tests for this specialization. Could you add some? (I think we already have quicktests for specializations.)

SkiFire13:
If you mean `tests/specializations.rs` I wonder if those tests are even doing something. The `mapper` function gets a `Box<dyn Iterator<Item = IterItem> + 'a>` which is (almost) unspecialized anyway due to the implementation of `Iterator` for `Box<dyn Iterator>`.

Looks like there's also a `tests/fold_specialization.rs`, but that doesn't have quicktests.

phimuemue:
Thank you.

bors r+

Ten0:
>  The `mapper` function gets a `Box<dyn Iterator<Item = IterItem> + 'a>` which is (almost) unspecialized anyway due to the implementation of `Iterator` for `Box<dyn Iterator>`

https://github.com/rust-lang/rust/blob/914a1e2c517dfbaa23a4ec4a3eebefb3e2c253c2/library/alloc/src/boxed.rs#L1566-L1581
Looks like it's true!
That's unexpected! Why doesn't rust specialize this? :o

Looks like we should replace the iterator dynamic dispatch in check_specialized with a macro (like the one you created), as it does not fulfill its purpose.

scottmcm:
> That's unexpected! Why doesn't rust specialize this? :o

Because all the generic methods on `Iterator` can't flow into the `dyn` thing.  So anything called on `dyn Iterator` can't use any of the `try_fold` overrides or similar.

Ten0:
Had finally figured out the issues with `Sized` (actually, none of the methods that take `self` can flow in the `dyn Iterator`) and was coming back to edit my comment when I saw your answer from 1 minute ago ^^



> Looks like we should replace the iterator dynamic dispatch in check_specialized with a macro (like the one you created), as it does not fulfill its purpose.

Actually already done in bf4ca16746d93d32cd0cc754bd2d2cad92e924a5

PR #564, opened 2021-07-23
Fix crates.io badge

atouchet:


phimuemue:
Thanks for this! I incorporated it into https://github.com/rust-itertools/itertools/pull/561.

PR #567, opened 2021-07-29
Fix documentation of `HomogeneousTuple`

steffahn:
Compare the docs of `Itertools::collect_tuple` that already mention the correct upper bound on the number of tuple elements.


phimuemue:
Thank you. I think there are some more mis-documented tuple things... We'll have to fix them.

bors r+

PR #571, opened 2021-08-10
Fix overflow in duplicates size hint

bjchambers:
The math for size hint on duplicates would panic (with an overflow) if more elements were pending than remained in the size hint of the input iterator. This wasn't being tested since `pending` is only updated once some elements are advanced from the iterator.

This is actually pretty common, since collecting to a vector calls `next` first (before getting the size hint), for some reason related to branch prediction.

This adds an explicit test for collecting a vector from duplicates as well as a quick check case that skips elements before checking the size hint. It's not clear why the exhaustive check inside quickcheck wasn't catching this.

### DIFF
--- a/src/duplicates_impl.rs
+++ b/src/duplicates_impl.rs
@@ -2,8 +2,8 @@ use std::hash::Hash;
 
 mod private {
     use std::collections::HashMap;
-    use std::hash::Hash;
     use std::fmt;
+    use std::hash::Hash;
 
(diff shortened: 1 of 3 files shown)


phimuemue:
Hi there, I think this is the same as https://github.com/rust-itertools/itertools/pull/552.

But we should probably incorporate your quickcheck once we merge it.

phimuemue:
Closing because we already have #552.

PR #574, opened 2021-08-20
Fix and refactor specializations tests

SkiFire13:
As found in #563 specializations tests don't actually test the specialize tests due to the use of `Box<dyn Iterator>`. This PR fix this problem by having `check_specialized` be a macro so that the code inside (what used to be the) closure directly calls methods on the iterator types.

This also does a tiny refactor, moving the single test in `tests/fold_specialization.rs` together with the others in `tests/specializations.rs`.

Note that this may conflict with #563 (but I hope it doesn't since they touch different parts of the file)


phimuemue:
Awesome, I just wanted to do this myself. Thanks!

bors r+

PR #587, opened 2021-12-10
Feature `process_results` as an `Itertools` method

### LINKED ISSUE #557: process_results on the Itertools trait
A typical use-case for process-results is e.g.
```rust
let final_map = itertools::process_results(
    some
        .long()
        .chain()
        .of()
        .iterator()
        .adapters(),
    |it| it.into_group_map(),
)?;
```

With the chain of iterator adapters getting longer and longer, this writing becomes less and less readable.

=> It would be much nicer to be able to write:
```rust
let final_map = some
    .long()
    .chain()
    .of()
    .iterator()
    .adapters()
    .process_results(|it| it.into_group_map())?;
```

This would be pretty easy to do by adding process_results in the `Itertools` trait.

danielhenrymantilla:
Fixes #557

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -864,6 +864,50 @@ pub trait Itertools : Iterator {
         adaptors::map_ok(self, f)
     }
 
+    /// “Lift” a function of the values of the current iterator so as to process
+    /// an iterator of `Result` values instead.
+    ///
+    /// `processor` is a closure that receives an adapted version of the iterator
+    /// as the only argument — the adapted iterator produces elements of type `Ok`,
+    /// as long as the original iterator produces `Ok` values.
+    ///
+    /// If the original iterable produces an error at any point, the adapted
+    /// iterator ends (short-circuits) and the `process_results` function will
+    /// return the error iself.
+    ///
+    /// Otherwise, the return value from the closure is returned wrapped
+    /// inside `Ok`.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use itertools::Itertools;
+    ///
+    /// type Item = Result<i32, &'static str>;
+    ///
+    /// let first_values: Vec<Item> = vec![Ok(1), Ok(0), Ok(3)];
+    /// let second_values: Vec<Item> = vec![Ok(2), Ok(1), Err("overflow")];
+    ///
+    /// // “Lift” the iterator .max() method to work on the Ok-values.
+    /// let first_max = first_values.into_iter().process_results(|iter| iter.max().unwrap_or(0));
+    /// let second_max = second_values.into_iter().process_results(|iter| iter.max().unwrap_or(0));
+    ///
+    /// assert_eq!(first_max, Ok(3));
+    /// assert!(second_max.is_err());
+    /// ```
+    ///
+    /// This is the iterator adaptor version of [`itertools::process_results()`].
+    ///
+    /// [`itertools::process_results`]: `crate::process_results`
+    fn process_results<F, Ok, Err, R>(self, processor: F) -> Result<R, Err> 
+        where Self: Iterator<Item = Result<Ok, Err>> + Sized,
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -864,6 +864,50 @@ pub trait Itertools : Iterator {
         adaptors::map_ok(self, f)
     }
 
+    /// “Lift” a function of the values of the current iterator so as to process
+    /// an iterator of `Result` values instead.
+    ///
+    /// `processor` is a closure that receives an adapted version of the iterator
+    /// as the only argument — the adapted iterator produces elements of type `Ok`,
+    /// as long as the original iterator produces `Ok` values.
```
phimuemue, line 872: This documentation differs slightly from what we have at https://docs.rs/itertools/0.10.3/itertools/fn.process_results.html, and I think, the choice of `Ok` is misleading, as it hints at a `Result` - https://docs.rs/itertools/0.10.3/itertools/fn.process_results.html goes with `T`, I suggest copying that to have it in line.

src/lib.rs
```diff
@@ -864,6 +864,50 @@ pub trait Itertools : Iterator {
...
+    ///
+    /// If the original iterable produces an error at any point, the adapted
+    /// iterator ends (short-circuits) and the `process_results` function will
+    /// return the error iself.
+    ///
+    /// Otherwise, the return value from the closure is returned wrapped
+    /// inside `Ok`.
```
phimuemue, line 879: This is misleading, as this `Ok` refers to `Result::Ok` (and not the local type `Ok`). Or am I missing something?

src/lib.rs
```diff
@@ -864,6 +864,50 @@ pub trait Itertools : Iterator {
...
+    ///
+    /// // “Lift” the iterator .max() method to work on the Ok-values.
+    /// let first_max = first_values.into_iter().process_results(|iter| iter.max().unwrap_or(0));
+    /// let second_max = second_values.into_iter().process_results(|iter| iter.max().unwrap_or(0));
+    ///
+    /// assert_eq!(first_max, Ok(3));
+    /// assert!(second_max.is_err());
+    /// ```
+    ///
+    /// This is the iterator adaptor version of [`itertools::process_results()`].
```
phimuemue, line 899: Nice.

src/lib.rs
```diff
@@ -864,6 +864,50 @@ pub trait Itertools : Iterator {
...
+    ///
+    /// [`itertools::process_results`]: `crate::process_results`
+    fn process_results<F, Ok, Err, R>(self, processor: F) -> Result<R, Err> 
+        where Self: Iterator<Item = Result<Ok, Err>> + Sized,
+              F: for<'r> FnOnce(ProcessResults<'r, Self, Err>) -> R,
```
phimuemue, line 904: Is the higher kinded lifetime (i.e. `for<'r>`) necessary?

src/lib.rs
```diff
@@ -864,6 +864,50 @@ pub trait Itertools : Iterator {
...
+              // Added to directly highlight the relevant impl in the documentation
+              for<'r> ProcessResults<'r, Self, Err>: Iterator<Item = Ok>,
```
phimuemue, line 906: Do we need this? If it's only for documentation, I'd prefer to remove it and not bother the type checker with it. If we really want it in the documentation, we can still state it explicitly.

phimuemue (commented):
Hi there, thanks for this, it's appreciated.

As of now, the majority of free function is documented as follows:
* Summary line
* `IntoIterator` enabled version of `method`.

To enforce some consistency, could you please go with that pattern?

sjackman:
@danielhenrymantilla Thanks for this work, Daniel. Gentle nudge. I'd love to see this PR merged.

phimuemue:
Thanks for bringing this up.

We took the idea, but went with the somewhat simpler https://github.com/rust-itertools/itertools/pull/680

PR #591, opened 2021-12-22
Canonicalize free functions

phimuemue:
As made visible by https://github.com/rust-itertools/itertools/pull/587, I think we should have (somewhat) uniform documentation of free functions.

On top of that, I think `fn equal` should actually use `Iterator::eq` (available since Rust 1.5), instead of only referring to it in the documentation.

Maybe someone finds time to have a look at this, as I'd like to avoid silly mistakes before merging.


jswrenn (approved):
bors r+

Issue #583, opened 2021-12-04
Why `PeekingNext` does not provide `peek` method?


TonalidadeHidrica:
In my understanding, `PeekingNext` is a iterator from which we can obtain the reference to the next element.  Currently this trait requires `peeking_next` function, that consumes the next element when the predicate is satisfied.  But what if I want the `peek` behavior, in which the next element is *not* consumed?  I think it is reasonable to provide such trait method too.  Is there a reason that there isn't?

jendrikw:
I think this can be implemented with something like this:

```rust
fn peek(&mut self) -> bool {
    let mut element;
    self.peeking_next(|x| {
        element = x;
        false
    });
    element
}
```

TonalidadeHidrica:
I see, that's not really clean but it does work.  Thanks.

jendrikw:
I agree that implementing `peeking_next` in terms of `peek` would be better.

PR #593, opened 2021-12-29
EitherOrBoth: Add or and or_else methods to simplify getting default values

carl-anders:
## Introducing `or` and `or_else` methods to `EitherOrBoth`
This feature reduces the amount of code required to get specific default values when using EitherOrBoth.
An example of the current way of using `zip_longest` with custom default values:
```rust
let a = (0..=4).into_iter();
let b = (0..=2).into_iter();
let c = a.zip_longest(b).map(|e| match e {
    EitherOrBoth::Both(l, r) => (l, r),
    EitherOrBoth::Left(l) => (l, 2),
    EitherOrBoth::Right(r) => (4, r),
});
// c will now contain an iterator with (0,0), (1,1), (2,2), (3,2), (4,2).
```
An example with the proposed `or` method:
```rust
let a = (0..=4).into_iter();
let b = (0..=2).into_iter();
let c = a.zip_longest(b).map(|e| e.or(4, 2));
// c will now contain an iterator with (0,0), (1,1), (2,2), (3,2), (4,2).
```
I have also included the `or_else` method which does the same but with closures.

There are no tests for the other `EitherOrBoth` methods, so I was unsure how to place them. However, I have added **DocTest**'s to both methods. These examples allows for easier to understand documentation, and also tests the validity of the methods.

> For new features, please first consider filing a PR to rust-lang/rust

The EitherOrBoth struct does not exist in rust std library.

## Concerns

The naming is slightly inconsistent when compared to rust's `std::Option`, considering the naming from there the added methods should be named `unwrap_or` and `unwrap_or_else`.
However, this would then become inconsistent with the existing method `EitherOrBoth::or_default`. Which is why I went with the chosen names.
I can change the method names if needed, but then we should consider changing `or_default` too.

## P.S.
The `CHANGELOG.md` file has the text `Add EitherOrBoth::or_default (#583)`. This number is wrong, it should be #538.

### DIFF
--- a/src/either_or_both.rs
+++ b/src/either_or_both.rs
@@ -165,6 +165,33 @@ impl<A, B> EitherOrBoth<A, B> {
     }
 
     /// Returns a tuple consisting of the `l` and `r` in `Both(l, r)`, if present.
+    /// Otherwise, returns the wrapped value for the present element, and the supplied
+    /// value for the other. The first (`l`) argument is used for a missing `Left`
+    /// value. The second (`r`) argument is used for a missing `Right` value.
+    ///
+    /// Arguments passed to `or` are eagerly evaluated; if you are passing
+    /// the result of a function call, it is recommended to use [`or_else`],
+    /// which is lazily evaluated.
+    ///
+    /// [`or_else`]: EitherOrBoth::or_else
+    ///
+    /// # Examples
+    ///
+    /// ```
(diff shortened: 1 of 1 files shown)


phimuemue:
Thanks for this; a nice generalization of `or_default`! I think we can take this as it is.

bors r+

PR #594, opened 2022-01-04
Add `equal() -> Option<Self::Item>` as companion to `all_equal() -> bool`

Byron:
This allows use the equal value or to default it, which otherwise is cumbersome and complex as indicated in this example from `nameless`:

```diff
diff --git a/examples/text-cat.rs b/examples/text-cat.rs
index de1ceea..aac4db5 100644
--- a/examples/text-cat.rs
+++ b/examples/text-cat.rs
@@ -10,12 +10,12 @@ use nameless::{InputTextStream, LazyOutput, MediaType, OutputTextStream};
 /// * `inputs` - Input sources, stdin if none
 #[kommand::main]
 fn main(output: LazyOutput<OutputTextStream>, inputs: Vec<InputTextStream>) -> anyhow::Result<()> {
-    let media_type = match inputs.iter().next() {
-        Some(first) if inputs.iter().map(InputTextStream::media_type).all_equal() => {
-            first.media_type().clone()
-        }
-        _ => MediaType::text(),
-    };
+    let media_type = inputs
+        .iter()
+        .map(InputTextStream::media_type)
+        .equal()
+        .cloned()
+        .unwrap_or_else(MediaType::text);

     let mut output = output.materialize(media_type)?;
```

### Open Questions

* Is `equal()` a fitting name? Are there better alternatives?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1882,6 +1882,33 @@ pub trait Itertools : Iterator {
         self.all(move |elt| used.insert(elt))
     }
 
+    /// Check whether all elements compare equal, and return `Some` with the equal element.
+    ///
+    /// Empty iterators return `None`, and so do iterators with non-equal elements.
+    ///
+    /// ```
+    /// use itertools::Itertools;
+    ///
+    /// let data = vec![1u8, 1, 1, 2, 2, 3, 3, 3, 4, 5, 5];
+    /// assert_eq!(data.iter().equal(), None);
+    /// assert_eq!(data[0..3].iter().equal(), Some(&1));
+    /// assert_eq!(data[3..5].iter().equal(), Some(&2));
+    /// assert_eq!(data[5..8].iter().equal(), Some(&3));
+    ///
+    /// let data : Option<usize> = None;
+    /// assert_eq!(data.into_iter().equal(), None);
+    /// ```
+    fn equal(&mut self) -> Option<Self::Item>
+    where
+        Self: Sized,
+        Self::Item: PartialEq,
+    {
+        match self.next() {
+            None => None,
+            Some(a) => self.all(|x| a == x).then(|| a),
+        }
+    }
+
     /// Consume the first `n` elements from the iterator eagerly,
     /// and return the same iterator again.
     ///
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1905,7 +1905,7 @@ pub trait Itertools : Iterator {
     {
         match self.next() {
             None => None,
-            Some(a) => self.all(|x| a == x).then(|| a),
+            Some(a) => if self.all(|x| a == x) { Some(a) } else { None }
         }
     }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1890,15 +1890,15 @@ pub trait Itertools : Iterator {
     /// use itertools::Itertools;
     ///
     /// let data = vec![1u8, 1, 1, 2, 2, 3, 3, 3, 4, 5, 5];
-    /// assert_eq!(data.iter().equal(), None);
-    /// assert_eq!(data[0..3].iter().equal(), Some(&1));
-    /// assert_eq!(data[3..5].iter().equal(), Some(&2));
-    /// assert_eq!(data[5..8].iter().equal(), Some(&3));
+    /// assert_eq!(data.iter().all_equal_item(), None);
+    /// assert_eq!(data[0..3].iter().all_equal_item(), Some(&1));
+    /// assert_eq!(data[3..5].iter().all_equal_item(), Some(&2));
+    /// assert_eq!(data[5..8].iter().all_equal_item(), Some(&3));
     ///
     /// let data : Option<usize> = None;
-    /// assert_eq!(data.into_iter().equal(), None);
+    /// assert_eq!(data.into_iter().all_equal_item(), None);
     /// ```
-    fn equal(&mut self) -> Option<Self::Item>
+    fn all_equal_item(&mut self) -> Option<Self::Item>
     where
         Self: Sized,
         Self::Item: PartialEq,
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -1882,6 +1882,33 @@ pub trait Itertools : Iterator {
...
+    /// assert_eq!(data.into_iter().equal(), None);
+    /// ```
+    fn equal(&mut self) -> Option<Self::Item>
+    where
+        Self: Sized,
+        Self::Item: PartialEq,
+    {
+        match self.next() {
+            None => None,
+            Some(a) => self.all(|x| a == x).then(|| a),
```
phimuemue, line 1908: `bool::then` was added in Rust 1.50, and I do not think we want to bump our minimum required Rust version because of this. Maybe you can change this to a simple `if/else`.

src/lib.rs
```diff
@@ -1882,6 +1882,33 @@ pub trait Itertools : Iterator {
...
+    /// assert_eq!(data[3..5].iter().equal(), Some(&2));
+    /// assert_eq!(data[5..8].iter().equal(), Some(&3));
+    ///
+    /// let data : Option<usize> = None;
+    /// assert_eq!(data.into_iter().equal(), None);
+    /// ```
+    fn equal(&mut self) -> Option<Self::Item>
+    where
+        Self: Sized,
+        Self::Item: PartialEq,
```
phimuemue, line 1904: Trait bounds in line with `all_equal` - nice.

phimuemue (commented):
Hi! Thanks for this.

I'm tempted to take this into itertools after replacing the usage of `bool::then`. Just noting some things we might want to consider:
* Naming/bikeshedding: Would `all_equal_item` (or similar) be more appropriate than `equal`?
* Return type: Should we distinguish the empty case from the non-equal-elements case? (We would then possibly return a `Result<Self::Item, EmptyOrNonEqual>`. I would be willing to accept the PR as it is and wait if we actually get complaints, but @jswrenn has the last word on this.)
* I thought that this function should possibly share code with `Itertools::all_equal`, but since it's so small, I'm happy with your implementation.

scottmcm:
`equal` to me is the same name as `eq`, so I find it very odd that it's not returning anything `bool`-like.

This is an efficient way to do `.dedup().exactly_one().ok()`, right?  So maybe that can provide naming inspiration.

To me, this feels like it wants to return something structurally equivalent to a `MinMaxResult`: Zero elements, One element, or Multiple elements.  Or maybe something like what [`exactly_one`](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.exactly_one) does...

src/lib.rs
Byron: Thanks for the hint. Addressed in 626b84428a6d50a97e869a7c06eb44a3dc9904f7 .

src/lib.rs
Byron: And now it's also renamed to `all_equal_item()`, which I love as it is discoverable when coming from `all_equal()`. See 426f0d6feba0a0af7f43501518d1bc70d47f7595 .

scottmcm: Indeed, should `all_equal` be implemented as `self.all_equal_item().is_some()` now?

Byron:
Thanks everyone for helping to improve this PR :)! I have addressed the suggestions of @phimuemue which included changing the name to `all_equal_item()`, which has advantages for discovering this method.

I understand @scottmcm suggests we could return a `Result`, but for that I would need guidance as I can't seem to find the connection to `exactly_one()` and how a special Err value can improve the usability.

src/lib.rs
Byron: A great catch, implemented in 3b1d4223c245aae0abe8303beaa964cbce816fa3 .

scottmcm:
I think the `Option` approach, as you have it now, is good, especially with the `all_equal_item` name -- a 2-state response is consistent with the `bool` of `all_equal`.

---

But to elaborate on the `Result` thing: think about what you might want to say about what went wrong to a user if they're *not* all equal.

If the method returned, say `Result<Item, Option<(Item, Item)>>`, then you could imagine
```rust
match r {
    Ok(x) => ... use x ...,
    Err(None) => "There weren't any items",
    Err(Some((a, b))) => format!("There were multiple distinct items, including at least {a} and {b}"),
}
```

With just a `None` you can't say something so useful.  And it's no more expensive to give out those items, because by the time you'd be able to return a `false`/`None` you have to have found the mismatched item.

Byron:
Thanks for elaborating, now I understand.

Great, thanks everyone!

Byron:
I have undone the previous attempt to reduce duplication as it breaks tests. Given the small amount of code I believe that's acceptable.

scottmcm:
Oh, I see.  `empty().all_equal()` is `true`, but `empty().all_equal_item()` is `None`.

I see why both of them are that way, but the difference is unfortunate.  It makes me second-guess what I previously thought was a great name 🙁

Byron:
I think the name is perfect, as

- it's discoverable with the method providing more information having the same prefix.
- it's not at all surprising that if there is no item, it must return `None`

I am sure there was a good reason to default `all_equal()` to true if there is no item at all, and think the problem is rather that both methods lack a way of differentiating the 'empty()` case. Doing so would allow the caller to decide what `all_equal()` means if there is no item at all.

Maybe something like `all_items_equal()` isn't a companion of `all_items()`, but would rather have a new companion which allows differentiating the empty case? Or should `all_items()` rather default to `false` if there is nothing to compare to in the first place? Or should there be simply another variant of this method that returns false then, making it the elusive companion I was talking about?

Seeing all the text above makes me smile a bit as I didn't expect this topic to be as complex as it turned out to be, thanks for luring it out of hiding, it's a conversation definitely worth having :)!

PR #602, opened 2022-02-07
Fix tuple_windows as lazy

### LINKED ISSUE #601: Laziness in tuple_windows
I was bit surprised by that `Itertools::tuple_windows` is not somewhat lazy and has side effect.
It is consuming the first tuple when calling `Itertools::tuple_windows` .

https://doc.rust-lang.org/nightly/core/iter/index.html#laziness says that adapters in `Iterator` is lazy and 
the `TupleWindows` has `#[must_use = "iterator adaptors are lazy and do nothing unless consumed]` since #227

So the question is
- Is this intended behavior?  
- Should (all) adapters in `Itertools` be lazy?
- Should the laziness be documented?

aobatact:
Fix #601 anyway because this change will reduce one clone.

### DIFF
--- a/src/size_hint.rs
+++ b/src/size_hint.rs
@@ -29,7 +29,6 @@ pub fn add_scalar(sh: SizeHint, x: usize) -> SizeHint {
 }
 
 /// Subtract `x` correctly from a `SizeHint`.
-#[cfg(feature = "use_alloc")]
 #[inline]
 pub fn sub_scalar(sh: SizeHint, x: usize) -> SizeHint {
     let (mut low, mut hi) = sh;
--- a/src/tuple_impl.rs
+++ b/src/tuple_impl.rs
@@ -167,7 +167,7 @@ where
     I: Iterator<Item = T::Item>,
     T: HomogeneousTuple,
 {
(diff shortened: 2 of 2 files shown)


jswrenn (changes requested):
I've rebased and rustfmt'd this PR, but it's still failing a test. Can you fix the bug?

Philippe-Cholet:
This is incompatible with my recent `size_hint` and very incompatible with my `ExactSizeIterator` implementation.

Philippe-Cholet:
I deleted my previous comment, it was wrong (I must have been sleeping ^^). We would definitely need to update `size_hint` but there is no issue about `ExactSizeIterator` (the resulting iterator being shorter than the inner one, not longer 😪). Probably

    use crate::size_hint::{self, SizeHint};

    // then
        fn size_hint(&self) -> SizeHint {
            // the current comment should be deleted
            size_hint::sub_scalar(self.iter.size_hint(), T::num_items() - 1)
        }

Philippe-Cholet:
@jswrenn This is ready for me, what do you think?

jswrenn (approved):

PR #606, opened 2022-03-10
Add collect_vec_option

amitagrawalcs:
collect_vec_option() return None if iterator is empty, otherwise Some.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,5 +1,5 @@
 #![warn(missing_docs)]
-#![crate_name="itertools"]
+#![crate_name = "itertools"]
 #![cfg_attr(not(feature = "use_std"), no_std)]
 
 //! Extra iterator adaptors, functions and macros.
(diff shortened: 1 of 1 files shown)


scottmcm:
Note that an empty `Vec` does not allocate, so this is non-orthogonal for no good reason.  If you want to do this, I suggest you make a `.none_if_empty()` extension method instead -- then you can `.collect_vec().none_if_empty()`, but *also* can use it in places that aren't a `collect`.  Or just write it with `Option::filter`: `Some(iter.collect_vec()).filter(|x| !x.is_empty())`.

More generally, I'm typically opposed to this operation because it adds a state that needs to be dealt with, but isn't supposed to happen: the `Some(vec![])`.  I elaborate more about it in <https://users.rust-lang.org/t/collecting-non-option-items-into-option-vec/68276/6?u=scottmcm>, but in short this really wants to be `Vec<T> -> Option<NonEmptyVec<T>>`, to avoid introducing that extra state.

jswrenn:
Thanks for the contribution! Unfortunately, I'm going to decline this PR. I'm in agreement with @scottmcm's assessment.

PR #608, opened 2022-03-25
Make Format/FormatWith use Cell

### LINKED ISSUE #607: FormatWith should use a Cell
The `RefCell` is unnecessary, as all we're doing is `.take()`, which is also available on `Cell`: https://doc.rust-lang.org/stable/std/cell/struct.Cell.html#method.take

https://github.com/rust-itertools/itertools/blob/943494c2be3977a42df701dfde90d903ae17d16d/src/format.rs#L14

This results in using more memory and generating more code than necessary.

Ten0:
Resolves #607

### DIFF
--- a/src/format.rs
+++ b/src/format.rs
@@ -1,5 +1,5 @@
+use std::cell::Cell;
 use std::fmt;
-use std::cell::RefCell;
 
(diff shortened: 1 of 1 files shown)


jswrenn (approved):
bors r+

Issue #223, opened 2017-09-25
The cutting block: These features will be removed from itertools eventually


bluss:
- [x] `.foreach()` -- when `Iterator::for_each` stabilizes and it's old enough so that the deprecation is not annoying. Introduced in Rust 1.21. **Deprecated in 0.8**
- [x] `.step()` #133 -- when `Iterator::step_by` stabilizes rust-lang/rust/issues/27741 Rust 1.28 **Deprecated in 0.8**
- [x] `.fold_while()` when `Iterator::try_fold` is stable Rust 1.27 **Deprecated in 0.8**
- [x] `repeat_call` when `repeat_with` is stable rust-lang/rust/pull/48156 Rust 1.28  **Deprecated in 0.8**
- [x] `.flatten()` when `Iterator::flatten` is stable #262  Rust 1.29 **Removed in 0.8**
- [x] `unfold()` when `std::iter::unfold` is stable rust-lang/rust/issues/55977

flxo:
Looks like `.flatten()` hit [stable](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.flatten) with 1.29.

DevinR528:
Seems like `unfold` was merged in as `from_fn`, issue may now be close-able?

PR #610, opened 2022-04-12
Remove unfold

density:
Removes `unfold`, which is no longer need since [from_fn](https://doc.rust-lang.org/core/iter/fn.from_fn.html) is stable.

Refs #223


jswrenn:
I'd be happy to merge a PR adding a deprecation notice, but would rather not remove `unfold` outright at this time. Their signatures are not identical, and so it's completely trivial to switch over to `from_fn`.

PR #613, opened 2022-05-04
Minmax set (prev "Add {min,max}_set(_by{_key)?)? functions")

phimuemue:
This PR supersedes https://github.com/rust-itertools/itertools/pull/323 (I did not know how I could amend to the original PR, so sorry for the "duplicate" PR here.)

The original PR by @zayenz looked rather good. I adjusted the stuff that came up during discussion back then:

* Teturn type `Vec` instead of `Option` - emptiness is sufficiently well represented by `Vec`.
* Functions require `Ord` instead of `PartialOrd` - just as `Iterator::min`.
* Avoid duplicate calls to `lt` by accepting a `FnMut(...)->Ordering` - seems canonical compared to the `bool`-solution.
* Use internal iteration instead of a manual `for`-loop.

Moreover, I simplified some bits.

### DIFF
--- a/src/extrema_set.rs
+++ b/src/extrema_set.rs
@@ -1,0 +1,42 @@
+/// Implementation guts for `min_set`, `min_set_by`, and `min_set_by_key`.
+pub fn min_set_impl<I, K, F, L>(mut it: I,
+                                mut key_for: F,
+                                mut lt: L) -> Option<Vec<I::Item>>
+    where I: Iterator,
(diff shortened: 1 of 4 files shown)


phimuemue:
@jswrenn @zayenz If there are no objections, I'd merge this into itertools.

zayenz:
This looks great to me and thanks for taking over and fixing it.

jswrenn (approved):

PR #615, opened 2022-05-23
Avoid warning

phimuemue:
Fixes two warnings that would clutter output.

* `#[inline]` not meaningful before macro call. I pulled it into the macro, but with a TODO asking whether this is sensible. As far as I know, we do not have an established guideline for this, so I thought it's ok to leave it in there.
* `#[must_use]` not meaningful on type alias. Pulled to the original type where not present before.


phimuemue:
bors +r

PR #616, opened 2022-05-23
Add `take_until` method

### LINKED ISSUE #597: `take_until` and `map_until`
A while ago there was a proposal by Jon Gjengset (`jonhoo`) to add `take_until` into Rust's standard library:
https://github.com/rust-lang/rust/issues/62208

While it may be not a good fit for the standard library, it might be a worthwhile addition to `itertools`.

As `michaelsproul` pointed out in the discussion,

> I don't think take_while_ref or peeking_take_while are well suited to this because they require you to assign the iterator to a local variable, and muck around with mutation, which breaks the flow of the iterator chain. The best I could do with them was this, which isn't as nice as take_until.
>
>If you don't mind, I'd be happy to try and get your code into itertools, giving you credit of course! (Co-authored-by on the commit)

e-rhodes:
Adds `take_until` method that returns elements as long as a predicate is satisfied, but including the first element that doesn't satisfy the predicate.

First discussed addition to std in <https://github.com/rust-lang/rust/issues/62208>.

Closes #597.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1411,8 +1411,27 @@ pub trait Itertools : Iterator {
     /// use itertools::Itertools;
     ///
     /// let items = vec![1, 2, 3, 4, 5];
-    /// let filtered: Vec<i32> = items.into_iter().take_until(|&n| n % 3 != 0).collect();
+    /// let filtered: Vec<_> = items.into_iter().take_until(|&n| n % 3 != 0).collect();
+    ///
     /// assert_eq!(filtered, vec![1, 2, 3]);
+    /// ```
+    ///
+    /// ```rust
+    /// use itertools::Itertools;
+    /// let items = vec![1, 2, 3, 4, 5];
+    ///
+    /// let take_until_result: Vec<_> = items
+    ///     .clone()
+    ///     .into_iter()
+    ///     .take_until(|&n| n % 3 != 0)
+    ///     .collect();
+    /// let take_while_result: Vec<_> = items
+    ///     .into_iter()
+    ///     .take_while(|&n| n % 3 != 0)
+    ///     .collect();
+    ///
+    /// assert_eq!(take_until_result, vec![1, 2, 3]);
+    /// assert_eq!(take_while_result, vec![1, 2]);
     /// ```
     ///
     /// ```rust
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1391,24 +1391,22 @@ pub trait Itertools : Iterator {
         adaptors::take_while_ref(self, accept)
     }
 
-    /// An iterator adaptor that consumes elements while the given predicate is
-    /// true, including the element for which the predicate first returned
-    /// false.
+    /// Returns an iterator adaptor that consumes elements while the given
+    /// predicate is `true`, *including* the element for which the predicate
+    /// first returned `false`.
     ///
     /// The [`.take_while()`][std::iter::Iterator::take_while] adaptor is useful
     /// when you want items satisfying a predicate, but to know when to stop
     /// taking elements, we have to consume that last element that doesn't
-    /// satisfy the predicate. This adaptor simply includest that element where
+    /// satisfy the predicate. This adaptor simply includes that element where
     /// [`.take_while()`][std::iter::Iterator::take_while] would drop it.
     ///
     /// The [`.take_while_ref()`][crate::Itertools::take_while_ref] adaptor
-    /// serves a similar purpose, but this adaptor doesn't require cloning the
-    /// underlying elements.
-    ///
-    /// # Examples
+    /// serves a similar purpose, but this adaptor doesn't require [`Clone`]ing
+    /// the underlying elements.
     ///
     /// ```rust
-    /// use itertools::Itertools;
+    /// # use itertools::Itertools;
     ///
     /// let items = vec![1, 2, 3, 4, 5];
     /// let filtered: Vec<_> = items.into_iter().take_until(|&n| n % 3 != 0).collect();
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -1389,6 +1391,74 @@ pub trait Itertools : Iterator {
...
+    ///     .into_iter()
+    ///     .map(NoCloneImpl)
+    ///     .collect();
+    /// let filtered: Vec<_> = non_clonable_items
+    ///     .into_iter()
+    ///     .take_while_inclusive(|n| n.0 % 3 != 0)
+    ///     .collect();
+    /// let expected: Vec<_> = vec![1, 2, 3].into_iter().map(NoCloneImpl).collect();
+    /// assert_eq!(filtered, expected);
+    fn take_while_inclusive<F>(&mut self, accept: F) -> TakeWhileInclusive<Self, F>
```
pickfire, line 1454: As suggested in the original pull github issue in rust, wouldn't something like `take_until` be easier to find?

e-rhodes: I don't think so. The motivation for this method is "`take_while` but keeping the last element". `take_until` doesn't imply anything about the behavior with the final element—the only thing it implies is that the semantics of the predicate are reversed, e.g.:

```rust
// take elements while they're less than 4
iter.take_while(|x| x < 4)
// take elements until one is greater than or equal to 4
iter.take_until(|x| x >= 4)
```

If I just read this code without reading any docs, I would probably assume these did the same thing, ond certainly wouldn't guess the motivation of having both methods. The negation becomes the most noticeable difference and distracts from the actual purpose. 

Plus, if I see `take_while` in std and I'm looking for a method that's almost that but lets me keep the last element, I'm much more likely to find `take_while_inclusive` than `take_until`. And I'd be annoyed to have to negate the predicate if I was converting one to the other.

scottmcm: I'm definitely used to `until COND` being shorthand for `while !COND`, like with <https://www.perltutorial.org/perl-until/>, not implying another iteration or similar.

smessmer:
Are there plans to merge this?

jswrenn:
Thanks for the ping, @smessmer. I'll give this a final round of review tomorrow morning, and merge if everything looks good.

src/take_while_inclusive.rs
```diff
@@ -0,0 +1,68 @@
...
+{
+    debug_fmt_fields!(TakeWhileInclusive, iter);
+}
+
+impl<'a, I, F> Iterator for TakeWhileInclusive<'a, I, F>
+where
+    I: Iterator,
+    F: FnMut(&I::Item) -> bool
+{
+    type Item = I::Item;
```
scottmcm, line 39: pondering: Would it be worth returning the result of the predicate in some way in the item type, since that information is known free anyway?

I was reminded of [`Position::Last`](https://docs.rs/itertools/latest/itertools/enum.Position.html#variant.Last), though it shouldn't be that type (since this doesn't know the other variants inherently), and I suppose one could always `.take_while_inclusive(…).with_position()` if they wanted.

My brain then thought about things like `Result<I::Item, I::Item>` so the last item is an `Err`.  But that makes me go one step further and think of a `take_while_map_inclusive` (hopefully with a better name!) where the predicate returns a `Result` (or a `ControlFlow`) instead of a `bool`, and stops after returning the `Err` (or the `Break`).

Dunno if any of that is useful or worth doing, but I figured I'd toss it out for discussion.

smessmer: Returning a custom type that Deref's to the actual item type but has additional information about the predicate could be a good idea.

Otherwise, the call site might be able to infer it:
All entries except the last one in the iterator had the predicate return true, and the last had the predicate return false, with the following exceptions:
- If the returned iterator is empty, then there is no last element that returned true.
-  If the returned iterator has the same length as the original iterator, then there is no last element that returned true

jswrenn:
@e-rhodes Have you tried submitting this as an addition to `Iterator`? As a rule, we avoid accepting proposals that could be suitable additions to `Iterator`, because such additions can later lead to conflicts between `Itertools` and `Iterator` (see the history of `flatten` and `intersperse` for examples of such conflicts).

e-rhodes:
@jswrenn In the rust-lang issue I linked, I got the impression that it wasn't deemed useful enough to be added to Iterator, and it was suggested that we add it to itertools instead (see [here](https://github.com/rust-lang/rust/issues/62208#issuecomment-613246041)). 

If you get a different impression and think it's likely to get accepted over there I'm happy to pr!

jswrenn:
Hm, that comment wasn't made by a member of the libs team. I've asked on Zulip whether the libs team is interested: https://rust-lang.zulipchat.com/#narrow/stream/219381-t-libs/topic/Status.20of.20.2362208/near/311918204

jswrenn:
It sounds like the next step is to [create an ACP](https://std-dev-guide.rust-lang.org/feature-lifecycle/api-change-proposals.html) for this method. Would you like to tackle that, @e-rhodes?

src/take_while_inclusive.rs
scottmcm: Oh, I should have read the Iterator docs better.  It [looks](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.map_while) like what I propose would be named `map_while_inclusive` to go with `take_while_inclusive`.

I'll add that to the ACP, if one gets opened.

e-rhodes:
@jswrenn sure! I'll take a look at it this weekend. Thanks for getting this started!

e-rhodes:
Created ACP here: https://github.com/rust-lang/libs-team/issues/142

Let me know if there's anything I should add/change!

tranzystorek-io:
Could use it for todays Advent Of Code 😛

jswrenn:
@tranzystorek-io, please let the libs team know! https://github.com/rust-lang/libs-team/issues/142

PR #618, opened 2022-06-02
add rustdoc and clippy to github workflow

hellow554:
This commit is quite a mouth full and I'm can understand if you want me to split this into smaller chunks.

Let me start explain it commit by commit.

First I created a clippy toml with the msrv set to 1.36.0, which is the version used in the [github workflow](https://github.com/rust-itertools/itertools/blob/1f0a8c1c9cd43f45764fb387386ccdce49aeb535/.github/workflows/ci.yml#L16).

Next I either fixed all the linted lines, or added a allow with a comment to it.  
I did the same for all the tests.

Next I saw, that there is currently a rustfmt which skips the complete source base, based on a commit two years ago: https://github.com/rust-itertools/itertools/commit/641671e8b8b8737a38039d572022e38b9b43b8d5

*"Temporarily disable"* 😉 

I then reformatted the whole codebase, which might break some PRs but they should be able to rebase it but just using `cargo fmt` easily.

Next, I fixed two warnings. One stray \`\`\` and the other one being changed in https://github.com/rust-lang/rust/pull/96676 where you have to specify a macro link.

Last, but not least, I added `rustfmt` `rustdoc` and `clippy` to the `ci.yml` to enable them by default on PRs.

---

Feel free to comment whatever you think. I try to answer or change it :)

Thanks for the awesome crate :)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,13 @@ jobs:
           components: clippy
       - run: RUSTFLAGS="--deny warnings" cargo clippy ${{ matrix.features }}
 
+  doc:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+      - run: RUSTDOCFLAGS="-Dwarnings" cargo doc --all-features
+
   msrv:
     runs-on: ubuntu-latest
     env:
--- a/src/adaptors/mod.rs
+++ b/src/adaptors/mod.rs
@@ -218,7 +218,7 @@ where
     /// Split the `PutBack` into its parts.
     #[inline]
     pub fn into_parts(self) -> (Option<I::Item>, I) {
-        let PutBack { top, iter } = self;
+        let Self { top, iter } = self;
         (top, iter)
     }
 
@@ -689,7 +689,7 @@ pub struct Tuple1Combination<I> {
 
 impl<I> From<I> for Tuple1Combination<I> {
     fn from(iter: I) -> Self {
-        Tuple1Combination { iter }
+        Self { iter }
     }
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1910,12 +1910,7 @@ pub trait Itertools: Iterator {
     where
         P: FnMut(&Self::Item) -> bool,
     {
-        for (index, elt) in self.enumerate() {
-            if pred(&elt) {
-                return Some((index, elt));
-            }
-        }
-        None
+        self.enumerate().find(|x| pred(&x.1))
     }
     /// Find the value of the first element satisfying a predicate or return the last element, if any.
     ///
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -4076,15 +4076,15 @@ impl<T> FoldWhile<T> {
     /// Return the value in the continue or done.
     pub fn into_inner(self) -> T {
         match self {
-            FoldWhile::Continue(x) | FoldWhile::Done(x) => x,
+            Self::Continue(x) | Self::Done(x) => x,
         }
     }
 
     /// Return true if `self` is `Done`, false if it is `Continue`.
     pub fn is_done(&self) -> bool {
         match *self {
-            FoldWhile::Continue(_) => false,
-            FoldWhile::Done(_) => true,
+            Self::Continue(_) => false,
+            Self::Done(_) => true,
         }
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2124,8 +2124,7 @@ pub trait Itertools: Iterator {
     /// ```
     fn dropping_back(mut self, n: usize) -> Self
     where
-        Self: Sized,
-        Self: DoubleEndedIterator,
+        Self: Sized + DoubleEndedIterator,
     {
         if n > 0 {
             (&mut self).rev().nth(n - 1);
@@ -3962,7 +3961,7 @@ pub trait Itertools: Iterator {
     }
 }
 
-impl<T: ?Sized> Itertools for T where T: Iterator {}
+impl<T> Itertools for T where T: Iterator + ?Sized {}
 
 /// Return `true` if both iterables produce equal sequences
 /// (elements pairwise equal and sequences of the same length),
--- a/src/tuple_impl.rs
+++ b/src/tuple_impl.rs
@@ -54,6 +54,7 @@ where
         }
     }
 
+    #[allow(clippy::unnecessary_lazy_evaluations)] // buffer.len is more complex than usual
     fn size_hint(&self) -> (usize, Option<usize>) {
         let buffer = &self.buf.as_ref()[self.cur..];
         let len = if buffer.is_empty() {
--- a/src/tuple_impl.rs
+++ b/src/tuple_impl.rs
@@ -34,7 +34,7 @@ where
     T: HomogeneousTuple,
 {
     fn new(buf: T::Buffer) -> Self {
-        TupleBuffer { cur: 0, buf }
+        Self { cur: 0, buf }
     }
 }
 
--- a/src/tuple_impl.rs
+++ b/src/tuple_impl.rs
@@ -245,7 +245,7 @@ where
 /// information.
 #[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
 #[derive(Debug, Clone)]
-pub struct CircularTupleWindows<I, T: Clone>
+pub struct CircularTupleWindows<I, T>
 where
     I: Iterator<Item = T::Item> + Clone,
     T: TupleCollect + Clone,
--- a/src/unique_impl.rs
+++ b/src/unique_impl.rs
@@ -110,9 +110,10 @@ where
             if let Entry::Vacant(entry) = used.entry(v) {
                 let elt = entry.key().clone();
                 entry.insert(());
-                return Some(elt);
+                Some(elt)
+            } else {
+                None
             }
-            None
         })
     }
 
--- a/src/unique_impl.rs
+++ b/src/unique_impl.rs
@@ -1,3 +1,8 @@
+// Use a Hashmap for the entry API in order to prevent hashing twice.
+// This can maybe be replaced with e a HashSet once `get_or_insert_with`
+// // or a proper Entry API for Hashset is stable and meets the msrv
+#![allow(clippy::zero_sized_map_values)]
+
 use std::collections::hash_map::Entry;
 use std::collections::HashMap;
 use std::fmt;
@@ -11,9 +16,7 @@ use std::iter::FusedIterator;
 #[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
 pub struct UniqueBy<I: Iterator, V, F> {
     iter: I,
-    // Use a Hashmap for the Entry API in order to prevent hashing twice.
-    // This can maybe be replaced with a HashSet once `get_or_insert_with`
-    // or a proper Entry API for Hashset is stable and meets this msrv
+    // see comment for `allow(clippy::zero_sized_map_values)` for reasoning
     used: HashMap<V, ()>,
     f: F,
 }
--- a/src/ziptuple.rs
+++ b/src/ziptuple.rs
@@ -39,8 +39,7 @@ pub struct Zip<T> {
 /// [`izip!()`]: crate::izip
 pub fn multizip<T, U>(t: U) -> Zip<T>
 where
-    Zip<T>: From<U>,
-    Zip<T>: Iterator,
+    Zip<T>: From<U> + Iterator,
 {
     Zip::from(t)
 }
--- a/tests/quick.rs
+++ b/tests/quick.rs
@@ -38,7 +38,7 @@ impl HintKind for Exact {
 
 impl qc::Arbitrary for Exact {
     fn arbitrary<G: qc::Gen>(_: &mut G) -> Self {
-        Exact {}
+        Self {}
     }
 }
 
@@ -69,7 +69,7 @@ impl qc::Arbitrary for Inexact {
         // Compensate for quickcheck using extreme values too rarely
         let ue_choices = &[0, ue_value, usize::max_value()];
         let oe_choices = &[0, oe_value, usize::max_value()];
-        Inexact {
+        Self {
             underestimate: *ue_choices.choose(g).unwrap(),
             overestimate: *oe_choices.choose(g).unwrap(),
         }
@@ -79,7 +79,7 @@ impl qc::Arbitrary for Inexact {
         let underestimate_value = self.underestimate;
         let overestimate_value = self.overestimate;
         Box::new(underestimate_value.shrink().flat_map(move |ue_value| {
-            overestimate_value.shrink().map(move |oe_value| Inexact {
+            overestimate_value.shrink().map(move |oe_value| Self {
                 underestimate: ue_value,
                 overestimate: oe_value,
             })
@@ -108,7 +108,7 @@ where
     HK: HintKind,
 {
     fn new(it: Range<T>, hint_kind: HK) -> Self {
-        Iter {
+        Self {
             iterator: it,
             fuse_flag: 0,
             hint_kind,
@@ -166,16 +166,16 @@ where
     HK: HintKind,
 {
     fn arbitrary<G: qc::Gen>(g: &mut G) -> Self {
-        Iter::new(T::arbitrary(g)..T::arbitrary(g), HK::arbitrary(g))
+        Self::new(T::arbitrary(g)..T::arbitrary(g), HK::arbitrary(g))
     }
 
-    fn shrink(&self) -> Box<dyn Iterator<Item = Iter<T, HK>>> {
+    fn shrink(&self) -> Box<dyn Iterator<Item = Self>> {
         let r = self.iterator.clone();
         let hint_kind = self.hint_kind;
         Box::new(r.start.shrink().flat_map(move |a| {
             r.end
                 .shrink()
-                .map(move |b| Iter::new(a.clone()..b, hint_kind))
+                .map(move |b| Self::new(a.clone()..b, hint_kind))
         }))
     }
 }
@@ -231,7 +231,7 @@ where
         let iter_count = g.gen_range(0, MAX_ITER_COUNT + 1);
         let hint_kind = qc::Arbitrary::arbitrary(g);
 
-        ShiftRange {
+        Self {
             range_start,
             range_end,
             start_step,
@@ -1307,14 +1307,14 @@ quickcheck! {
 #[derive(Clone, Debug, PartialEq, Eq)]
 struct Val(u32, u32);
 
-impl PartialOrd<Val> for Val {
-    fn partial_cmp(&self, other: &Val) -> Option<Ordering> {
+impl PartialOrd<Self> for Val {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
         Some(self.cmp(other))
     }
 }
 
 impl Ord for Val {
-    fn cmp(&self, other: &Val) -> Ordering {
+    fn cmp(&self, other: &Self) -> Ordering {
         self.0.cmp(&other.0)
     }
 }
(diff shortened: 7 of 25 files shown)


tests/quick.rs
```diff
@@ -258,12 +258,12 @@ where
         let mut it = get_it();
 
         for _ in 0..(counts.len() - 1) {
-            if let None = it.next() {
+            if it.next().is_none() {
                 panic!("Iterator shouldn't be finished, may not be deterministic");
             }
```
phimuemue, line 256: When we're at it: Shouldn't this be an `assert!(it.next().is_none(), "Iterator shouldn't be finished, may not be deterministic")`?

.github/workflows/ci.yml
```diff
@@ -24,9 +24,12 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: dtolnay/rust-toolchain@stable
+      - run: cargo fmt --all -- --check
```
phimuemue, line 27: Does this imply that non-`cargo fmt`'d PRs are rejected? If so, I would demote this to at most a warning.

.rustfmt.toml
```diff
@@ -1,3 +0,0 @@
-# Temporarily disable rustfmt completely to avoid conflicts of newly formatted
-# code with old PRs.
-ignore = ["/"]
```
phimuemue, line 3: If we really enable `rustfmt`, I'd prefer deleting this file in a separate commit. (So that the commit could easily be undone without affecting anything else.)

phimuemue (commented):
Wohoo, there's some really nice improvements in there! Maybe you can rebase against the most recent git, as some of the things should already be fixed.

In addition, may I ask you for a favor that would imho simplify reviewing quite a bit:

* Taking up your idea: Could you split all the lints into separate commits? Sure, it'd be more commits, but it avoids the constant context switching. In addition, we could theoretically cherry-pick the ones that are already good-to-go, while leaving others for later.
* Possibly postpone the `cargo fmt`-related stuff to an own PR (after all the other fixes arrived) - or at least do it as the last commit so the first steps could be cherry-picked. 
  
  We once had the discussion `cargo fmt`ing everything, and we plan to do it eventually, but it would probably break virtually all existing PRs - something we do not want to do lightly. 

  In addition, I am still reluctant to reject PRs that do not adhere to `cargo fmt`: I think it would be unfortunate if a contributor would have to rework stuff because of a `x+1` vs `x + 1`. However, I'd still be happy about an automated solution that would commit the `cargo fmt` as a PR automatically if it observes that it is required. (See https://github.com/rust-itertools/itertools/pull/421 for more discussion.)

tests/quick.rs
hellow554: `is_some` but yeah. Strange that clippy didn't lint this. I will investigate it :)

src/lib.rs
```diff
@@ -2750,7 +2750,6 @@ pub trait Itertools: Iterator {
     /// itertools::assert_equal(oldest_people_first,
     ///                         vec!["Jill", "Jack", "Jane", "John"]);
     /// ```
-    /// ```
```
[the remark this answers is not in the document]
hellow554: Not sure what you mean. That's already a seperate commit together with the `izip` fix: https://github.com/rust-itertools/itertools/pull/618/commits/86f1f278fa15682d5beef2c35a1a99e7b0023f70

src/ziptuple.rs
```diff
@@ -36,6 +36,7 @@ pub struct Zip<T> {
 ///
 /// assert_eq!(results, [0 + 3, 10 + 7, 29, 36]);
 /// ```
+/// [`izip!()`]: crate::izip
```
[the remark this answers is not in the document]
hellow554: Not sure what you mean. That's already a seperate commit together with the \`\`\` fix: https://github.com/rust-itertools/itertools/pull/618/commits/86f1f278fa15682d5beef2c35a1a99e7b0023f70

hellow554:
So this exploded a bit. It went from 5 to 31 commits, but yeah, it may be easier for you to review, I can understand that.

I removed the fmt part completly and will open a different PR, so no worries about that anymore.

src/lib.rs
```diff
@@ -1748,12 +1748,10 @@ pub trait Itertools : Iterator {
     fn find_position<P>(&mut self, mut pred: P) -> Option<(usize, Self::Item)>
         where P: FnMut(&Self::Item) -> bool
     {
-        let mut index = 0usize;
-        for elt in self {
+        for (index, elt) in self.enumerate() {
             if pred(&elt) {
                 return Some((index, elt));
             }
-            index += 1;
```
phimuemue, line 1756: I saw places where the compiler generated different code for `enumerate` vs an explicit counter loop. I do not know whether one of them was strictly more performant.

I would accept this, unless there are objections. (@jswrenn - do you know something?)

src/tuple_impl.rs
```diff
@@ -63,7 +63,7 @@ impl<T> Iterator for TupleBuffer<T>
         } else {
             buffer.iter()
                   .position(|x| x.is_none())
-                  .unwrap_or_else(|| buffer.len())
+                  .unwrap_or(buffer.len())
```
phimuemue, line 66: Out of curiosity: Was this suggested by `clippy`? If not, I'd stay with the original

tests/quick.rs
```diff
@@ -258,9 +258,7 @@ where
         let mut it = get_it();
 
         for _ in 0..(counts.len() - 1) {
-            if it.next().is_none() {
-                panic!("Iterator shouldn't be finished, may not be deterministic");
-            }
+            assert!(it.next().is_some(), "Iterator shouldn't be finished, may not be deterministic");
```
phimuemue, line 261: Thinking about this a second time raises the question: Is it problematic that `it.next` appears within the `assert`? (In my ideal world, an `assert` should be side-effect free.)

tests/test_core.rs
```diff
@@ -234,7 +234,7 @@ fn count_clones() {
     // Check that RepeatN only clones N - 1 times.
 
     use core::cell::Cell;
-    #[derive(PartialEq, Debug)]
+    #[derive(PartialEq, Eq, Debug)]
```
phimuemue, line 232: Depends. Might `Foo` be explicitly used to test stuff that only relies on `PartialEq` - and not `Eq`?

phimuemue (commented):
Cool! Thanks for your effort, it's appreciated.

I commented the places that jumped to my eye. (All others look good to me.) Maybe you could dispel my concerns about the few remaining bits.

src/lib.rs
scottmcm: This could be written to call `find` instead, right?  That might be a good way to avoid the concern, since `find` will then use internal iteration too, which should be just as good as the manual `index`, if not better.

Something like this, maybe:
```rust
self.enumerate().find(|x| pred(&x.1))
```

Or there's also the way that's more like how `position` works,
```rust
self.try_fold(0, |i, x| {
    if pred(&x) { ControlFlow::Break((i, x)) }
    else { ControlFlow::Continue(i + 1) }
}).break_value()
```
but I don't know if your MSRV can do that one -- it might not be better, either.

src/unique_impl.rs
```diff
@@ -56,7 +56,7 @@ impl<I, V, F> Iterator for UniqueBy<I, V, F>
     type Item = I::Item;
 
     fn next(&mut self) -> Option<Self::Item> {
-        while let Some(v) = self.iter.next() {
+        for v in self.iter.by_ref() {
```
scottmcm, line 59: Beware that `by_ref` is currently an optimization firewall, so you might not want this change.

See https://github.com/rust-lang/rust/pull/95602 for an example.

Since you want something `&mut self` (like `next` was) here, you could again consider moving the loop into the iterator method.  For example, I think you could phrase this as a `self.iter.find_map(...)` instead for the `for` loop.

src/adaptors/mod.rs
```diff
@@ -1032,7 +1032,7 @@ impl<I, F> Iterator for Positions<I, F>
     type Item = usize;
 
     fn next(&mut self) -> Option<Self::Item> {
-        while let Some(v) = self.iter.next() {
```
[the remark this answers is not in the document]
scottmcm: As I mentioned in the other comment, I suggest either not using `by_ref`, or switching to internal.

Something like
```rust
self.iter.find_map(|v| {
    let i = self.count;
    self.count = i + 1;
    (self.f)(v).then(i)
})
```
(Or some MSRV-compliant equivalent)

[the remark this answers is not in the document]
scottmcm: Alternatively, you could consider having `Positions` store an `iter::Enumerate<I>` instead of `I`+`usize`.

src/tuple_impl.rs
hellow554: Yes: https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_lazy_evaluations

But there's an open issue for that, because I find that strange too: https://github.com/rust-lang/rust-clippy/issues/8109

tests/quick.rs
hellow554: I can see your concern. In my world, side effects in `assert` is ok, in `debug_assert` it's bad (and there's a lint for that (written by me): https://rust-lang.github.io/rust-clippy/master/index.html#debug_assert_with_mut_call)

But I would understand if you would like to have that different. Please tell me.

tests/test_core.rs
hellow554: https://rust-lang.github.io/rust-clippy/master/index.html#derive_partial_eq_without_eq

.github/workflows/ci.yml
```diff
@@ -27,6 +27,8 @@ jobs:
       - run: cargo check --no-default-features
       - run: cargo check --no-default-features --features "use_alloc"
       - run: cargo test
+      - run: RUSTDOCFLAGS="-Dwarnings" cargo doc --all-features
+      - run: cargo clippy --all -- -Dwarnings
```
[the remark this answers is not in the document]
hellow554: Yes, it would be rejected. Same goes for `cargo doc` (and `cargo fmt` in the future as well).

Isn't that the point of having a CI, that the code is equally good (or bad)?

hellow554:
I try to resolve the issues.

I see some concerns and I think thats ok. Everything you think is not worth changing or might be bad can be annotated by `#[allow(clippy::xxx)]` along with a comment.  
That's totally fine, really.

I'm not a perfect rust developer, neither is clippy perfect at detecting whether a change is good or bad.  
But I think, clippy suggests a lot of useful lints worth checking at least. Let's try to solve this one together :)

src/tuple_impl.rs
phimuemue: I could not find authoritative reference that calling `len` is guaranteed to only read a field, so please revert this to `unwrap_or_else`, and possibly add a lint.

tests/quick.rs
phimuemue: Would be fine for me. @jswrenn Do you have objections?

.github/workflows/ci.yml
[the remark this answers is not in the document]
phimuemue: # tl;dr

I'd like to have both a low bar for (especially new) contributors and try to acchieve a certain quality in the code base. Thus:
* I'm in favor of rejecting wrong documentation, as the result is user-facing and should be correct.
* I'm unsure if rejecting PRs with lint problems is a good idea.
* I'm very skeptical to reject PRs because of formatting.

## cargo clippy

I assume the codebase will quickly converge to a lint-free state (especially after your changes are merged). Once in that state, avoiding new lint problems should be easy.

From my personal experience, I found it demotivating if a PR is rejected because of a clippy-false-positive (or one that does not matter in the circumstances at hand). I'd like to avoid this situation as far as possible. 

* For one-off contributors, requesting and waiting for changes might be just as much effort as just merging and fixing the lint, possibly indicating that `clippy` should only warn instead of reject the PR.
* I assume frequent contributors do not introduce many new lints. (They may have clippy running on their local machines, so are aware of lint problems from the start.)

Generally, I think I'd prefer to only issue a warning (thus requiring human interaction to resolve it) instead of plainly rejecting the PR, assuming that it does not generate too much work on the contributors and the maintainers. I would be open to have `clippy` reject PRs if it's easier/better to implement/maintain/work with/others really want it/etc, as it sometimes gives really useful advice.

## cargo fmt

I guess the main advantage of having a `cargo fmt`d code base is that contributors have an idea how our code is written. However, for this it should be enough to "`cargo fmt` from time to time" (read: after a successfully merged PR). As such, I'd prefer *not* to reject PRs because of formatting problems (again to avoid frustration) - and instead have the formatting problems fixed automatically by our CI.

phimuemue:
Hi. I took the liberty to merge the first batch of this. (See https://github.com/rust-itertools/itertools/pull/624 - sorry if this was a stupid idea, I'm a github noob.)

Could you rebase so that this PR is updated, and it's clear what remains to be reviewed?

hellow554:
I think that was a good idea, so no worries :)

Rebased.

I think at least https://github.com/rust-itertools/itertools/pull/618/commits/bf2c114c859023c16ead2d5ebb77d5063c4a91f9 should be pulled in as well, but we can do that later. Without it you may get a lot of false clippy suggestions.

clippy.toml
```diff
@@ -0,0 +1 @@
+msrv = "1.36.0"
```
phimuemue, line 1: Do we need some remark that this line has to be updated along other files? Apart from that, I'm fine with this.

phimuemue (commented):
I took the freedom to merge a second batch of this. 

@hellow554 Thanks you! If you find time, could you let us know if you want to further push things forward on this PR?

clippy.toml
hellow554: It depends. I can't find any word about the msrv, apart from the github workflow, so first step would be to mention it in the readme I guess?
Second: there's the [`rust-version`](https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field) field in the `Cargo.toml` file that can be used as well. A clippy PR is pending to parse that field in addition to the `clippy.toml` file: https://github.com/rust-lang/rust-clippy/pull/8774

As soon as that lands, the `clippy.toml` can be deleted again, so I don't think that that should be documented? 😅

phimuemue:
Hi again. If you still want to tackle this: Could you rebase and sqash the fixup into the original commits to avoid back-and-forth changes (and thus, simplify review)? And could you leave the `unwrap_or_else` in place, and just allow the lint?

hellow554:
Done :)

I very much thank you for your patience and effort you put into this! It's not easy to review this kind of stuff. Many, many thanks!

phimuemue:
> Done :)
> 
> I very much thank you for your patience and effort you put into this! It's not easy to review this kind of stuff. Many, many thanks!
> 
> Btw: I found many false-positives while using clippy, which helps all users, so many thanks for the many test cases ;)

My pleasure. I merged the third batch in https://github.com/rust-itertools/itertools/pull/630.

hellow554:
Rebased again. Take your time, I'm not in a hurry

.github/workflows/ci.yml
```diff
@@ -27,6 +27,8 @@ jobs:
       - run: cargo check --no-default-features
       - run: cargo check --no-default-features --features "use_alloc"
       - run: cargo test
+      - run: RUSTDOCFLAGS="-Dwarnings" cargo doc --all-features
+      - run: cargo clippy --all
```
danieleades, line 31: why not use a nightly target for clippy?

probably outside the scope of this PR, but why group CI targets by toolchain at all? Seems an odd way to slice it. Would be preferable to have this as separate, parallel jobs

danieleades:
The MSRV should be set in `Cargo.toml` these days, using the `rust-version` key.

.github/workflows/ci.yml
danieleades, line 31:
```suggestion
      - run: cargo clippy --all --all-targets
```

danieleades: see #675 for an example of what i mean

Philippe-Cholet:
I come here just after we added clippy to our CI (and rustfmt a while back). I did not know this PR but it has good content I agree with as I made some changes lately that were part of this without knowing it. Sadly, this was a too big PR to be merged in a reasonable short time and eventually got forgotten. Smaller focused PRs are better, I learnt that myself. I see it was a good discussion though.

I think we might want to salvage part of this (in other smaller PRs). About documentation maybe (`RUSTDOCFLAGS="-Dwarnings" cargo doc --all-features`) and resolve part of what `clippy` **`--all`** tell us. Am I forgotting something here?

Philippe-Cholet:
I first read your commits too fast as you were fixing clippy lints of lint groups (such as "pedantic/nursery") that our new CI job does not consider (and probably should not to not burden new contributors too much?).

I'm not gonna close this, it is a bit outdated but still valuable to cherry pick and update in more atomic PRs.

hellow554:
I rebased the work to the latest master. I'm very happy with your decision to leave this open and maybe cherry pick the things you like. That's totally fine to me.
Feel free to browse the commits.

Philippe-Cholet:
I kept `find_position` improvement (merged), `Eq` implementation for `MinMaxResult` (merged) and two annoying lints (PR sent), I'm gonna stop here though.

The only remaining question for me here before closing is about adding or not `RUSTDOCFLAGS="-Dwarnings" cargo doc --all-features` to the CI. I'm not familiar with `cargo doc` much yet. What kind of error would it prevent: wrong links? bad formatting?

> I'm in favor of rejecting wrong documentation, as the result is user-facing and should be correct. - @phimuemue

hellow554:
> What kind of error would it prevent: wrong links? bad formatting?

Correct, both of them. For examples of what this denies see https://doc.rust-lang.org/rustdoc/lints.html

Philippe-Cholet:
I was thinking wrong _markdown text_ formatting such as `This is **bold* text.` but fair enough, thanks for the insightful link.

jswrenn:
We should absolutely deny warnings in CI!

Philippe-Cholet:
I don't know to how many PRs this one lead to but it surely was a great base, thank you.

danieleades:
> We should absolutely deny warnings in CI! 

Only if the version of the linter in CI is pinned. It isn't currently.

Otherwise you can introduce breakages in contributor PRs that have nothing to do with the changes in the PR.

This should be done regardless of whether warnings are denied, in my opinion.

Philippe-Cholet:
@danieleades It sure seems a good idea (even if it should be a lot of breakages).
In [CI](https://github.com/rust-itertools/itertools/blob/master/.github/workflows/ci.yml), instead of "stable", use a pinned version such as `1.75.0` and update it from time to time?
I guess _dependabot_ should be able to help us. However, #758 wanted us to drop 1.43.1 (our msrv) so it's not that helping yet.

danieleades:
> @danieleades It sure seems a good idea (even if it should not be a lot of breakages).
> In [CI](https://github.com/rust-itertools/itertools/blob/master/.github/workflows/ci.yml), instead of "stable", use a pinned version such as `1.75.0` and update it from time to time?
> I guess _dependabot_ should be able to help us. However, #758 wanted us to drop 1.43.1 (our msrv) so it's not that helping yet.

I'd suggest pinning to a recent nightly version of clippy. The 'rust-version' (MSRV) config will prevent clippy from suggesting incompatible changes

mightyiam:
Once a while, a new stable rust toolchain is released, along with some new clippy lints, usually. Some of them are on by default. And some of those may trigger on our code. Usually, they are easy to resolve. And if they're not, we can allow on the item. Using "stable" rolling clippy means that in those cases, development will be blocked until they are resolved. That should motivate whomever wishes to see their PR merged to resolve the lint in a prior PR. Then, rebasing is easy. There could even be a button for it in PR pages (that's an option in the repo settings). Therefore I vote rolling stable clippy.

Philippe-Cholet:
(As a member,) I have access to a "rebase" button.
I also (for now) would choose small rare blocks over manual pin maintenance. We can still change our mind later. And if there was a big block, we can still temporarily pin it in the CI to resolve the block later.

danieleades:
actually i think if you use the dtolnay toolchain action to pin to a stable version dependabot will create PRs to bump the version.
That gives you the best of both worlds- linting is pinned and consistent across PRs, and dependabot automagically creates PRs which highlight any new lints or changes to be addressed as they are released.

mightyiam:
That sounds good, as well.

danieleades:
> That sounds good, as well.

see #846

PR #619, opened 2022-06-03
Do rustfmt on the codebase

hellow554:
This is somewhat a sequel of #421

I removed the rustfmt file which ignores all files and did `cargo fmt` on the codebase.

I know, that you have concerns about PRs that will get a merge conflict once this landed. 

I'm not sure about which ones we are talking since there are PRs from 2020 and older, which might have a conflict anyway.

So, what do people need to do: just run `cargo fmt` on their codebase, rebase it with master and everything should just work™.

This would help to have a consistent code style as well as a readable one in some places.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: dtolnay/rust-toolchain@1.36.0
+      - run: cargo fmt --all -- --check
       - run: cargo check --no-default-features
       - run: cargo check --no-default-features --features "use_alloc"
       - run: cargo check
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -1,3 +1,0 @@
-# Temporarily disable rustfmt completely to avoid conflicts of newly formatted
-# code with old PRs.
-ignore = ["/"]
--- a/benches/bench1.rs
+++ b/benches/bench1.rs
@@ -1,10 +1,10 @@
 use criterion::{black_box, criterion_group, criterion_main, Criterion};
-use itertools::Itertools;
 use itertools::free::cloned;
 use itertools::iproduct;
+use itertools::Itertools;
 
-use std::iter::repeat;
 use std::cmp;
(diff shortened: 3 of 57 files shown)


.github/workflows/ci.yml
```diff
@@ -14,6 +14,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: dtolnay/rust-toolchain@1.36.0
+      - run: cargo fmt --all -- --check
```
danieleades, line 17: would it not be better to add a separate CI target for rustfmt? and why use the MSRV for running the formatting?

something like-

```yaml
  fmt:
    name: format
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v3
      - uses: dtolnay/rust-toolchain@nightly
        with:
          components: rustfmt
      - uses: actions-rs/cargo@v1
        with:
          command: fmt
          args: --all -- --check
```

danieleades (changes requested):

danieleades:
I'm fully in support of this PR!

The best time to introduce formatting is yesterday. The second best time is today.

I don't think the original motivation of temporarily disabling rustfmt to avoid PR conflicts makes sense, unless you had some foreknowledge that future PRs were going to dry up for some reason. That's demonstrably not the case here. Instead, i think it's better to just rip the band-aid off.

.github/workflows/ci.yml
hellow554: not sure if introducing yet another toolchain is worth here. You can use the stable chain instead, which already exists.

But I'm down for using checkout@3 as well as dtolnay's rust-toolchain.
I have a weird connection with actions-rs, because some of their actions are no longer maintained, but cargo looks good to me, so yeah, that's worth a rewrite in the github workflow.

danieleades: > not sure if introducing yet another toolchain is worth here

in what sense? I would personally split these into separate targets even if they both used the stable toolchain

jswrenn:
I'm going to tackle this myself as part of #702.

PR #623, opened 2022-06-07
Replace `Hashmap<V, ()>` by `HashSet<V>`

hellow554:
The `Hashmap` was chosen because of the `Entry` API, but one can work around
that easiliy.


src/unique_impl.rs
```diff
@@ -133,10 +129,9 @@ impl<I> DoubleEndedIterator for Unique<I>
 {
     fn next_back(&mut self) -> Option<Self::Item> {
         while let Some(v) = self.iter.iter.next_back() {
-            if let Entry::Vacant(entry) = self.iter.used.entry(v) {
-                let elt = entry.key().clone();
-                entry.insert(());
-                return Some(elt);
+            if !self.iter.used.contains(&v) {
+                self.iter.used.insert(v.clone());
```
scottmcm, line 133: Yes, this code is easy to write, but it seems like it also loses the point of why the entry API was used in the first place -- this needs to hash the element twice (once in `contains`, once in `insert`), whereas the original code only has to hash it once (in `entry`).

Given that the `HashMap` vs `HashSet` doesn't leak out of the implementation, it seems to me like it's worth a slightly more complex implementation when it can save some hashing work.

hellow554: Alright, thanks for the review and your great knowledge.

I'll rework the PR to make the intention clear :)

PR #628, opened 2022-07-02
if use_std is enabled, also enable either/use_std

sunshowers:
I've been using either through itertools, and I'd love to have `either/use_std` enabled if `use_std` is enabled.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,7 @@ version = "0.2"
 
 [features]
 default = ["use_std"]
-use_std = ["use_alloc"]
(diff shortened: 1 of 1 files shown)


phimuemue (approved):
Hi there, I am (sadly) no expert on this, but after skimming [the docs](https://doc.rust-lang.org/cargo/reference/features.html#dependency-features), this looks *very* reasonable to me.

@jswrenn You're usually good at judging these things. Do you have objections?

jswrenn:
This sounds fine to me!

bors r+

PR #629, opened 2022-07-04
Improvements to `EitherOrBoth`

JoJoJet:
* Added more methods for accessing the left and right variants.
* Added methods for inserting left and right values (using unsafe code).
* Improved some grammar.

### DIFF
--- a/src/either_or_both.rs
+++ b/src/either_or_both.rs
@@ -300,6 +300,42 @@ impl<A, B> EitherOrBoth<A, B> {
         }
     }
 
+    /// Returns a mutable reference to the left value. If the left value is not present,
+    /// it is replaced with `val`.
+    pub fn left_or_insert(&mut self, val: A) -> &mut A {
+        self.left_or_insert_with(|| val)
+    }
+
+    /// Returns a mutable reference to the right value. If the right value is not present,
+    /// it is replaced with `val`.
+    pub fn right_or_insert(&mut self, val: B) -> &mut B {
+        self.right_or_insert_with(|| val)
+    }
+
+    /// If the left value is not present, replace it the value computed by the closure `f`.
+    /// Returns a mutable reference to the now-present left value.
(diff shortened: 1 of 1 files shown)


phimuemue:
Hi there, thanks for addressing the compilation issues.

Could you rebase the changes so we do not need to review back-and-forth changes?

JoJoJet:
> Could you rebase the changes so we do not need to review back-and-forth changes?

Done, I think.

src/either_or_both.rs
```diff
@@ -42,36 +44,107 @@ impl<A, B> EitherOrBoth<A, B> {
...
+    /// If `Both`, return `Some` containing the left and right values. Otherwise, return `None`.
     pub fn both(self) -> Option<(A, B)> {
         match self {
             Both(a, b) => Some((a, b)),
             _ => None,
         }
     }
 
+    /// If `Left` or `Both`, return the left value. Otherwise, convert the right value and return it.
+    pub fn into_left(self) -> A
```
oxalica, line 127: I prefer `left_biased`, which implies the fallback behavior on `Right`. So does `into_right` -> `right_biased`.
This `*_biased` wording is also used by `rowan`, the parser library of `rust-analyzer`, in an enum like our `EitherOrBoth` but allows empty.
https://docs.rs/rowan/0.15.11/rowan/enum.TokenAtOffset.html#method.left_biased

oxalica (changes requested):

src/either_or_both.rs
oxalica: > I don't think that the `*_biased` naming convention really captures what's going on here.

To my intuition, `left_biased` implies that "prefer Left, but fallback to return Right if there's no Left". `into_left` sounds like "we know there's a Left inside, just unwrap it" (like `into_inner`?).
But you're right, there's a conversion when `Right(_)` is encountered, and `into_left` captures this behavior.

jswrenn (approved):
bors r+

PR #634, opened 2022-07-11
Improved docstrings for `free::chain`

JoelMon:
Improved the description of the function and added example.

### DIFF
--- a/src/free.rs
+++ b/src/free.rs
@@ -128,16 +128,21 @@ pub fn zip<I, J>(i: I, j: J) -> Zip<I::IntoIter, J::IntoIter>
     i.into_iter().zip(j)
 }
 
-/// Create an iterator that first iterates `i` and then `j`.
+
+/// Takes two iterables and creates a new iterator over both in sequence. 
 ///
 /// [`IntoIterator`] enabled version of [`Iterator::chain`].
 ///
+/// ## Example
 /// ```
 /// use itertools::chain;
+/// 
+/// let mut result:Vec<i32> = Vec::new();
 ///
-/// for elt in chain(&[1, 2, 3], &[4]) {
-///     /* loop body */
+/// for element in chain(&[1, 2, 3], &[4]) {
+///     result.push(*element);
 /// }
+/// assert_eq!(result, vec![1, 2, 3, 4]);
 /// ```
 pub fn chain<I, J>(i: I, j: J) -> iter::Chain<<I as IntoIterator>::IntoIter, <J as IntoIterator>::IntoIter>
     where I: IntoIterator,


src/free.rs
```diff
@@ -128,17 +128,24 @@ pub fn zip<I, J>(i: I, j: J) -> Zip<I::IntoIter, J::IntoIter>
     i.into_iter().zip(j)
 }
 
-/// Create an iterator that first iterates `i` and then `j`.
+/// Returns an iterator combining two iterators consecutively.
```
phimuemue, line 131: Let's just adapt the [documentation from std](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.chain): "Takes two iterables and creates a new iterator over both in sequence."

src/free.rs
```diff
@@ -128,17 +128,24 @@ pub fn zip<I, J>(i: I, j: J) -> Zip<I::IntoIter, J::IntoIter>
...
 /// use itertools::chain;
+/// 
+/// let a = [1, 2, 3];
+/// let b = [4];
+/// let mut result:Vec<i32> = Vec::new();
 ///
-/// for elt in chain(&[1, 2, 3], &[4]) {
-///     /* loop body */
+/// for element in chain(&a , &b) {
+///     result.push(*element);
```
phimuemue, line 144: Please inline the arguments `a` and `b` into the call to `chain` (i.e. write `chain(&[1,2,3], &[4])`), so that it's visually similar to the resulting `vec![1,2,3,4]`.

src/free.rs
```diff
@@ -128,17 +128,24 @@ pub fn zip<I, J>(i: I, j: J) -> Zip<I::IntoIter, J::IntoIter>
...
 /// }
+/// assert_eq!(result, vec![1, 2, 3, 4]);
 /// ```
+#[deprecated(note="Use std::iter::Chain instead", since="1.0.0")]
```
phimuemue, line 148: Please remove this line, as `std::iter::Chain` refers to the `struct`, but not to a function (which does (not yet) exist).

phimuemue (commented):
Hi there, thanks for this - improvements to documentation is always good!

It would be nice if you could incorporate the suggestions (or refute them with good reason), and sqash your changes into a single commit.

src/free.rs
JoelMon: Made suggested changes.

src/free.rs
JoelMon: Implemented suggestion.

JoelMon:
I think I finally got this git squashing, force pushing stuff down. It's about time. 😆

src/free.rs
```diff
@@ -128,17 +128,23 @@ pub fn zip<I, J>(i: I, j: J) -> Zip<I::IntoIter, J::IntoIter>
     i.into_iter().zip(j)
 }
 
-/// Create an iterator that first iterates `i` and then `j`.
+
+/// Takes two iterators and creates a new iterator over both in sequence.
```
phimuemue, line 132: Could you change "iterators" to "iterables" (seems to be in line with what we have elsewhere).

src/free.rs
phimuemue: I'm not sure if your commits went in as you thought. I think we should not add the `deprecatted` notice.

phimuemue (changes requested):
Thanks for the changes. Could you have a second look there? With the suggested changes, I think we're good to go.

src/free.rs
JoelMon: Made the change

src/free.rs
JoelMon: While figuring out how to properly rebase and push I ended up having to revert a few times, this line must have snuck through. :+1:

JoelMon:
@phimuemue, would you be able to take a look at https://github.com/rust-itertools/itertools/pull/633 as well?

phimuemue:
Thanks a bunch!

bors r+

PR #637, opened 2022-08-12
new feature: partitions of a vector

### LINKED ISSUE #611: Feature Request: `partitions`
It would be neat to have a way of iterating over all possible partitions of some vector. Example usages:

`all_partitions(&vec![7,8,9])` should iterate over
```
[[7,8,9]]
[[7,8],[9]]
[[7,9],[8]]
[[7],[8,9]]
[[7],[8],[9]]
```

while `partitions(&vec![6,7,8,9])` should iterate over
```
[[6],[7,8,9]]
[[6,7],[8,9]]
[[6,8],[7,9]]
[[6,9],[7,8]]
[[6,7,8],[9]]
[[6,7,9],[8]]
[[6,8,9],[7]]
```

Remark: I asked this question on [StackOverflow](https://stackoverflow.com/questions/68682502/iterating-over-all-possible-partitions-of-a-given-set), and implemented a version of `all_partitions()`. I am happy to create a PR if this is something that could belong to the `itertools` package.

bruderjakob17:
This is a first draft resulting from the discussion in #611.

Closes #611.


skylerberg:
@bruderjakob17 and @phimuemue, this is a cool feature and I'm interested in using this feature! What's needed to get this ready to merge? I'd be happy to contribute anything needed.

src/partitions.rs
```diff
@@ -0,0 +1,74 @@
...
+        }
+        // if there is no incrementable index, we have enumerated the last possible partition.
+        return None;
+    }
+}
+/// Returns an Iterator over all partitions of the given Vec.
+/// Example usage:
+/// ```
+/// for partition in partitions(&vec![7,8,9]){
+///     println!("{:?}", partition);
```
phimuemue, line 69: Can you make the documentation to contain two or three examples that are fully expanded (e.g. by `assert_eq(iter.partitions(), vec![vec![1,2,3], vec![4], ...])`? Also, please document what happens when the iterator has duplicate elements (as far as I understand, they are treated as different elements, because we only operate on indices).

Maybe even add a note that the order of partitions is not guaranteed to stay the same over different itertools versions.

src/partitions.rs
```diff
@@ -0,0 +1,74 @@
...
+                self.index_map[index] += 1;
+                // set all following entries to the lexicographically smallest suffix that makes the
+                // index_map viable (see linked post), i.e. to zero.
+                for x in index + 1..self.index_map.len() {
+                    self.index_map[x] = 0;
+                }
+                return Some(self.create_partition());
+            }
+        }
+        // if there is no incrementable index, we have enumerated the last possible partition.
```
phimuemue, line 61: I suggest to change documentation to describe the algorithm high-level and completely remove the very-fine grained comments (such as "increment the incrementable index").

High-level algo:
* Partition represented as an index sequence: index_map[i]==p means: element with index i lands in the permutation class p. index_map[0]==0 always by convention.
* For an index sequence to describe a partition, for each prefix index_map[0..i], we must have each integer from 0 to max(index_map[0..i]) present (because a skipped integer would mean that we could assign a partition's class a lower number without changing the permutation itself).
* Adding a new element to an existing partition described by index_map[0..i] amounts to appending an index to index_map[0..i]: To construct new partitions, the new element can be added to each existing partition class (amounts to iterating index_map[i] from 0 to max(index_map[0..i])) or it can form a new partition class by itself (amounts to setting index_map[i] = max(index_map[0..i])+1).
* Thus, in each step, we look for the rightmost index that can be incremented according to the aforementioned rules, increment it, and set the indices after it to 0.
* (For more info, see [stackoverflow link]).

phimuemue (changes requested):
Thanks for this. I think this goes in the right direction.

* Could you alter the documentation to more describe the high-level.
* I think we should test this extensively. Maybe we should look up the number of partitions on Wikipedia and assert that our algorithm yields the correct number of partitions. Maybe we should quickcheck that each resulting partition is unique and is actually a partition of the input.
* We should explicitly think and test for partitions of an empty iterator: I think it should yield `Some([])` once and then `None`. Do you agree? This should be included in the tests, too.
* My main worry is how we could test that this algorithm actually generates all the partitions, and each of them exactly once. Maybe starting out with a vector and a shuffled copy, and assert that both of them yield - maybe in differing order - the same partitions?

PR #638, opened 2022-08-12
Chore: update tree_fold1 doc to remove reference to deprecated fold1

willemneal:
Not sure if this is needed, but I thought since it's deprecated it's best to move away from it.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2271,7 +2271,7 @@ pub trait Itertools : Iterator {
     /// ```
     ///
     /// Which, for non-associative functions, will typically produce a different
-    /// result than the linear call tree used by `fold1`:
+    /// result than the linear call tree used by [`Iterator::reduce`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.reduce):
     ///
     /// ```text
     /// 1 2 3 4 5 6 7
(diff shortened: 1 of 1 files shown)


phimuemue:
Hi! Thanks a bunch, it's appreciated.

From [reading](https://docs.rs/itertools/latest/src/itertools/lib.rs.html#2363) [other](https://docs.rs/itertools/latest/src/itertools/lib.rs.html#2953) [samples](https://docs.rs/itertools/latest/src/itertools/lib.rs.html#1979), it seems that links to other methods can be written simpler. Do you mind updating to the simpler syntax?

willemneal:
Sure!

src/lib.rs
```diff
@@ -2271,15 +2271,15 @@ pub trait Itertools : Iterator {
     /// ```
     ///
     /// Which, for non-associative functions, will typically produce a different
-    /// result than the linear call tree used by `fold1`:
+    /// result than the linear call tree used by [`Iterator::reduce`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.reduce):
```
willemneal, line 2274:
```suggestion
    /// result than the linear call tree used by [`Iterator::reduce`]:
```

src/lib.rs
```diff
@@ -2271,15 +2271,15 @@ pub trait Itertools : Iterator {
...
     ///
     /// ```text
     /// 1 2 3 4 5 6 7
     /// │ │ │ │ │ │ │
     /// └─f─f─f─f─f─f
     /// ```
     ///
-    /// If `f` is associative, prefer the normal `fold1` instead.
+    /// If `f` is associative, prefer the normal [`Iterator::reduce`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.reduce) instead.
```
willemneal, line 2282:
```suggestion
    /// If `f` is associative, prefer the normal [`Iterator::reduce`] instead.
```

phimuemue:
Thank you!

bors r+

PR #640, opened 2022-09-04
Should we release 0.10.4?

phimuemue:
Hi @jswrenn: I thought about publishing v0.10.4. It's not *that* much, but I think it's good that the new functions get exposure to real use cases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## 0.10.4
+  - Add `EitherOrBoth::or` and `EitherOrBoth::or_else` (#593)
+  - Add `min_set`, `max_set` et al. (#613, #323)
+  - Use `either/use_std` (#628)
+  - Documentation fixes (#612, #625, #632, #633, #634, #638)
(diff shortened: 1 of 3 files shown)


jswrenn:
Sounds good! I'll merge this, and publish a release (probably tomorrow morning).

Sorry for my absence; I'm hoping to start being more active again. What's the status of the rustfmt/clippy PRs? Are they still in-progress, or stalled? Should I avoid merging another PRs until those merge?

phimuemue:
> Sorry for my absence; I'm hoping to start being more active again. 

No worries.

> What's the status of the rustfmt/clippy PRs? Are they still in-progress, or stalled? Should I avoid merging another PRs until those merge?

I think #618 has some very good parts in it, but I did not yet get around to merging... I'll see when I can do them (feel free to do it before me if you want). I think it would be ok to postpone #619 at least until #618 is settled. 

From my point of view, you can merge the PRs you consider appropriate. And as a last resort (read: if we introduce conflicts in #618) I think @hellow554 did a good enough job so that merging should not be too hard.

PR #644, opened 2022-09-12
Implement `PeekingNext` for `PeekingTakeWhile`.

olson-sean-k:
This PR implements `PeekingNext` for `PeekingTakeWhile` by composing its predicate with the predicate given to `PeekingNext::peeking_next`. This allows `Itertools::peeking_take_while` to be chained and for subsequent calls, including those across function boundaries, to function as expected while restoring items in the originating iterator.

See also #643, which implements `PeekingNext` for mutable references. In combination, these changes allow code to generically accept types implementing `PeekingNext` where `Itertools::peeking_take_while` can be used by the caller to prepare an iterator and subsequently by a function where restoring items in the originating iterator is important (i.e., the function cannot simply use `Iterator::peekable` etc., because `Iterator::next` would unconditionally be called on the originating iterator).

### DIFF
--- a/src/peeking_take_while.rs
+++ b/src/peeking_take_while.rs
@@ -115,6 +115,18 @@ impl<'a, I, F> Iterator for PeekingTakeWhile<'a, I, F>
     }
 }
 
+impl<'a, I, F> PeekingNext for PeekingTakeWhile<'a, I, F>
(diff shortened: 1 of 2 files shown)


jswrenn (approved):
bors r+

PR #646, opened 2022-09-20
Use SPDX license format

atouchet:


phimuemue:
Thanks for this!

I merge this as it seems correct to me:
* Itertools states it's "Dual-licensed to be compatible with the Rust project" (see here for Rust's `license` fields: https://github.com/rust-lang/rust/search?l=TOML&q=license).
* Itertools states it's "Licensed under the Apache License, Version 2.0 https://www.apache.org/licenses/LICENSE-2.0 or the MIT license https://opensource.org/licenses/MIT, at your option", which seems to be exactly the case described in https://spdx.github.io/spdx-spec/using-SPDX-short-identifiers-in-source-files/#e4-representing-multiple-licenses as a "disjunctive license".

bors r+

Issue #495, opened 2020-11-22
Proposed method: all_equal_value`


Lucretiel:
It'd be nice to have a version of `all_equal` that returns the value in the iterator where all the items in the iterator compare equal to that value. Proposed implementation:

```
/// Return the first item in the iterator if it compares equal to all the other
/// items in the iterator. Return `None` if the iterator is empty, or if it
/// contains any items that do not equal the first item.
fn all_equal_value(mut self) -> Option<Self::Item> {
    let first = self.next()?;

    if self.all(|item| first == item) {
        Some(first)
    } else {
        None
    }
}
```

jswrenn:
Sounds like a nice complement to https://github.com/rust-itertools/itertools/pull/241!

PR #648, opened 2022-09-22
Add all_equal_value method

hottea773:
This resolves https://github.com/rust-itertools/itertools/issues/495 by adding an `all_equal_value` method.

(I'm not certain it's valuable enough to add, but thought I might as well submit a PR to find out.)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -904,7 +904,7 @@ pub trait Itertools : Iterator {
 
     /// Return an iterator adaptor that flattens every `Result::Ok` value into
     /// a series of `Result::Ok` values. `Result::Err` values are unchanged.
-    /// 
+    ///
     /// This is useful when you have some common error type for your crate and
     /// need to propagate it upwards, but the `Result::Ok` case needs to be flattened.
     ///
@@ -914,7 +914,7 @@ pub trait Itertools : Iterator {
     /// let input = vec![Ok(0..2), Err(false), Ok(2..4)];
     /// let it = input.iter().cloned().flatten_ok();
     /// itertools::assert_equal(it.clone(), vec![Ok(0), Ok(1), Err(false), Ok(2), Ok(3)]);
-    /// 
+    ///
     /// // This can also be used to propagate errors when collecting.
     /// let output_result: Result<Vec<i32>, bool> = it.collect();
     /// assert_eq!(output_result, Err(false));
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -1858,6 +1858,33 @@ pub trait Itertools : Iterator {
...
+    /// let data = vec![1, 1, 1, 2, 2, 3, 3, 3, 4, 5, 5];
+    /// assert_eq!(data.iter().all_equal_value(), None);
+    /// assert_eq!(data[0..3].iter().all_equal_value(), Some(&1));
+    /// assert_eq!(data[3..5].iter().all_equal_value(), Some(&2));
+    /// assert_eq!(data[5..8].iter().all_equal_value(), Some(&3));
+    ///
+    /// let data : Option<usize> = None;
+    /// assert_eq!(data.into_iter().all_equal_value(), None);
+    /// ```
+    fn all_equal_value(&mut self) -> Option<Self::Item>
```
scottmcm, line 1875: One thing I learned from [Stepanovand McJones](http://elementsofprogramming.com/) is that algorithms should return all the information it learned, when doing so is basically free.  So my instinct here is to return more information in the failure cases -- especially the "hey, which things didn't match?" that the caller might want to put in a human-useful error message.

That would imply [something](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=755462e7b3d872c0feafa7182566baf8) more like
```rust
pub fn all_equal_value<T: PartialEq>(mut it: impl Iterator<Item = T>) -> Result<T, Option<(T, T)>> {
    let first = it.next().ok_or(None)?;
    let other = it.find(|x| !(x == &first));
    if let Some(other) = other {
        Err(Some((first, other)))
    } else {
        Ok(first)
    }
}
```

(That return type is isomorphic to [`MinMaxResult`](https://docs.rs/itertools/latest/itertools/enum.MinMaxResult.html), but I think something that can be `.unwrap()`ed is better for this method.)

hottea773: That makes sense, I've updated the MR accordingly. Thanks.

phimuemue: > One thing I learned from [Stepanovand & McJones](http://elementsofprogramming.com/) is that algorithms should return all the information it learned, when doing so is basically free.

I somewhat like this thinking as it may serve as a general guideline (that we could think about adopting in itertools), but I have an genuine question here: Wouldn't this rather imply a return type of `Result<T, Option<(T, T, Iterator)>>` (so that the remaining elements could be retrieved from the iterator, too)?

scottmcm: Good point!  I think that, rather than returning `Self` in the error, what it means is that it should be `&mut self`, like how [`all`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.all) is `&mut self` since it too will short-circuit if not all the items match.

I should have noticed this when writing my demo -- since it calls `next`+`find`, which are both `&mut self`, there's no need for it to take ownership of the iterator.

(And I do like it as a guideline, but it's not absolute.  It's possible that there should still be simpler versions.  For example, `.any(p)` is "just" `.find(p).is_some()`, but that doesn't mean that `any` necessarily shouldn't exist.  So it's also possible that there should be both `all_equal_value(&mut self) -> bool` and `find_mismatch(&mut self)` that returns the details.)

phimuemue:
Thanks for this!

bors r+

PR #649, opened 2022-10-07
Use step_by instead of step in merge examples

### LINKED ISSUE #639: Docs: `merge` methods example still uses the deprecated `step` instead of the standard libraries' `step_by`
While reading the itertools docs, I noticed that the `step` method is deprecated since 0.8.0 in favor of the standard libraries `step_by` method. 
However several methods' examples still use `step` in their example code, leading to people using it. The following methods are affected:
- `merge`
- `merge_join_by`
- `kmerge`

DidiBear:
The uses `Iterator::step_by` in doc examples of these methods since `Itertools::step` has been deprecated:
- [`merge`](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.merge)
- [`merge_join_by`](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.merge_join_by)
- [`kmerge`](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.merge_join_by) 

Resolves #639

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -935,8 +935,8 @@ pub trait Itertools : Iterator {
     /// ```
     /// use itertools::Itertools;
     ///
-    /// let a = (0..11).step(3);
-    /// let b = (0..11).step(5);
+    /// let a = (0..11).step_by(3);
+    /// let b = (0..11).step_by(5);
     /// let it = a.merge(b);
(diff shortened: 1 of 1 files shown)


phimuemue:
Thanks for that! Always good to show best practice in the docs.

bors r+

PR #653, opened 2022-10-14
Document and test the difference between stable and unstable sorts

### LINKED ISSUE #652: Document and test the difference between stable and unstable sorts
At present the documentation for the sort functions does not make it clear that they are stable/unstable or explain what that means.   
I think it would be helpful to new users to spell this out clearly in the documentation.  

We can use the same wording as `Slice::sort` 'This sort is stable (i.e., does not reorder equal elements)'

The functions in question are:  

- sorted_unstable
- sorted_unstable_by
- sorted_unstable_by_key
- sorted
- sorted_by
- sorted_by_key
- sorted_by_cached_key

The doc tests for the stable `sorted_by` functions should make use of this property.  
Pull request incoming

wainwrightmark:
Closes #652

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2655,7 +2655,7 @@ pub trait Itertools : Iterator {
     /// use itertools::Itertools;
     ///
     /// // sort people in descending order by age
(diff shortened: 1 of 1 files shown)


jswrenn (approved):
bors r+

PR #654, opened 2022-10-19
Adding _by, by_key, largest variants of k_smallest

### LINKED ISSUE #586: Extend k_smallest with largest and key options
Currently this crate only has `k_smallest`. Compare this to the offering for `min`/`max` (which are the same function, just limited to k = 1):

```
position_max
position_max_by
position_max_by_key
position_min
position_min_by
position_min_by_key
```

I think it would be nice if we extended `k_smallest` to:

```
k_smallest
k_smallest_by
k_smallest_by_key
k_largest
k_largest_by
k_largest_by_key
```

ejmount:
This PR resolves #586 by adding the following public functions
```
k_smallest_by
k_smallest_by_key
k_largest
k_largest_by
k_largest_by_key
```

I used a custom binary heap to implement them (and rewrite `k_smallest`) so that the logic could be agnostic as to where the memory came from.  This is because I was originally motivated to make this PR by implementing `k_smallest` for const sizes so as to not require an allocator. Using a custom heap implementation makes that extension straightforward, but I've not included it just now to make this first PR more manageable. Unfortunately, this requires an MSRV of 1.36, although the clippy config seems to indicate that's already the case.

(The heap type is *not* publicly exposed to crate users, it is only implementation detail.)

The quickcheck test for `k_smallest` is completely untouched and still passes, so I am confident the heap logic is correct. I've also added doctests to some but not all of the new functions, which also pass, Let me know if you feel more documentation (in- or externally facing) is useful.

### DIFF
--- a/src/k_smallest.rs
+++ b/src/k_smallest.rs
@@ -1,22 +1,96 @@
-use alloc::collections::BinaryHeap;
-use core::cmp::Ord;
+use alloc::vec::Vec;
+use core::cmp::Ordering;
 
-pub(crate) fn k_smallest<T: Ord, I: Iterator<Item = T>>(mut iter: I, k: usize) -> BinaryHeap<T> {
-    if k == 0 {
-        return BinaryHeap::new();
+/// Consumes a given iterator, returning the minimum elements in **ascending** order.
+pub(crate) fn k_smallest_general<I, F>(mut iter: I, k: usize, mut comparator: F) -> Vec<I::Item>
+where
+    I: Iterator,
+    F: FnMut(&I::Item, &I::Item) -> Ordering,
+{
+    /// Sift the element currently at `origin` away from the root until it is properly ordered.
+    ///
+    /// This will leave **larger** elements closer to the root of the heap.
+    fn sift_down<T, F>(heap: &mut [T], is_less_than: &mut F, mut origin: usize)
+    where
+        F: FnMut(&T, &T) -> bool,
+    {
+        #[inline]
+        fn children_of(n: usize) -> (usize, usize) {
+            (2 * n + 1, 2 * n + 2)
+        }
+
+        while origin < heap.len() {
+            let (left_idx, right_idx) = children_of(origin);
+            if left_idx >= heap.len() {
+                return;
+            }
+
+            let replacement_idx =
+                if right_idx < heap.len() && is_less_than(&heap[left_idx], &heap[right_idx]) {
+                    right_idx
+                } else {
+                    left_idx
+                };
+
+            if is_less_than(&heap[origin], &heap[replacement_idx]) {
+                heap.swap(origin, replacement_idx);
+                origin = replacement_idx;
+            } else {
+                return;
+            }
+        }
     }
 
-    let mut heap = iter.by_ref().take(k).collect::<BinaryHeap<_>>();
+    if k == 0 {
+        return Vec::new();
+    }
+    let mut storage: Vec<I::Item> = iter.by_ref().take(k).collect();
 
-    iter.for_each(|i| {
-        debug_assert_eq!(heap.len(), k);
-        // Equivalent to heap.push(min(i, heap.pop())) but more efficient.
-        // This should be done with a single `.peek_mut().unwrap()` but
-        //  `PeekMut` sifts-down unconditionally on Rust 1.46.0 and prior.
-        if *heap.peek().unwrap() > i {
-            *heap.peek_mut().unwrap() = i;
-        }
-    });
+    let mut is_less_than = move |a: &_, b: &_| comparator(a, b) == Ordering::Less;
 
-    heap
+    // Rearrange the storage into a valid heap by reordering from the second-bottom-most layer up to the root.
+    // Slightly faster than ordering on each insert, but only by a factor of lg(k).
+    // The resulting heap has the **largest** item on top.
+    for i in (0..=(storage.len() / 2)).rev() {
+        sift_down(&mut storage, &mut is_less_than, i);
+    }
+
+    if k == storage.len() {
+        // If we fill the storage, there may still be iterator elements left so feed them into the heap.
+        // Also avoids unexpected behaviour with restartable iterators.
+        iter.for_each(|val| {
+            if is_less_than(&val, &storage[0]) {
+                // Treating this as an push-and-pop saves having to write a sift-up implementation.
+                // https://en.wikipedia.org/wiki/Binary_heap#Insert_then_extract
+                storage[0] = val;
+                // We retain the smallest items we've seen so far, but ordered largest first so we can drop the largest efficiently.
+                sift_down(&mut storage, &mut is_less_than, 0);
+            }
+        });
+    }
+
+    // Ultimately the items need to be in least-first, strict order, but the heap is currently largest-first.
+    // To achieve this, repeatedly,
+    // 1) "pop" the largest item off the heap into the tail slot of the underlying storage,
+    // 2) shrink the logical size of the heap by 1,
+    // 3) restore the heap property over the remaining items.
+    let mut heap = &mut storage[..];
+    while heap.len() > 1 {
+        let last_idx = heap.len() - 1;
+        heap.swap(0, last_idx);
+        // Sifting over a truncated slice means that the sifting will not disturb already popped elements.
+        heap = &mut heap[..last_idx];
+        sift_down(heap, &mut is_less_than, 0);
+    }
+
+    storage
+}
+
+#[inline]
+pub(crate) fn key_to_cmp<T, K, F>(key: F) -> impl Fn(&T, &T) -> Ordering
+where
+    F: Fn(&T) -> K,
+    K: Ord,
+{
+    move |a, b| key(a).cmp(&key(b))
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2950,14 +2950,108 @@ pub trait Itertools: Iterator {
     /// itertools::assert_equal(five_smallest, 0..5);
     /// ```
     #[cfg(feature = "use_alloc")]
-    fn k_smallest(self, k: usize) -> VecIntoIter<Self::Item>
+    fn k_smallest(mut self, k: usize) -> VecIntoIter<Self::Item>
     where
         Self: Sized,
         Self::Item: Ord,
     {
-        crate::k_smallest::k_smallest(self, k)
-            .into_sorted_vec()
-            .into_iter()
+        // The stdlib heap has optimised handling of "holes", which is not included in our heap implementation in k_smallest_general.
+        // While the difference is unlikely to have practical impact unless `Self::Item` is very large, this method uses the stdlib structure
+        // to maintain performance compared to previous versions of the crate.
+        use alloc::collections::BinaryHeap;
+
+        if k == 0 {
+            return Vec::new().into_iter();
+        }
+
+        let mut heap = self.by_ref().take(k).collect::<BinaryHeap<_>>();
+
+        self.for_each(|i| {
+            debug_assert_eq!(heap.len(), k);
+            // Equivalent to heap.push(min(i, heap.pop())) but more efficient.
+            // This should be done with a single `.peek_mut().unwrap()` but
+            //  `PeekMut` sifts-down unconditionally on Rust 1.46.0 and prior.
+            if *heap.peek().unwrap() > i {
+                *heap.peek_mut().unwrap() = i;
+            }
+        });
+
+        heap.into_sorted_vec().into_iter()
+    }
+
+    /// Sort the k smallest elements into a new iterator using the provided comparison.
+    ///
+    /// This corresponds to `self.sorted_by(cmp).take(k)` in the same way that
+    /// [Itertools::k_smallest] corresponds to `self.sorted().take(k)`, in both semantics and complexity.
+    /// Particularly, a custom heap implementation ensures the comparison is not cloned.
+    #[cfg(feature = "use_alloc")]
+    fn k_smallest_by<F>(self, k: usize, cmp: F) -> VecIntoIter<Self::Item>
+    where
+        Self: Sized,
+        F: Fn(&Self::Item, &Self::Item) -> Ordering,
+    {
+        k_smallest::k_smallest_general(self, k, cmp).into_iter()
+    }
+
+    /// Return the elements producing the k smallest outputs of the provided function
+    ///
+    /// This corresponds to `self.sorted_by_key(cmp).take(k)` in the same way that
+    /// [Itertools::k_smallest] corresponds to `self.sorted().take(k)`, in both semantics and time complexity.
+    #[cfg(feature = "use_alloc")]
+    fn k_smallest_by_key<F, K>(self, k: usize, key: F) -> VecIntoIter<Self::Item>
+    where
+        Self: Sized,
+        F: Fn(&Self::Item) -> K,
+        K: Ord,
+    {
+        self.k_smallest_by(k, k_smallest::key_to_cmp(key))
+    }
+
+    /// Sort the k largest elements into a new iterator, in descending order.
+    /// Semantically equivalent to `k_smallest` with a reversed `Ord`
+    /// However, this is implemented by way of a custom binary heap
+    /// which does not have the same performance characteristics for very large `Self::Item`
+    /// ```
+    /// use itertools::Itertools;
+    ///
+    /// // A random permutation of 0..15
+    /// let numbers = vec![6, 9, 1, 14, 0, 4, 8, 7, 11, 2, 10, 3, 13, 12, 5];
+    ///
+    /// let five_largest = numbers
+    ///     .into_iter()
+    ///     .k_largest(5);
+    ///
+    /// itertools::assert_equal(five_largest, vec![14,13,12,11,10]);
+    /// ```
+    #[cfg(feature = "use_alloc")]
+    fn k_largest(self, k: usize) -> VecIntoIter<Self::Item>
+    where
+        Self: Sized,
+        Self::Item: Ord,
+    {
+        self.k_largest_by(k, Self::Item::cmp)
+    }
+
+    /// Sort the k largest elements into a new iterator using the provided comparison.
+    /// Functionally equivalent to `k_smallest_by` with a reversed `Ord`
+    #[cfg(feature = "use_alloc")]
+    fn k_largest_by<F>(self, k: usize, cmp: F) -> VecIntoIter<Self::Item>
+    where
+        Self: Sized,
+        F: Fn(&Self::Item, &Self::Item) -> Ordering,
+    {
+        self.k_smallest_by(k, move |a, b| cmp(b, a))
+    }
+
+    /// Return the elements producing the k largest outputs of the provided function
+    #[cfg(feature = "use_alloc")]
+    fn k_largest_by_key<F, K>(self, k: usize, key: F) -> VecIntoIter<Self::Item>
+    where
+        Self: Sized,
+        F: Fn(&Self::Item) -> K,
+        K: Ord,
+    {
+        self.k_largest_by(k, k_smallest::key_to_cmp(key))
     }
 
     /// Collect all iterator elements into one of two
--- a/tests/test_std.rs
+++ b/tests/test_std.rs
@@ -492,23 +492,50 @@ fn sorted_by() {
 }
 
 qc::quickcheck! {
-    fn k_smallest_range(n: u64, m: u16, k: u16) -> () {
+    fn k_smallest_range(n: i64, m: u16, k: u16) -> () {
         // u16 is used to constrain k and m to 0..2¹⁶,
         //  otherwise the test could use too much memory.
-        let (k, m) = (k as u64, m as u64);
+        let (k, m) = (k as usize, m as u64);
 
+        let mut v: Vec<_> = (n..n.saturating_add(m as _)).collect();
         // Generate a random permutation of n..n+m
-        let i = {
-            let mut v: Vec<u64> = (n..n.saturating_add(m)).collect();
-            v.shuffle(&mut thread_rng());
-            v.into_iter()
-        };
+        v.shuffle(&mut thread_rng());
 
-        // Check that taking the k smallest elements yields n..n+min(k, m)
-        it::assert_equal(
-            i.k_smallest(k as usize),
-            n..n.saturating_add(min(k, m))
-        );
+        // Construct the right answers for the top and bottom elements
+        let mut sorted = v.clone();
+        sorted.sort();
+        // how many elements are we checking
+        let num_elements = min(k, m as _);
+
+        // Compute the top and bottom k in various combinations
+        let smallest = v.iter().cloned().k_smallest(k);
+        let smallest_by = v.iter().cloned().k_smallest_by(k, Ord::cmp);
+        let smallest_by_key = v.iter().cloned().k_smallest_by_key(k, |&x| x);
+
+        let largest = v.iter().cloned().k_largest(k);
+        let largest_by = v.iter().cloned().k_largest_by(k, Ord::cmp);
+        let largest_by_key = v.iter().cloned().k_largest_by_key(k, |&x| x);
+
+        // Check the variations produce the same answers and that they're right
+        for (a,b,c,d) in izip!(
+            sorted[..num_elements].iter().cloned(),
+            smallest,
+            smallest_by,
+            smallest_by_key) {
+            assert_eq!(a,b);
+            assert_eq!(a,c);
+            assert_eq!(a,d);
+        }
+
+        for (a,b,c,d) in izip!(
+            sorted[sorted.len()-num_elements..].iter().rev().cloned(),
+            largest,
+            largest_by,
+            largest_by_key) {
+            assert_eq!(a,b);
+            assert_eq!(a,c);
+            assert_eq!(a,d);
+        }
     }
 }
 


src/k_smallest.rs
```diff
@@ -1,20 +1,245 @@
...
+use core::cmp::{Ord, Ordering, Reverse};
+use core::mem::{replace, transmute, MaybeUninit};
 
-pub(crate) fn k_smallest<T: Ord, I: Iterator<Item = T>>(mut iter: I, k: usize) -> BinaryHeap<T> {
-    if k == 0 { return BinaryHeap::new(); }
+fn k_smallest_dynamic<T, I: Iterator<Item = T>>(
+    iter: I,
+    k: usize,
+    order: impl Fn(&T, &T) -> Ordering,
+) -> std::vec::IntoIter<T> {
```
jswrenn, line 8: Using `std::` instead of `alloc::` is going to break users who can't pull in `std` but can pull in `alloc`.

ejmount: Good point, been fixed to use `alloc`

phimuemue:
Hi there! Thanks for tackling this - I'm happy you went for it.

I hope I don't miss the forest for the trees, but I am pondering whether `unsafe` is really needed here. It seems that it is used here because `MaxHeap` manually tracks memory via `storage: &mut [MaybeUninit<T>]` and `len: usize`. But wouldn't a `Vec` be perfectly fine for this? (Rust's own [`BinaryHeap`](https://doc.rust-lang.org/src/alloc/collections/binary_heap.rs.html#268) uses `Vec`, too.)

Moreover I think that we could benefit from inlining some functions:
* `push_pop`: It checks `self.storage.is_empty` internally, but I think we could use an early out for this case at the call site.
* `fn extend<T: IntoIterator<Item = A>>(&mut self, iter: T)`: I think the users do not benefit if we implement this trait for an internal data structure. It deals with stuff that is not used in practice (such as respecting `self.len`, although the only call to it assures that `self.len==0`), and thus makes things a bit more complex than necessary.
* `total_sort`/`capped_heapsort`: They transport information of internal data structures, and as it stands now, one has to go follow function calls to find out what these values actually mean. If everything was locally, this would be a bit easier to grasp.

ejmount:
> I hope I don't miss the forest for the trees, but I am pondering whether `unsafe` is really needed here. It seems that it is used here because `MaxHeap` manually tracks memory via `storage: &mut [MaybeUninit<T>]` and `len: usize`. But wouldn't a `Vec` be perfectly fine for this? (Rust's own [`BinaryHeap`](https://doc.rust-lang.org/src/alloc/collections/binary_heap.rs.html#268) uses `Vec`, too.)

It would do for the current implementation, but in the future I wanted to be able to do something like,

```rust
fn k_smallest_static<'a, T, I, F, const K: usize>(iter: I, order: F) -> impl Iterator<Item = T>
where
    T: 'a,
    I: Iterator<Item = T>,
    F: Fn(&T, &T) -> std::cmp::Ordering,
{
    let mut storage: [MaybeUninit<T>; K] = /* ... */;
    let num_elements = capped_heapsort(iter, &mut storage, order);
    storage
        .into_iter()
        .take(num_elements)
        .map(|e| unsafe { e.assume_init() })
}
```
and have it be usable without any dependency on `alloc`. 


> Moreover I think that we could benefit from inlining some functions:

I try to write code that's robust against the caller making mistakes by forgetting special conditions. I totally agree with you that that's created overhead. I've changed the flow of creating and consuming the heap with the main differences being
* `capped_heapsort` short-circuits on zero-length storage, so all the checks for that case disappear from the heap itself
* The `extend` logic is merged into the construction process, with no trait. This also means it is (locally) obvious that the heap will start empty, so the adjustments for `self.len` disappear too. 
* `total_sort` is inlined into `capped_heapsort`, which now returns a (local) `Range` for better context of what the return value signifies

phimuemue:
Hi, thanks for addressing some issues.

> It would do for the current implementation, but in the future I wanted to be able to do something like,
> 
> ```rust
> fn k_smallest_static<'a, T, I, F, const K: usize>(iter: I, order: F) -> impl Iterator<Item = T>
> ```

Noble motives, I like that (pretty much, actually). However, I'd still argue to simply go with `Vec` for now, because if we want to support `static`-variants of our algorithms, we probably also want other algorithms to work with `Vec` or (essentially) `ArrayVec` (e.g. `min_set` and its cousins). And in that case, it's probably much easier to go through the codebase, look for `Vec`, and replace those by a trait `VecOrArrayVec` instead of tying the static/dynamic decision to one particular algorithm (in this case: `k_smallest`). That is, i'd like to move the memory management issues out of the algorithms into the containers used by the algorithms, and make the algorithms accept different containers. @jswrenn Do you have thoughts about this?

Then, I'd imagine the algorithm `unsafe`-free and something like:
* `let container = iter.by_ref().take(k).collect()`
* if iterator already exhausted, sort `container` using a sorting function (possibly avoiding heapify if it helps speed)
* if there's more to come from the iterator, `push_pop` the remaining elements, and sort afterwards (possibly exploiting heap-structure if it helps speed)

On top, I'd suggest more inlining:
* `capped_heapsort`: It internally checks for `storage.is_empty`, which could be done in the caller (avoiding to call it entirely) by testing `k==0`.
* `from_iter`: As it stands now, it seems to rely e.g. on `heap.len` being `0` - something that's obvious if it was inlined.
* `push_pop`: Calls e.g. `self.get(0)`, which lifts values into `Option`. However, `push_pop` is only called for non-empty storage, which would be more obvious if it was inlined.
* To add to the previous point: Having a `compare` *might* be a good idea, but I was thinking if we could get rid of the detour through `Option` entirely (without too much hassle).

src/k_smallest.rs
```diff
@@ -1,20 +1,230 @@
...
+fn k_smallest_dynamic<T, I: Iterator<Item = T>>(
+    iter: I,
+    k: usize,
+    order: impl Fn(&T, &T) -> Ordering,
+) -> IntoIter<T> {
+    let mut storage = Vec::new();
+    storage.resize_with(k, MaybeUninit::uninit);
+    let Range { end, .. } = capped_heapsort(iter, &mut storage, order);
+    storage.truncate(end);
+    let initialized: Vec<_> = unsafe { transmute(storage) };
```
phimuemue, line 15: Sidenote: Is this `transmute` even allowed? (I think it's not.) [The docs](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#layout-1) state that the layout of `MaybeUninit<T>` is the same as the one for `T`, but that types containing `MaybeUninit<T>`/`T` do *not* necessarily have the same layout. (Rust apparently could e.g. reorder members of `Vec<MaybeUninit<T>>` in another way than the members of `Vec<T>`.) It seems the way out would be [`from_raw_parts`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.from_raw_parts), but I'd rather avoid `unsafe` in the first place.

ejmount:
> Noble motives, I like that (pretty much, actually). However, I'd still argue to simply go with `Vec` for now, because if we want to support `static`-variants of our algorithms, we probably also want other algorithms to work with `Vec` or (essentially) `ArrayVec` (e.g. `min_set` and its cousins). And in that case, it's probably much easier to go through the codebase, look for `Vec`, and replace those by a trait `VecOrArrayVec` instead of tying the static/dynamic decision to one particular algorithm (in this case: `k_smallest`). That is, i'd like to move the memory management issues out of the algorithms into the containers used by the algorithms, and make the algorithms accept different containers.

I think this is a much better strategy than what I had in mind, so I've restructured the heap logic to use a plain Vec for now with an eye to replacing it with a trait in the future. That neatly eliminated all of the `unsafe` also. I also like that strategy because it avoids an explosion of very similar methods doing fundamentally the same thing, which I had been concerned about. 

>     * if iterator already exhausted, sort `container` using a sorting function (possibly avoiding heapify if it helps speed)

While I don't have benchmarks, I can't imagine many circumstances where heapify would be slower, since it works in `O(k)` time rather than the `O(k*ln k)` required for a sort. 

>     * `capped_heapsort`: It internally checks for `storage.is_empty`, which could be done in the caller (avoiding to call it entirely) by testing `k==0`. Also, for someone new, the meaning of the `range` returned from `capped_heapsort` is probably still non-obvious, (in addition of its begin (that is always 0) being ignored at the single call site).

Most of the reason for that check being placed in `capped_heapsort` rather than earlier was because I was expecting that function to have a static caller as well, which is no needed. In the new version, `capped_heapsort` has disappeared entirely into the caller, and being able to rely on `truncate` means there's nothing to return even if it did still exist. 

>     * `from_iter`: As it stands now, it seems to rely e.g. on `heap.len` being `0` - something that's obvious if it was inlined.

I'm afraid I don't follow - `from_iter` knows that the heap begins empty because it is created there in the first place. This is hopefully clearer in the new version that uses `collect` as you were suggesting earlier.
 
>     * `push_pop`: Calls e.g. `self.get(0)`, which lifts values into `Option`. However, `push_pop` is only called for non-empty storage, which would be more obvious if it was inlined.
>     * To add to the previous point: Having a `compare` going through `Option`s _might_ be a good idea, but I was thinking if we could get rid of the detour through `Option` entirely (without too much hassle).

`compare` goes through `Option` primarily for the benefit of `sift_down`, where the difference between elements not existing and already being in the right order is immaterial. It does have the benefit that `push_pop` works on empty storage, but as you say, that never happens in practice, so I'm not sure if it's worth bringing attention to.

src/k_smallest.rs
ejmount: I suspect you're right that the transmute isn't valid and `from_raw_parts` would be needed, but as mentioned below, the new implementation is entirely safe.

Cargo.toml
```diff
@@ -39,9 +39,10 @@ default-features = false
 version = "0.2"
 
 [features]
-default = ["use_std"]
+default = ["use_std", "const_generics"]
 use_std = ["use_alloc", "either/use_std"]
 use_alloc = []
+const_generics = []
```
phimuemue, line 45: Please don't include this.

src/k_smallest.rs
```diff
@@ -1,20 +1,211 @@
...
+        fn children_of(n: usize) -> (usize, usize) {
+            (2 * n, 2 * n + 1)
+        }
+
+        let (left_idx, right_idx) = children_of(origin);
+        let (left_item, right_item) = (self.get(left_idx), self.get(right_idx));
+
+        if left_item.is_none() { // the left is the earlier child, so if it doesn't exist there's nothing to swap with
+            return;
+        }
```
phimuemue, line 131: This essentially checks if `left_idx` is valid. Instead of constructing `left_item`, and then deducing whether the index was valid, we should simply check if the index is valid directly.

src/k_smallest.rs
```diff
@@ -1,20 +1,211 @@
...
+
+        let cmp = self
+            .compare(left_item, right_item)
+            .unwrap_or(Ordering::Greater); // The right item may not exist, so default to picking the left, i.e. lower
+
+        let replacement_idx = match cmp {
+            Ordering::Greater | Ordering::Equal => left_idx,
+            Ordering::Less => right_idx,
+        };
```
phimuemue, line 140: After some thinking, I would suggest that we should avoid going through `Option` entirely here - instead properly checking if the indices are valid, and only then accessing the respective elements. This should allow us to avoid `fn compare`.

src/k_smallest.rs
```diff
@@ -1,20 +1,211 @@
...
+    I: Iterator<Item = T>,
+    F: Fn(&T) -> K,
+    K: Ord,
+{
+    let iter = iter.map(|v| Pair(key(&v), v));
+
+    let results: Vec<_> = k_smallest_general(iter, k, Ord::cmp)
+        .map(|Pair(_, t)| t)
+        .collect();
+    results.into_iter()
```
phimuemue, line 59: I guess the intent is to avoid multiple evaluations of `key` for the same argument. Please correct me if I'm mistaken.

We should do along the lines of `std` here (`sort_by_key` vs `sort_by_cached_key`), and simply allow multiple evaluations per element instead of allocating an auxiliary vector for intermediate results. If it's a problem, users can still employ your workaround of `map(|v| Pair(key(&v), v))` or use `k_smallest_by_key(iter, k, make_cached_key(key))` (again transferring the caching out of the algorithm, so that the caching could also be used for e.g. `min_set_by_key`), or `itertools` could - as a last resort - offer `k_smallest_by_cached_key`.

src/k_smallest.rs
```diff
@@ -1,20 +1,211 @@
...
-        debug_assert_eq!(heap.len(), k);
-        // Equivalent to heap.push(min(i, heap.pop())) but more efficient.
-        // This should be done with a single `.peek_mut().unwrap()` but
-        //  `PeekMut` sifts-down unconditionally on Rust 1.46.0 and prior.
-        if *heap.peek().unwrap() > i {
-            *heap.peek_mut().unwrap() = i;
+        let cmp = self.compare(original_item, replacement_item);
+        if Some(Ordering::Less) == cmp {
+            self.storage.swap(origin, replacement_idx);
+            self.sift_down(replacement_idx);
```
phimuemue, line 147: Nitpick (but I wouldn't be too strict about this if we do not run into actual problems): Should we write `sift_down` iteratively, avoiding the stack memory consumption of `log(n)`?

phimuemue (changes requested):
Thank you, looks much better already.

I don't want to bother you, but please do the following:

* Inline *all* auxiliary functions that are called once (in particular `from_iter`, `heapify`, `push_pop` (which should not use `compare`, but just use the comparator directly without going through `Option`), `unwrap_sorted`, `pop`). Because many of them have side effects manipulating internal data structures, it's easier to grasp what they can and cannot assume about the current state.
* After doing the above, I think we do not need `MaxHeap::len` anymore, and use a local variable instead. Maybe more possibilities for simplification arise.
* Make `MaxHeap::storage` a `&mut [T]` (that can be backed by a `Vec`, `ArrayVec`, `Array`, and possibly others). I think we want to have contiguous memory, but that's about it. If everything is inlined, we may even be able to avoid `struct MaxHeap` altogether, being able to simply deal with local variables.

phimuemue:
Probably this should also include tests (maybe via quickcheck, comparing `k_smallest...` against collecting into a vector, sorting, and then picking the first/last `k` elements.

Cargo.toml
ejmount: Oops, that wasn't supposed to be included, I thought I had it in a different branch

ejmount: Also realized I had made quite a few changes to that file for no net purpose, so I rebased the change out entirely.

src/k_smallest.rs
ejmount: Fair enough, made that change

src/k_smallest.rs
ejmount: Yeah, this set up is so that `key` only has to be evaluated once. 

However, the auxillary vector is only necessary to be consistent with the other methods returning `VecIntoIter`. If the return type can be an arbitrary iterable of T, then it could equally be a `Map<>` of the original `Vec<Pair>`, and no second allocation would be needed

src/k_smallest.rs
ejmount: My expectation would be that the compiler [already does that](https://llvm.org/docs/Passes.html#tailcallelim-tail-call-elimination), so it takes `O(1)` stack space regardless. However, it might be worth writing it explicitly as a loop regardless, so the space consumption is clearer to the reader.

src/k_smallest.rs
phimuemue: Just to be clear: The `Vec<T>` is fine for me, but not the `Vec<Pair<K, T>>`. Instead, this should call `k_smallest_general(original_unmapped_iter, k, |a,b| key(a).cmp(key(b))` - and not force all users to allocate an intermediate `Vec<Pair<K, T>>`.

src/k_smallest.rs
```diff
@@ -1,20 +1,214 @@
...
+        } else {
+            None
+        }
+    }
 
-    let mut heap = iter.by_ref().take(k).collect::<BinaryHeap<_>>();
+    /// Sift the element currently at `origin` **away** from the root until it is properly ordered
+    fn sift_down(&mut self, origin: usize) {
+        fn children_of(n: usize) -> (usize, usize) {
+            (2 * n, 2 * n + 1)
```
phimuemue, line 123: Is this really correct? This looks dubious, as index `0` is mapped to `(0,1)`, while I think it should be `(1,2)`.

ejmount: I'd agree. It worked, but I think the result was a slightly misshapen tree. Has now been corrected

ejmount:
I've been making updates to this as time allows, and I've just pushed the following changes

* the calculation for child node indices is now correct for 0
* `sift_down` is now iterative
* All uses of `Option` have been removed and all the bounds checking logic is now explicit
* all of the single-call auxillary heap functions have been inlined, and now work on locals rather than an explicit heap type. `len` actually disappeared entirely because its part of the slice being passed.

I've not added any new tests, but will see about doing so when I get more time. 

Re; keyed sorts, I'm afraid I don't quite understand the part of your comment about the `Vec<T>` being fine. Just to make sure we're all on the same page, my understanding is that there are two ways to do this with a single allocation: 

1. time/eval efficiently: 
    ```rust
    fn k_smallest_by_key<T, I, F, K>(iter: I, k: usize, key: F) -> impl Iterator<Item=T>
    where
        I: Iterator<Item = T>,
        F: Fn(&T) -> K,
        K: Ord,
    {
        let iter = iter.map(|v| Pair(key(&v), v));
        k_smallest_general(iter, k, Ord::cmp).map(|Pair(_, t)| t)
    }
    ```
2. space efficiently: 
    ```rust
    fn k_smallest_by_key<T, I, F, K>(iter: I, k: usize, key: F) -> VecIntoIter<T>
    where
        I: Iterator<Item = T>,
        F: Fn(&T) -> K,
        K: Ord,
    {
        k_smallest_general(iter, k, |a,b| key(&a).cmp(key(&b)))
    }
    ```

There is, AFAIK, no way to do this time efficiently using only a `Vec<T>`

If we only have one of them, my inclination would be that it should be the first one, because I expect the use cases where callers care about the extra space required would be rare enough that it wouldn't be a burden to use the 2nd version explicitly. Additionally, in the standard library, `sort_by_cached_key` [came](https://github.com/rust-lang/rust/issues/34447) much [later](https://github.com/rust-lang/rust/pull/48639) than `sort_by_key`, which (given the amount of overhead involved in new fatures) suggests to me that `sort_by_key` not caching originally wasn't a design choice that was considered and rejected rather than missed entirely. 

However, the call is up to you, and I'd appreciate if you could specify whether you want the first, second or both options included.

phimuemue:
> Re; keyed sorts, I'm afraid I don't quite understand the part of your comment about the Vec<T> being fine. Just to make sure we're all on the same page, my understanding is that there are two ways to do this with a single allocation:

Well described. Please go for the space efficient solution.

ejmount:
Done. With regard to tests, would you be able to be more specific about what you think should be covered by quickcheck? I assume you don't mean all six variations separately, since they're mostly implemented in terms of each other.

phimuemue:
Maybe you can look what's already there and extend this. If there are no quicktests yet, you could e.g. collect into a vector, sort the vector, and get the first/last `k` items - and assert that the result is the same as if you'd done `k_smallest` (or one of its cousins).

I'd appreciate if you'd write tests for each variation. You're right, they are implemented in terms of each other, but the tests shouldn't assume this.

    Am Freitag, 28. Oktober 2022 um 13:10:48 MESZ hat Ewan Mount ***@***.***> Folgendes geschrieben:  
 
 


Done. With regard to tests, would you be able to be more specific about what you think should be covered by quickcheck? I assume you don't mean all six variations separately, since they're mostly implemented in terms of each other.

ejmount:
I've just added a quickcheck that tests all 6 variations against fully sorting the input data

ejmount:
The build failed earlier because of an accidental reference to `std::vec::Vec;`, which is now corrected

phimuemue:
Hi there, first of all: Thanks @ejmount for staying in touch.

I saw that you implement `k_smallest` in terms of your custom heap (i.e. it delegates to `k_smallest_by`). However, this potentially degrades performance, as [Rust's `BinaryHeap` uses "holes" to improve performance](https://doc.rust-lang.org/src/alloc/collections/binary_heap.rs.html#543-550).

Writing an own binary heap turns out to be tricky, so I'd lean to go with a safe implementation for now, sacrificing the holes for the more general variants, but possibly state this in the documentation. We can then try later to write a safe custom heap.

That is, before merging I'd suggest:
* The existing `k_smallest` should use Rusts own `BinaryHeap` to ensure performance does not degrade.
* New `k_smallest_...`-variants should *for now* use a safe custom binary heap (I don't feel competent to review an `unsafe` binary heap), but should state this in the docs.
* If possible, the existing `k_smallest` and the new variants should share code, possibly abstracting the heap type via traits.

Before continuing on this, I'd ask @jswrenn and/or @scottmcm for their opinion on this.

ejmount:
This came back to my attention recently and I've made the changes you've suggested for `k_smallest`, (although *not* `k_largest`, it having no existing callers) but I couldn't see any particularly good way to share functionality between that and the other variants, given that `k_smallest_general` is very self-contained. 

AFAICT, the code as it stands is identical, in both functionalty and performance, from the POV of any existing users. Assuming my upate just now addresses your concerns in the previous post, are there any other blockers for this being merged? An [earlier comment](https://github.com/rust-itertools/itertools/pull/654#pullrequestreview-1152287127) is still flagged as outstanding, but I don't think it's still applicable

douweschulte:
Great work! I was just looking for this functionality as I need `k_largest_by` for an algorithm I am working on. I see the PR has become a bit stale, is there something I can do to help this PR over the finish line?

Philippe-Cholet:
@douweschulte I don't know the details about this one. It's definitely something I'd like to see merged at some point though.
@phimuemue Can you shed some light on those extensions of `k_smallest`?

Philippe-Cholet:
I would like to revive this (most-loved) PR, and update this myself if needed. Based on https://github.com/rust-itertools/itertools/pull/654#issuecomment-1349675510:

@jswrenn and/or @scottmcm What is your opinion on this?

douweschulte:
I have some additional info now as well. Since the above comment I have copy pasted this PRs code into a couple of my own projects. I have only used the `k_largest` function but it has worked wonders!

src/k_smallest.rs
```diff
@@ -1,20 +1,92 @@
...
+pub(crate) fn k_smallest_general<T, I: Iterator<Item = T>>(
+    mut iter: I,
+    k: usize,
+    mut comparator: impl FnMut(&T, &T) -> Ordering,
+) -> IntoIter<T> {
+    if k == 0 {
+        return Vec::new().into_iter();
+    }
+    let mut storage: Vec<T> = iter.by_ref().take(k).collect();
+    let mut heap = &mut storage[..];
```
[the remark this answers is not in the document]
phimuemue: If I'm reading this correctly, the largest element goes to `heap[0]`. We should comment this somewhere.

src/k_smallest.rs
```diff
@@ -1,20 +1,92 @@
...
+
+pub(crate) fn reverse_key<T, K, F>(f: F) -> impl Fn(&T) -> Reverse<K>
+where
+    F: Fn(&T) -> K,
+{
+    move |a| Reverse(f(a))
+}
+
+/// Sift the element currently at `origin` **away** from the root until it is properly ordered
+fn sift_down<T>(
```
phimuemue, line 61: Please make this a local function within `k_smallest_general`, it reduces jumping through the code when reading it.

src/lib.rs
```diff
@@ -2732,13 +2732,110 @@ pub trait Itertools : Iterator {
...
+            debug_assert_eq!(heap.len(), k);
+            // Equivalent to heap.push(min(i, heap.pop())) but more efficient.
+            // This should be done with a single `.peek_mut().unwrap()` but
+            //  `PeekMut` sifts-down unconditionally on Rust 1.46.0 and prior.
+            if *heap.peek().unwrap() > i {
+                *heap.peek_mut().unwrap() = i;
+            }
+        });
+
+        heap.into_sorted_vec().into_iter()
```
phimuemue, line 2761: I assume this was copy-pasted from the previous implementation.

src/lib.rs
```diff
@@ -2732,13 +2732,110 @@ pub trait Itertools : Iterator {
...
+    where
+        Self: Sized,
+        F: Fn(&Self::Item) -> K,
+        K: Ord,
     {
-        crate::k_smallest::k_smallest(self, k)
-            .into_sorted_vec()
-            .into_iter()
+        let key = crate::k_smallest::reverse_key(key);
+        self.k_smallest_by_key(k, key)
```
phimuemue, line 2838: Shouldn't this redirect to `self.k_largest_by` (similar to `k_smallest_by_key`)? (Would probably allow us to use `reverse_cmp` in only k_largest_by.)

src/lib.rs
```diff
@@ -2732,13 +2732,110 @@ pub trait Itertools : Iterator {
...
+    ///
+    /// itertools::assert_equal(five_largest, vec![14,13,12,11,10]);
+    /// ```
+    #[cfg(feature = "use_alloc")]
+    fn k_largest(self, k: usize) -> VecIntoIter<Self::Item>
+    where
+        Self: Sized,
+        Self::Item: Ord,
+    {
+        self.k_smallest_by(k, k_smallest::reverse_cmp(Self::Item::cmp))
```
phimuemue, line 2815: Could this return `self.k_smallest_by(k, |x| x)`? (Would probably allow us to use `reverse_cmp` in only `k_largest_by`.)

phimuemue (commented):
Thanks for this. I skimmed this. I'd like some changes, mainly to make the code more compact to avoid jumping back and forth while reading (local functions, narrowly scope some variables).

On top, I suggest to return `Vec` in all cases. (Seems canonical with what we already have.)

@Philippe-Cholet I'd like to hand this over to you, as I do not have enough capacity right now. Please let me know if that's not ok.

src/k_smallest.rs
```diff
@@ -1,20 +1,92 @@
...
+    while heap.len() > 1 {
+        let last_idx = heap.len() - 1;
+        heap.swap(0, last_idx);
+        // Leaves the length shorter than the number of elements
+        // so that sifting does not disturb already popped elements
+        heap = &mut heap[..last_idx];
+        sift_down(heap, &mut comparator, 0);
+    }
+
+    storage.into_iter()
```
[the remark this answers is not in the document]
ejmount: That's fine, but do you also mean that the public-facing functions should return `Vec<T>` (and be inconsistent with `k_smallest`'s existing expectations) or just change the internal interface?

src/k_smallest.rs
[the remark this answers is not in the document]
ejmount: > Could we scope `storage` and `heap` as narrow as possible? `storage` needed throughout the function, `heap` only for the sorting part below => scope `heap` as narrow as possible. (And possibly rename `storage` -> `heap`, `heap` -> `heap_suffix` or similar.)

I see what you mean about the scope, that should be an easy change. I'm less sure about the naming, because I think my original reasoning was that by the point `heap` shrinks to nothing, `storage` is *not* a heap, it's in strict order. (It also shrinks to the left, so it'd be `heap_prefix`) 

> If I'm reading this correctly, the largest element goes to `heap[0]`. We should comment this somewhere.

I had to check this carefully and I'm still not sure, so I'm definitely going to comment that somewhere or ideally, name the comparator something more suggestive.

src/lib.rs
Philippe-Cholet: Being fan of `k_smallest`, I immediately recognize it to be the previous implemention yes, which is fine for me.

src/k_smallest.rs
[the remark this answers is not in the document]
Philippe-Cholet: I would disagree to return `Vec<_>` to the user. The current `k_smallest` returns `VecIntoIter` so I think it's best that all 5 others do that as well (since there is no reallocation if the user do `.collect_vec()` on it).
I would not mind if it's just the internal interface.

Philippe-Cholet:
@phimuemue I assigned myself here so I agree to take this off your hands.
Agree to make the code more compact.
(about `Vec`, see other comment)

@ejmount Hello, I was not sure you would still be around so it's nice to see you here.
I'll be around as well and this is definitely top-prio for me.
Since the time you worked on this, the repo has been finally rustfmt'ed (and CI reject compiler and clippy warnings) so I think the current conflicts are just about formatting. Personally, I would first squash everything and rebase on master to start clean. Or we can end by that, as you wish.

src/k_smallest.rs
```diff
@@ -1,20 +1,92 @@
-use alloc::collections::BinaryHeap;
-use core::cmp::Ord;
+use alloc::vec::{Vec, IntoIter};
+use core::cmp::{Ordering, Reverse};
 
-pub(crate) fn k_smallest<T: Ord, I: Iterator<Item = T>>(mut iter: I, k: usize) -> BinaryHeap<T> {
-    if k == 0 { return BinaryHeap::new(); }
+/// Consumes a given iterator, returning the minimum elements in **ascending** order.
+pub(crate) fn k_smallest_general<T, I: Iterator<Item = T>>(
```
Philippe-Cholet, line 5: Probably not sensitive but I think `T` should be replaced by `I::Item` (and just `I: Iterator`).

Philippe-Cholet (commented):
I'm not adding much since **phimuemue** already did a bunch here.

src/lib.rs
```diff
@@ -2732,13 +2732,110 @@ pub trait Itertools : Iterator {
...
-    fn k_smallest(self, k: usize) -> VecIntoIter<Self::Item>
-        where Self: Sized,
-              Self::Item: Ord
+    fn k_smallest(mut self, k: usize) -> VecIntoIter<Self::Item>
+    where
+        Self: Sized,
+        Self::Item: Ord,
+    {
+        // The stdlib heap has optimised handling of "holes", which is not included in our heap implementation in k_smallest_general.
+        // While the difference is unlikely to have practical impact unless `T` is very large, this method uses the stdlib structure
```
Philippe-Cholet, line 2741: Another nitpick `Self::Item` instead of `T`.

ejmount:
Hi @Philippe-Cholet 

> Since the time you worked on this, the repo has been finally rustfmt'ed (and CI reject compiler and clippy warnings) so I think the current conflicts are just about formatting. Personally, I would first squash everything and rebase on master to start clean. Or we can end by that, as you wish.

I've rebased all of this branch on top of the latest master, so that should clear up any formatting conflicts. (And that's very handy that I don't have to be careful not to check in unrelated formatting, since I run fmt by default) 

I've also done my best to address all the review comments you and @phimuemue have left, let me know if anything still needs to be changed

src/k_smallest.rs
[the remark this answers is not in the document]
ejmount: Scope has been shrunk, names changed, and lots more comments added so there's more explanation of what's going on. Hope that helps

src/k_smallest.rs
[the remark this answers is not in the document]
ejmount: Done

src/lib.rs
```diff
@@ -2950,14 +2950,109 @@ pub trait Itertools: Iterator {
...
-        crate::k_smallest::k_smallest(self, k)
-            .into_sorted_vec()
-            .into_iter()
+        // The stdlib heap has optimised handling of "holes", which is not included in our heap implementation in k_smallest_general.
+        // While the difference is unlikely to have practical impact unless `Self::Item` is very large, this method uses the stdlib structure
+        // to maintain performance compared to previous versions of the crate.
+        use alloc::collections::BinaryHeap;
+
+        if k == 0 {
+            return vec![].into_iter();
```
Philippe-Cholet, line 2964: `Vec::new()` instead of `vec![]` then the CI check should pass.

tests/test_std.rs
```diff
@@ -492,23 +492,50 @@ fn sorted_by() {
...
+
+        let largest = v.iter().cloned().k_largest(k);
+        let largest_by = v.iter().cloned().k_largest_by(k, Ord::cmp);
+        let largest_by_key = v.iter().cloned().k_largest_by_key(k, |&x| x);
+
+        // Check the variations produce the same answers and that they're right
+        for (a,b,c,d) in izip!(
+            sorted[..num_elements].iter().cloned(), 
+            smallest, 
+            smallest_by,
```
Philippe-Cholet, line 523: Format check fails in CI because of 6 trailing spaces (3 here and 3 others a few lines down).

src/k_smallest.rs
```diff
@@ -1,22 +1,97 @@
...
+    F: FnMut(&I::Item, &I::Item) -> Ordering,
+{
+    /// Sift the element currently at `origin` away from the root until it is properly ordered
+    ///
+    /// This will leave **larger** elements closer to the root of the heap
+    fn sift_down<T, F>(heap: &mut [T], is_less_than: &mut F, mut origin: usize)
+    where
+        F: FnMut(&T, &T) -> bool,
+    {
+        fn children_of(n: usize) -> (usize, usize) {
```
Philippe-Cholet, line 17: Nit: I would add `#[inline]`. We don't inline it ourselves because it has meaning to us but the compiler should.

src/k_smallest.rs
```diff
@@ -1,22 +1,97 @@
...
-    iter.for_each(|i| {
-        debug_assert_eq!(heap.len(), k);
-        // Equivalent to heap.push(min(i, heap.pop())) but more efficient.
-        // This should be done with a single `.peek_mut().unwrap()` but
-        //  `PeekMut` sifts-down unconditionally on Rust 1.46.0 and prior.
-        if *heap.peek().unwrap() > i {
-            *heap.peek_mut().unwrap() = i;
-        }
-    });
+    // Rearrange the into a valid heap by reordering from the second-bottom-most layer up to the root
```
Philippe-Cholet, line 50: "_Rearrange the storage into_" I guess.

src/k_smallest.rs
```diff
@@ -1,22 +1,97 @@
...
+    for i in (0..=(storage.len() / 2)).rev() {
+        sift_down(&mut storage, &mut is_less_than, i);
+    }
+
+    if k == storage.len() {
+        // If we fill the storage, there may still be iterator elements left
+        // So feed them into the heap
+        // Also avoids unexpected behaviour with restartable iterators
+        iter.for_each(|val| {
+            // `for_each` is potentially more performant for deeply nested iterators, see its docs.
```
Philippe-Cholet, line 62: If curious, this is the whole point of #755.
And I wrote a "CONTRIBUTING.md" that should cover this good enough (or it should be updated there).
So I prefer to remove this comment.

src/k_smallest.rs
```diff
@@ -1,22 +1,97 @@
...
+            if is_less_than(&val, &mut storage[0]) {
```
Philippe-Cholet, line 63: `&storage[0]` is enough.
I thought clippy [`unnecessary_mut_passed`](https://rust-lang.github.io/rust-clippy/master/index.html#/unnecessary_mut_passed) would catch that, but no. Maybe because it's a closure.

src/k_smallest.rs
```diff
@@ -1,22 +1,97 @@
...
+        // Sifting over a truncated slice means that the sifting will not disturb already popped elements
+        heap = &mut heap[..last_idx];
+        sift_down(heap, &mut is_less_than, 0);
+    }
+
+    storage
+}
 
-    heap
+pub(crate) fn key_to_cmp<T, K, F>(key: F) -> impl Fn(&T, &T) -> Ordering
```
Philippe-Cholet, line 91: Nit: we don't have a guideline about `#[inline]` use but all functions I see in the stdlib that return a closure have this attribute.

Philippe-Cholet (commented):
After the changes here, I think we are good 🎉.
If it does not bother you, add some periods to sentences.

I'll update the documentation later (like links, formatting preferences).
GH history seems fine to me. However, at the end, I would prefer if you squashed your commits.

src/k_smallest.rs
ejmount: Oops, I didn't notice this was in CONTRIBUTING. Removed that comment

src/k_smallest.rs
ejmount: It doesn't even flag it if it the closure is changed to take `&I::Item` specifically. Wonder if that's a clippy bug

Philippe-Cholet: The code for the lint seems to deliberately ignore closures. I searched "closure unnecessary_mut_passed" but nothing helpful.

ejmount:
Github seems to want @phimuemue's original comment ticked off before this can land - I'm sure if there's anything I can do about that on my end given all their comments have been resolved in the meantime.

Philippe-Cholet (approved):
Thank you for this 🎉

@phimuemue It seems a previous review of yours is blocking us.
I'll update documentation of Itertools methods after this (just in case you review).

phimuemue (approved):
Thanks @Philippe-Cholet, thanks @ejmount. I have only capacity to skim this, but looks good to me. 

I saw you kept `VecIntoIter` - fine with me, although I wonder why `k_smallest` should return a different type from `min_set`. - Maybe we revisit this somewhen in the future.

Philippe-Cholet:
@phimuemue Prior to this and now, `k_smallest` returns `VecIntoIter` (change that would be a breaking change) so for me variants should do the same. And if the user collect to a vector immediately then there is no reallocation, so that's not an issue.
It's true that different return types for such similar things is a bit inconsistent.

PR #663, opened 2022-12-06
Adds `try_fold_ok` and `try_fold_options`

### LINKED ISSUE #662: Add `try_` versions of `fold_{ok, options}`
Similar to how `fold_X` is equivalent to `try_fold(init, |acc, maybe_elem| maybe_elem.map(|elem| f(acc, elem)))`[1], I'd like to introduce `try_` versions that allow for fallible computation instead (i.e. replace `.map` with `.and_then` above).

I can whip together a quick PR that adds these, shouldn't take more than 2 or 3 hours with docs plus tests.

[1]: As an aside, why aren't `fold_ok` and `fold_options` written in terms of `try_fold`? Is it a versioning issue?

khoover:
Resolves #662

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2239,9 +2239,9 @@ pub trait Itertools : Iterator {
     ///
     /// The first accumulator value is the `start` parameter.
     /// Each iteration passes the accumulator value and the next value inside `Ok`
-    /// to the fold function `f`. If this value is `Err`, then the error is
-    /// returned. Otherwise the value inside the `Ok` output is extracted and used
-    /// as the next accumulator value.
+    /// to the fold function `f`. If the output is `Err`, then the error is
+    /// returned. Otherwise the value inside the `Ok` variant is extracted and
+    /// used as the next accumulator value.
     ///
     /// For example the sequence *Ok(1), Ok(2), Ok(3)* will result in a
     /// computation like this:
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -2230,6 +2230,65 @@ pub trait Itertools : Iterator {
...
+    ///           .try_fold_ok(0, |sum, &elt| if elt >= 0 { Ok(sum + elt) } else { Err("Negative number") }),
+    ///     Err("Odd number")
+    /// );
+    /// ```
+    fn try_fold_ok<A, E, B, F>(&mut self, mut start: B, mut f: F) -> Result<B, E>
+        where Self: Iterator<Item = Result<A, E>>,
+              F: FnMut(B, A) -> Result<B, E>
+    {
+        for result_elt in self {
+            start = result_elt.and_then(|elt| f(start, elt))?;
```
scottmcm, line 2287: Is this the same as
```suggestion
            start = f(start, result_elt?)?;
```
?

khoover: Since all the error types are the same, yes.

scottmcm: I ask mostly because I'm not sure whether adding another itertools method is worth it, if the difference is between 

```rust
.try_fold_results(init, |a, x| a.blah(x))
```
and

```rust
.try_fold(init, |a, x| a.blah(x?))
```

Is there an ergonomic reason that it'd be worth having the extra method anyway?

khoover: No, I suppose once you throw `?` in there's no good reason for the `try_` variants. It really is `.try_fold(init, |a, x| f(a, x?))`.

PR #672, opened 2023-01-31
move the msrv configuration from 'clippy.toml' to 'Cargo.toml'

danieleades:
what it says on the tin.

The MSRV can now be set in Cargo.toml rather than clippy.toml. Clippy parses the field from Cargo.toml.

This has the added advantage that if you try to compile this crate with a toolchain older than the MSRV you'll get a nicer error message. Or you *would* if the MSRV wasn't so old that it doesn't make use of the `rust-version` key...

still, it's the thought the counts. This change is future-proof, and is also backwards compatible, since unrecognised keys cause warnings, but not errors.


phimuemue:
Prior to Rust 1.56, cargo emits a warning when encountering `rust-version` (see https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field), but that would be ok for me.

@jswrenn Would you be ok with that?

jswrenn:
That's fine by me.

bors r+

PR #673, opened 2023-01-31
tidy Cargo.toml

danieleades:
tidies `Cargo.toml` (especially the dev-dependencies)

Also bumps the dev dependencies

Note that there's no impact on the MSRV since the MSRV check in CI is currently not checking dev targets


phimuemue:
Nice, thanks.

bors r+

PR #674, opened 2023-01-31
add dependabot config

danieleades:
adds dependabot config for bumping github action versions.

I've excluded rust dep updates for now, simply because the dev dependencies are out of date, but will require breaking changes to update. Doesn't look like there's anything in CI right now that would catch breaking changes in the dev dependencies, so best to park that for now.


jswrenn (approved):
bors r+

PR #680, opened 2023-03-02
Feature `process_results` as an `Itertools` method

### LINKED ISSUE #557: process_results on the Itertools trait

Philippe-Cholet:
Fixes #557 
I hope it's okay for me to propose an alternative pull request to #587 because it seems stuck for some time: misleading generics names Ok/Err, unnecessary ? higher kinded lifetime.

I will be around to change things if necessary, the documentation maybe, what should I write?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -926,6 +926,43 @@ pub trait Itertools : Iterator {
         flatten_ok::flatten_ok(self)
     }
 
+    /// “Lift” a function of the values of the current iterator so as to process
+    /// an iterator of `Result` values instead.
+    ///
+    /// `processor` is a closure that receives an adapted version of the iterator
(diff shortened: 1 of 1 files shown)


phimuemue:
Thanks for this.

bors r+

PR #686, opened 2023-03-23
Impl Clone for CircularTupleWindows

### LINKED ISSUE #685: Implement Clone for CircularTupleWindows
The current implementation of `CircularTupleWindows` uses `std::iter::Cycle<I>`, which already requires the underlying iterator implement `Clone`, so this is a simple addition.

zachs18:
Fixes #685

### DIFF
--- a/src/tuple_impl.rs
+++ b/src/tuple_impl.rs
@@ -201,7 +201,7 @@ impl<I, T> FusedIterator for TupleWindows<I, T>
 /// See [`.circular_tuple_windows()`](crate::Itertools::circular_tuple_windows) for more
 /// information.
 #[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct CircularTupleWindows<I, T: Clone>
     where I: Iterator<Item = T::Item> + Clone,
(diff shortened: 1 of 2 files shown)


tests/tuples.rs
```diff
@@ -61,6 +61,30 @@ fn tuple_windows() {
...
+    assert_eq!(Some((3, 4, 5, 1)), iter.next());
+    assert_eq!(Some((3, 4, 5, 1)), iter2.next());
+    assert_eq!(Some((4, 5, 1, 2)), iter.next());
+    assert_eq!(Some((4, 5, 1, 2)), iter2.next());
+    assert_eq!(Some((5, 1, 2, 3)), iter.next());
+    assert_eq!(Some((5, 1, 2, 3)), iter2.next());
+    assert_eq!(None, iter.next());
+    assert_eq!(None, iter2.next());
+}
+
```
jswrenn, line 87: Could you make this a quickcheck test instead? You can use [`assert_equal`](https://docs.rs/itertools/latest/itertools/fn.assert_equal.html) to assert that both iterators produce identical results.

zachs18: Sure!

jswrenn (approved):
bors r+

PR #687, opened 2023-03-24
feat: add combinations_up_to

maratik123:
Add combinations_up_to to iterator functions

### DIFF
--- a/src/combinations_up_to.rs
+++ b/src/combinations_up_to.rs
@@ -7,10 +7,9 @@ use std::iter::FusedIterator;
 /// See [`.combinations_up_to()`](Itertools::combinations_up_to) for more information.
 #[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
 pub struct CombinationsUpTo<I: Iterator> {
-    end: bool,
-    k: usize,
-    orig_iter: I,
     iter: Combinations<I>,
+    orig_iter: I,
+    end: bool,
 }
 
 impl<I> Clone for CombinationsUpTo<I>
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -1524,6 +1528,59 @@ pub trait Itertools : Iterator {
...
+    ///     vec![1, 4],
+    ///     vec![2, 3],
+    ///     vec![2, 4],
+    ///     vec![3, 4],
+    ///     vec![1],
+    ///     vec![2],
+    ///     vec![3],
+    ///     vec![4],
+    ///     vec![]
+    /// ]);
```
jswrenn, line 1557: I wonder if the name `combinations_up_to` is perhaps a little misleading, since the combinations are emitted in _descending_ order of size.

src/combinations_up_to.rs
```diff
@@ -0,0 +1,89 @@
...
+{
+    type Item = Vec<I::Item>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.end {
+            None
+        } else if let Some(result) = self.iter.next() {
+            if result.is_empty() {
+                self.end = true;
+            }
```
jswrenn, line 57: Do you think we can get rid of the `end` field by checking if `self.iter.k() == 0`?

jswrenn (changes requested):

Philippe-Cholet:
I think the name `combinations_up_to` is nice. What's weird for me is that it starts with "k" items then decrease instead of starting with 0 items then increase, that would feel more natural, something like: `vec![], vec![1], vec![2], vec![3], vec![4], vec![1, 2], vec![1, 3], ...` which is not even the reverse order of the current behavior.
Then maybe we could impl DoubleEndedIterator for it.

maratik123:
> EDIT: Currently I think it's something like `(0..=k).rev().flat_map(|i| iter.clone().combinations(i))`. Is that needed?

Yes, agree with you. Seems that this PR is not needed. Closing it now.

PR #689, opened 2023-03-30
fix(unique): reserve capacity for hashmap

Millione:
Reserve the hashmap capacity for performance


Philippe-Cholet:
So `HashMap::with_capacity(iter.size_hint().0.saturating_add(1))` instead of `HashMap::new()` and "style changes".
For an iterator with many equal elements, it would reserve a lot of unnecessary memory.

jswrenn (commented):
I'm in agreement with @Philippe-Cholet. For some use cases, this could be a severe pessimization of memory usage.

PR #692, opened 2023-04-13
Fix max_set_by_key documentation

jakoritarleite:
The `max_set_by_key` method has little error in the documentation, it says the method returns all mininum elements, when it should say it returns all maximum elements.


phimuemue:
Good catch. Thank you.

PR #693, opened 2023-04-14
Add example and improve docs for EitherOrBoth::reduce

2ndDerivative:
Also changed "product" to avoid confusion with multiplication.

This was to address the confusion that led to PR #690 which can be closed in my opinion if this change is added instead. 
Hit me up for any feedback!


2ndDerivative:
I added another signature because I commited from a different machine than usual, please just squash these three commits

phimuemue:
Thanks for this.

bors r+

PR #708, opened 2023-06-26
Change the return type of `at_most_one()` to `AtMostOneResult`

### LINKED ISSUE #707: `at_most_one` should have a different return type
`at_most_one()` validates that an iterator has no more than one element, then returns it if the iterator had exactly one element, or some result that indicates that the iterator was empty. If it had more than one element, the original iterator is "restored" and returned back to the user. This is a useful concept that I occasionally need.

That said, I believe the API of the `itertools` implementation is not as ergonomic as it could be. The existing implementation (#523) is based on `Itertools::exactly_one()` and returns a `Result<Option<Self::Item>, ExactlyOneError<Self>>`, where `ExactlyOneError` yields the same elements as the original iterator and also implements `Error`.

This return type has 2 flaws that keep me from using this function: 
1. The "good" result (`Ok()`) is not necessarily that the iterator had one element *or* zero elements. I often want to check that the iterator had exactly one element (and use it), but also behave differently in the zero elements case and in the many elements case (so in that case, zero elements is also "bad"). Similarly, having more than one element is not necessarily the *only* error case, it's just a different case. In other words, the zero/one/more than one distinction is better represented with three states and not two. 
2. The return type uses `Result<Option>`, which is somewhat annoying to use.

I purpose changing the return type to something like
```rust
enum AtMostOneResult<I: Iterator> {
    Zero,
    One(I::Item),
    MoreThanOne(MoreThanOne<I>),
}
```
where `MoreThanOne` yields the same elements as the original iterator. 

I have implemented this feature (+ tests, docs) and will open a PR. #702 talks about making breaking changes, so this may be an appropriate time for this.

pickx:
Fixes #707.

I hope this implementation is fine - any suggestions or changes are welcome.

### DIFF
--- a/src/at_most_one.rs
+++ b/src/at_most_one.rs
@@ -6,7 +6,7 @@ use core::mem;
 /// elements in the iterator.
 /// 
 /// See [`.at_most_one()`](crate::Itertools::at_most_one) for more detail.
-#[derive(PartialEq, Eq, Clone)]
+#[derive(PartialEq, Eq, Clone, Debug)]
 pub enum AtMostOneResult<I: Iterator> {
     /// The iterator was empty and therefore had zero elements.
     Zero,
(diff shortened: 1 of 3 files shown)


phimuemue:
I'm hesitant to merge this (please see https://github.com/rust-itertools/itertools/issues/707#issuecomment-1607135881 for details).

phimuemue:
Closed as per #707.

PR #711, opened 2023-07-02
Merge `Merge/MergeBy/MergeJoinBy` implementations

### LINKED ISSUE #701: Merge `MergeBy` and `MergeJoinBy`
It seems that `merge_join_by`/`MergeJoinBy` is actually a more generic variant of `merge`/`MergeBy`.

However, currently, the `MergeJoinBy` and `MergeBy` are implemented independently (`PutBack` vs `Peekable`, fused behavior, requiring `Ordering` vs `bool`). The past showed that uniform implementations may be better.

This change should not introduce breaking changes for users: `merge`, `merge_by` and `merge_join_by` should stay usable as they are. 

However, I'd imagine that the general `merge_join_by` could also accept a comparator returning `bool` (instead of `Ordering`) - this would free the users from having to deal with the `EitherOrBoth::Both` case. I.e. I suggest that users should be able to write:

```
use itertools; // 0.10.5
use itertools::{Either, EitherOrBoth};

fn main() {
    let numbers = [1,2,3,4,5];
    let strings = ["one", "two", "three", "four"];
    
    for x in itertools::merge_join_by(
        numbers,
        strings,
        |number, string| number.cmp(&string.len()),
    ) {
        match x {
            EitherOrBoth::Left(number) => {dbg!(number);},
            EitherOrBoth::Right(string) => {dbg!(string);},
            EitherOrBoth::Both(number, string) => {dbg!(number, string);},
        };
    }
    
    for x in itertools::merge_join_by(
        numbers,
        strings,
        |number, string| number < &string.len(), // <-- Note that we return bool instead of Ordering
    ) {
        match x {
            Either::Left(number) => {dbg!(number);},
            Either::Right(string) => {dbg!(string);},
            // <-- Note that we do not need to deal with the "Both" case, as our comparator returns bool
        };
    }
}
```

When we tackle this, we should - as a first step - probably move all the `merge`-related functions into an own module.

Philippe-Cholet:
`Merge` and `MergeBy` are moved from "adaptors" to "merge_join" modules.
Then `Merge`, `MergeBy` and `MergeJoinBy` are now instances of a new (common and private) struct: `InternalMergeJoinBy`.
Internally, instead of `OrderingOrBool` and `FnMut(&L, &R) -> Out`, we use `MergePredicate<L, R, Out=...>` (`Out` being `Ordering` or `bool`).
And with `MergeFuncT` and `MergeFuncLR` also kept private, the functions return those types:

    merge                      -> InternalMergeJoinBy<I, J, MergeFuncT<MergeLte>>
    merge_by                   -> InternalMergeJoinBy<I, J, MergeFuncT<F>>
    merge_join_by(...bool)     -> InternalMergeJoinBy<I, J, MergeFuncLR<F, bool>>
    merge_join_by(...Ordering) -> InternalMergeJoinBy<I, J, MergeFuncLR<F, Ordering>>

The only changes for an external user:

- `MergeJoinBy` sadly requires now a fourth type (it was that or share the internal `MergeFuncLR` publicly).
- `MergeJoinBy` now implements `FusedIterator` too.

Closes #701 

**PS:** In `fn merge_join_by`, the constraint `T: OrderingOrBool<I::Item, J::Item>,` has been removed. Even if the documentation is clear, we could add it back in some way if you want.


phimuemue:
Hi @Philippe-Cholet, thanks for this. I try to pick this up, but I am very unsure that `MergeJoinBy` *really* needs a fourth type parameter, so I for now only cherry-picked the first commit here.

Philippe-Cholet:
@phimuemue A bit disappointed, I started again and to my own surprise I quickly found a way (fresh eyes I guess). So https://github.com/rust-itertools/itertools/pull/711/commits/0a3f122a58cb57c478b4813887e1cfc5a08c0e25
It results in a big type definition but it works.

Philippe-Cholet:
I thought I could use `MergePredicate` instead of my new (private) trait `FuncLR` but it complains when running tests: "_the trait bound `for<'a, 'b> fn(&'a u8, &'b u8) -> bool {<u8 as PartialOrd>::ge}: itertools::merge_join::MergePredicate<u8, u8>` is not satisfied_".
I think it would require to add different lifetimes to `MergePredicate` trait or implementations. I won't unless you think it's better than create this quick `FuncLR` trait.

Philippe-Cholet:
I'm closing this. This is the result of some useful experiments but became too messy to be properly reviewed and it deserves a cleaner PR I'm currently working on.
At least, the first commit has been cherry-picked.

PR #713, opened 2023-07-13
Add conversion into (Option<A>,Option<B>) to EitherOrBoth

JustusFluegel:
Hey!

I recently found a conversion into `(Option<A>,Option<B>)` quite useful. Just wondered if there was any interest adding this directly to upstream here. If not, just close this pr :)

Thanks

### DIFF
--- a/src/either_or_both.rs
+++ b/src/either_or_both.rs
@@ -507,6 +507,7 @@ impl<A, B> Into<Option<Either<A, B>>> for EitherOrBoth<A, B> {
 }
 
 impl<A, B> From<EitherOrBoth<A, B>> for (Option<A>, Option<B>) {
+    #[inline(always)]
     fn from(val: EitherOrBoth<A, B>) -> Self {
         val.left_and_right()
     }
--- a/src/either_or_both.rs
+++ b/src/either_or_both.rs
@@ -494,9 +494,9 @@ impl<A, B> Into<Option<Either<A, B>>> for EitherOrBoth<A, B> {
     }
 }
 
-impl<A, B> Into<(Option<A>, Option<B>)> for EitherOrBoth<A, B> {
-    fn into(self) -> (Option<A>, Option<B>) {
-        match self.map_any(Some, Some) {
+impl<A, B> From<EitherOrBoth<A, B>> for (Option<A>, Option<B>) {
+    fn from(val: EitherOrBoth<A, B>) -> Self {
+        match val.map_any(Some, Some) {
             EitherOrBoth::Left(l) => (l, None),
             EitherOrBoth::Right(r) => (None, r),
             EitherOrBoth::Both(l, r) => (l, r),
--- a/src/either_or_both.rs
+++ b/src/either_or_both.rs
@@ -65,7 +65,7 @@ impl<A, B> EitherOrBoth<A, B> {
         }
     }
 
-    /// Return tupel of options corresponding to the left and right value respectively
+    /// Return tuple of options corresponding to the left and right value respectively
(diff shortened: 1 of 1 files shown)


Testare:
I was literally about to cut an issue for this! I've been using this library for a little while, but every time I use zip_longest I seem to bump up against the problem where I can't just operate on left if it exists and right if it exists. If not an explicit conversion, at least an "either()" function that would convert it to this type just like it has `left()`, `right()`, and `both()`.

scottmcm:
Maybe have it be an `into_option` method instead of a `From` impl?  That's what `MinMaxResult` uses for this same operation: <https://docs.rs/itertools/latest/itertools/enum.MinMaxResult.html#method.into_option>.

scottmcm:
Hmm, is there anything else that's `Into<(Option<_>, Option<_>)>`?

I understand `impl Into<Cow<str>>` parameter types and such (though I'm usually not a fan), but tuple-of-options seems like something that wouldn't really have very many types that it can accept -- the only thing that comes to mind is there's a new one from array of options, but that's only relevant in homogeneous situations.

JustusFluegel:
I mean having both and then rust delegating the into impl to the method (maybe with #[inline(always)]) wouldn't hurt either, right?

JustusFluegel:
I'll just add the method for now, I think `left_and_right` is quite nice

phimuemue:
Hi there, thanks a lot for this.

Please remove the `impl From`, and just keep the `fn left_and_right`, which is imho nicely aligned with the existing `fn left` and `fn right`.

_Side note:_ I guess we'd ideally want to be able to write something like `eitherorboth.map_or_default(Some)`, but this is - to the best of my knowledge - impossible for now. If anybody has an idea how we can make this work in stable rust, please leave a note.

scottmcm:
*Re the side note*: You're thinking that in terms of needing a generic closure, so that `Some` could be both `A -> Option<A>` *and* `B -> Option<B>`?

phimuemue:
> _Re the side note_: You're thinking that in terms of needing a generic closure, so that `Some` could be both `A -> Option<A>` _and_ `B -> Option<B>`?

Exactly. Admittedly, Rust allows me to define this:

```
impl<L, R> EitherOrBoth<L, R> {
    fn map_or_default<F, L2, R2>(self, mut f: F) -> (L2, R2)
        where
            F: FnMut(L)->L2 + FnMut(R)->R2,
            L2: Default,
            R2: Default,
    {
        match self {
            EitherOrBoth::Left(l) => (f(l), Default::default()),
            EitherOrBoth::Right(r) => (Default::default(), f(r)),
            EitherOrBoth::Both(l, r) => (f(l), f(r)),
        }
    }
}
```

But the above function cannot be called (at least I did not find any straightforward way). (In stable Rust.)

Philippe-Cholet:
@phimuemue I suppose we could go with `eitherorboth.map_or_default(Some, Some)` instead. After all, we could theorically want to do `eitherorboth.map_or_default(Some, Stuff)`.

Then, looking at `or_default` and without the `.map_or_default(Some, Some)` above, I just realized that it could be this below, right?!

    pub fn left_and_right(self) -> (Option<A>, Option<B>) {
        self.map_any(Some, Some).or_default()
    }

JustusFluegel:
I don't know if that would be a good idea but we could theoretically change the `left()` and `right()` impl to use `self.left_and_right().0` and `self.left_and_right().1` respectively.

phimuemue:
Thank you. Hope your exam went well.

bors r+

Issue #579, opened 2021-10-05
`map_err` and other error-only methods


Kinrany:
Feature request: add methods like `map_err` that work on errors only, similar to `map_ok`.

To match the existing `_ok` methods, that would be:

- `filter_map_err`
- `filter_err`
- `flatten_err`
- `fold_err`
- `map_err`

Out of these five, `map_err` and `filter_err` seem the most obviously useful.

PR #714, opened 2023-07-23
Add `map_err` and `err_into` iterator adapters

khollbach:
This PR adds `map_err` and `err_into` methods to `Itertools`.
* `.map_err(f)` is just like `.map_ok(f)`, but it applies `f` to each _`Result::Err`_ value.
* `.err_into()` is simply `.map_err(Into::into)`.

These methods already exist on [`Stream`](https://docs.rs/futures/0.3.28/futures/stream/trait.Stream.html)s ("async iterators") in [`TryStreamExt`](https://docs.rs/futures/0.3.28/futures/stream/trait.TryStreamExt.html#method.err_into), and I think the ergonomics are quite pleasant. E.g., the following pattern becomes less cumbersome:
```rust
my_stream.map(|result| result.map_err(Into::into))
my_stream.map_err(Into::into) // nice
my_stream.err_into() // even nicer
```

---

I've included tests for iterator specialization, and doctests for both methods. Implementation details live in `crate::apaptors::map`, alongside `MapOk` and `MapInto`.

Let me know if there's any feedback or concerns; or if this feature is just not needed.

Thanks!

### DIFF
--- a/src/adaptors/map.rs
+++ b/src/adaptors/map.rs
@@ -157,3 +157,38 @@ where
         r.map_err(|v| self.0(v))
     }
 }
+
+/// An iterator adapter to convert a nested `Result::Err` using [`Into`].
+///
+/// See [`.map_err()`](crate::Itertools::map_err) for more information.
+pub type ErrInto<I, F> = MapSpecialCase<I, MapSpecialCaseFnErrInto<F>>;
+
+/// Create a new `ErrInto` iterator.
+pub(crate) fn err_into<I, F, T, E, E2>(iter: I) -> ErrInto<I, F>
+where
+    I: Iterator<Item = Result<T, E>>,
+    E: Into<E2>,
+{
+    MapSpecialCase {
+        iter,
+        f: MapSpecialCaseFnErrInto(PhantomData),
+    }
+}
+
+#[derive(Clone)]
+pub struct MapSpecialCaseFnErrInto<E2>(PhantomData<E2>);
+
+impl<F> std::fmt::Debug for MapSpecialCaseFnErrInto<F> {
+    debug_fmt_fields!(MapSpecialCaseFnErrInto,);
+}
+
+impl<T, E, E2> MapSpecialCaseFn<Result<T, E>> for MapSpecialCaseFnErrInto<E2>
+where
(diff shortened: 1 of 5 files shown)


khollbach:
Related issue: https://github.com/rust-itertools/itertools/issues/579

Zynh0722:
Not entirely sure about procedure, but is there a good reason not to merge this sort of thing?

particularly `map_err`

Zynh0722:
That sounds rather reasonable haha. I was sure there'd be good reason for it not already existing considering `map_ok` does exist. 

It could definitely just be me not being super comfortable with iterators of results. I mostly went looking for it because I am janking anyhow errors around in some application code I wrote, and I found an open PR which served to peak my curiosity.

Thank you for your input!

phimuemue:
> But I'd also be tempted just to not have either, really.

+1

Tying these special case `map_err` (and cousins) to `Iterator` makes for a less-composable API. And if there is `Iterator::map_err`, one could also argue for `Option::map_err` or (as in the proposal) `Stream::map_err`... and many others.

If you insist on one-liners, I'd prefer a function that curries `Result::map_err` so that only the `Result` is needed:

```
fn curry_map_err<U>(f: impl FnMut(E) -> E2) -> impl FnMut(Result<U, E>)->Result<U, E2> {
 move |r| r.map_err(f)
},
```

And use it for all mappings over results:

```
iterator.map(curry_map_err(f))
option.map(curry_map_err(f))
stream.map(curry_map_err(f))

// The err_into thing:
anything.map(curry_map_err(Into::into))
```

I get this is controversial, but imho we should ditch the special case mappers. @jswrenn @Philippe-Cholet What do you think?

Philippe-Cholet:
I think we might want to deprecate some things once we do #844 (we might want the current discussion there) and there is no rush to deprecate them (0.13.0 already has quite some changes).

Even if your alternative is a 1-liner, I would not call it convenient.

Just pasting some [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=8f29a9c099400bdcfdfdf277934f50b4).

phimuemue:
Closing. We established not to include this kind of method.

PR #717, opened 2023-07-26
Add a number of methods to `EitherOrBoth`

glandium:
These are methods that I found useful for my use.

I also have a
```
impl<A, B> EitherOrBoth<EitherOrBoth<A, B>, EitherOrBoth<A, B>> {
    pub fn transpose(self) -> EitherOrBoth<EitherOrBoth<A, A>, EitherOrBoth<B, B>> { ... }
}
```
that is not part of this PR because it seemed really too niche. (My own real usecase is starting from `EitherOrBoth<Either<A, B>, Either<A, B>>`)

### DIFF
--- a/src/either_or_both.rs
+++ b/src/either_or_both.rs
@@ -716,6 +716,33 @@ impl<T> EitherOrBoth<T, T> {
             Both(a, b) => f(a, b),
         }
     }
+
+    /// Equivalent to `map_any`, but using the same function for both values when
+    /// they have the same type.
+    pub fn map<F, U>(self, f: F) -> EitherOrBoth<U, U>
+    where
+        F: Fn(T) -> U
+    {
+        self.map_any(&f, &f)
+    }
+
+    /// Equivalent to `filter_any`, but using the same function for both values when
+    /// they have the same type.
+    pub fn filter<F>(self, f: F) -> Option<EitherOrBoth<T, T>>
+    where
+        F: Fn(&T) -> bool
+    {
+        self.filter_any(&f, &f)
+    }
+
+    /// Equivalent to `filter_map_any`, but using the same function for both values when
+    /// they have the same type.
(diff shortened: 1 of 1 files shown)


phimuemue:
Hi there. `Option<EitherOrBoth<L, R>>` is isomorphic to `(Option<L>, Option<R>)`, and the latter is arguably more straightforward.

We now have [`left_and_right`](https://github.com/rust-itertools/itertools/blob/e0b8a12/src/either_or_both.rs#L72) which converts an `EitherOrBoth` to `(Option<L>, Option<R>)`. As such, I'd like to close this, because all the proposed can be implemented in terms of `Option`-combinators:

* `filter_{left|right|any}`: Call `Option::filter` on the respective tuple elements.
* `filter_map_{left|right|any}`: Call `Option::and_then` on the respective tuple elements.

I know that calling the `Option`-combinators on the respective tuple elements may involve creating a temporary, but that could be mitigated by implementing methods on tuples `(T0, T1)` so that one could e.g write `tuple_containing_two_options.map_first_element(Option::filter)`.

Would your use-case prohibit `(Option<L>, Option<R>)` instead of `Option<EitherOrBoth<L, R>>`? I.e. could you use `left_and_right` in your case?

glandium:
In my case, Option<EitherOrBoth<L, R>> is more straightforward to use because `?` can be used, which leaves me with an `EitherOrBoth<L, R>`, which is ultimately what I needed.

I'll add that since I submitted this PR, I've... actually stopped needed these methods, so they might very well be very niche-y.

phimuemue:
I see, thanks for the explanation.

I close this for now and will come back to it if there seems to be more need for them.

Note to myself: Maybe we can implement ‘EitherOrBoth<Option, Option>‘  to have a method ‘ filter_out _none‘, returning the suggested type. Might compose well enough to cover the use cases.

glandium:
map, filter and filter_map on EitherOrBoth<T, T> might be more easily useful, though, to apply the same closure to both sides.

PR #719, opened 2023-07-28
Allow to use EitherOrBoth<T> to mean the same as EitherOrBoth<T, T>

glandium:


glandium:
No, I was just trying to avoid the repeated type, which can be bothersome when it's long.

jswrenn (changes requested):
There are a few places in our tests that could take advantage of this change:
https://github.com/rust-itertools/itertools/blob/0897f56ad6a305b9686b0a4447344bc8c7e5c580/tests/merge_join.rs#L59-L98

Can you modify them accordingly?

jswrenn (approved):

PR #721, opened 2023-07-31
add `{left,right}_or_{right,left}` methods to `EitherOrBoth`

Testare:
Been using this library for a bit, wanted to create some methods to cover a use case I kept coming across: I was zipping two iterators of the same type (zip_longest), taking the value from one if it was there but wanting to default to the other if not. Thought it would be a good contribution upstream! Did my homework to make sure it wasn't a duplicate of another PR, though saw other good PR's in this space.

I considered names like `either_favor_right`/`either_favor_left`, but `left_or_right`/`right_or_left` seemed the most ergonomic IMHO. It's a simple enough change if you disagree though.

### DIFF
--- a/src/either_or_both.rs
+++ b/src/either_or_both.rs
@@ -490,6 +490,44 @@ impl<T> EitherOrBoth<T, T> {
             Both(a, b) => f(a, b),
         }
     }
+
+    /// Returns either left or right, or if both are present returns right
+    ///
+    /// Complement of [`left_or_right`](EitherOrBoth::left_or_right).
+    ///
+    /// # Examples
+    /// ```
+    /// # use itertools::EitherOrBoth;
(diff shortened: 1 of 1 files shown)


Testare:
I'm fine with implementing the method that way. Reduce might cover this, though I didn't realize until a week or two before this PR that reduce was the method I was looking for. I felt like these names might better complement the methods "left", "right", and "both".

Perhaps it would be good to call out in the doc string for these methods that if you want different behavior for the both case, to look at "reduce." But if you feel this isn't a good fit for the package, I'd understand =]

phimuemue:
I updated the docs for `EitherOrBoth::reduce` in https://github.com/rust-itertools/itertools/commit/052423a8c971e1ef46890840cd662ab7ddb2f809.

PR #723, opened 2023-08-07
`try_len` method

Easyoakland:
Add method to get the length of an iterator if the `size_hint` is consistent.
Reasons why this would be useful:
1. It makes it more obvious when getting the length with `it.try_len().unwrap()` than with the match statement.
2. Iterator adapters that increase the length of iteration are not supposed to implement `ExactSizeIter` as indicated [here](https://doc.rust-lang.org/std/iter/trait.ExactSizeIterator.html#when-shouldnt-an-adapter-be-exactsizeiterator). This could be used in those cases.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3840,6 +3840,22 @@ pub trait Itertools : Iterator {
     {
         MultiUnzip::multiunzip(self)
     }
+
+    /// Returns the length of the iterator if one exists.
+    ///
+    /// ```
+    /// use itertools::Itertools;
+    ///
+    /// assert_eq!([0; 10].into_iter().try_len(), Some(10));
+    /// assert_eq!((10..15).try_len(), Some(5));
+    /// assert_eq!((15..10).try_len(), Some(0));
+    /// assert_eq!((10..).try_len(), None);
+    /// ```
+    fn try_len(self) -> Option<usize>
+    where Self: Sized
+    {
+        size_hint::try_len(self)
+    }
 }
 
 impl<T: ?Sized> Itertools for T where T: Iterator { }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3857,10 +3857,12 @@ pub trait Itertools : Iterator {
     /// assert_eq!((10..).try_len(), Err((usize::MAX, None)));
     /// assert_eq!((10..15).filter(|x| x % 2 == 0).try_len(), Err((0, Some(5))));
     /// ```
-    #[inline]
-    fn try_len(&self) -> Result<usize, size_hint::SizeHint>
-    {
-        size_hint::try_len(self.size_hint())
+    fn try_len(&self) -> Result<usize, size_hint::SizeHint> {
+        let sh = self.size_hint();
+        match sh {
+            (lo, Some(hi)) if lo == hi => Ok(lo),
+            _ => Err(sh),
+        }
     }
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3851,7 +3851,7 @@ pub trait Itertools : Iterator {
     /// assert_eq!((15..10).try_len(), Some(0));
     /// assert_eq!((10..).try_len(), None);
     /// ```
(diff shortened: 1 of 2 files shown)


Philippe-Cholet:
Reading this, I remembered the `size_hint` [doc](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.size_hint) saying it should be _fully_ trusted (bold below is mine):
> `size_hint()` is primarily intended to be used for optimizations such as reserving space for the elements of the iterator, **but must not be trusted** to e.g., omit bounds checks in unsafe code. An incorrect implementation of `size_hint()` should not lead to memory safety violations.

So maybe the documentation of `try_len` should warn about an incorrect implementation of `size_hint`.

Easyoakland:
@Philippe-Cholet 

> So maybe the documentation of `try_len` should warn about an incorrect implementation of `size_hint`.
> And to have a similar signatures to other functions in `size_hint.rs` (and not have `&impl Iterator`), I would do ...

Fixed in [952d8c4](https://github.com/rust-itertools/itertools/pull/723/commits/952d8c4c87826a0413f0ae7ed8a5aee257d1c0a9). If a length is needed for unsafe operations I believe one is supposed to bound by https://doc.rust-lang.org/std/iter/trait.TrustedLen.html

phimuemue:
Hi there. I see this might be handy in some circumstances.

* Let's return `Result<usize, (usize, Option<usize>)>` so that the callers do not need to call `size_hint` a second time in the failure case.
* Please inline `size_hint::try_len`.
* I am unsure about the name. I guess that most users looking for something like this start out by searching for `size_hint`. Maybe `unique_size_hint`?
* In the docs, I did not directly understand what "relies on" means. Maybe "inherits guarantees and restrictions from `size_hint`"?
* Do we need the `Sized` trait bound?

Easyoakland:
> Let's return Result<usize, (usize, Option<usize>)> so that the callers do not need to call size_hint a second time in the failure case.

Sure. I suppose it would save a re-computation if its useful in the error case.

> Please inline size_hint::try_len.

Either you meant `Itertools::try_len()` or I've already done that. If you meant the former that's now done as well.
How does one determine when to inline a function? I've seen the `#[inline]` annotation used before but I haven't been able to figure out when to apply it. I've generally just used it when similar functions nearby also have it or if benchmarks show an improvement.

> I am unsure about the name. I guess that most users looking for something like this start out by searching for size_hint. Maybe unique_size_hint?

I chose this name to mirror the naming as in [ExactSizeIterator](https://doc.rust-lang.org/std/iter/trait.ExactSizeIterator.html) and [here](https://github.com/rust-lang/rust/issues/34433). The primary use case is for getting the length from chaining/adapting two iterators known to not overflow a `usize` or a specialization-like behavior on `ExactSizeIterator` in a function implementation.

> In the docs, I did not directly understand what "relies on" means. Maybe "inherits guarantees and restrictions from size_hint"?

Sure.

> Do we need the Sized trait bound?

Good catch. Originally it was needed because it incorrectly took `impl Iterator`. Now that it takes the `sizeHint` we don't need the bound.

phimuemue:
thank you.

> Either you meant `Itertools::try_len()` or I've already done that.

I meant to literally eliminate size_hint::try_len by doing the computation directly in Itertools::try_len. As it stands, it only creates an indirection without real need.

Regarding ‘[inline]‘: you’re right, there’s no clear pattern in the code base, so let’s just not use it here.

Easyoakland:
> I meant to literally eliminate size_hint::try_len by doing the computation directly in Itertools::try_len. As it stands, it only creates an indirection without real need.
> 
> Regarding ‘[inline]‘: you’re right, there’s no clear pattern in the code base, so let’s just not use it here.

fixed in [068de35](https://github.com/rust-itertools/itertools/pull/723/commits/068de35d95d345ea6bf112ecdf8de504cdb3b334)

phimuemue:
Thanks.

bors r+

Issue #722, opened 2023-08-01
Iterators generating (many) vectors


Philippe-Cholet:
## Problem: slow because of repetitive heap-allocation
The 5 iterators below generate `Vec<...>` items.

	multi_cartesian_product(self)
	combinations(self, k: usize)
	combinations_with_replacement(self, k: usize)
	permutations(self, k: usize)
	powerset(self)
It can be slow, because of the repetitive allocations, even if the user does not need to permanently store each permutation.

I think that in most cases, the user uses each vector to produce something else and then each vector is discarded. In that case, because those iterators generate (up to) many many items (product of the lengths, factorial-related, powers of two), it creates **many** temporary vectors which can slow down code execution.

## Proposed solution: map slices
An alternative iterator could own one single (internal/private) vector and only give a reference to it to a function implementing `FnMut(&[...]) -> R`. Those iterators would then generate `R` items instead of vectors.

	// current methods unchanged, new methods:
	multi_cartesian_product_map<R, F>(self, func: F)
	combinations_map<R, F>(self, k: usize, func: F)
	combinations_with_replacement_map<R, F>(self, k: usize, func: F)
	permutations_map<R, F>(self, k: usize, func: F)
	powerset_map<R, F>(self, func: F)
	// all with  F: FnMut(&[...]) -> R

### Usage examples
`it.combinations(k)` and `it.combinations_map(k, ToOwned::to_owned)` would produce the same elements, but the first should be used!

`it.combinations(k).map(closure)` and `it.combinations_map(k, closure)` could produce the same elements if it does not require a vector. For what I tested, it's then roughly **_4 to 5 times faster_** to go through all combinations, unless the iterator has very few elements (then `.map` should be used).

In `it.permutations(5).filter_map(|perm: Vec<_>| some_option)`, if `some_option` does not need perm to be an owned vector then it could become `it.permutations_map(5, |perm: &[_]| some_option).flatten()` where there is no heap-allocation for each permutation.

### Implementation (**[commits here](https://github.com/rust-itertools/itertools/compare/master...Philippe-Cholet:vec_items)**)
New module `vec_items` (behind `use_alloc` feature) with:
- `pub trait VecItems<T> {  type Output;  fn new_item<I: Iterator<Item = T>>(&mut self, iter: I) -> Self::Output;  }`
- `pub struct CollectToVec;` implementing `VecItems` with `Vec<T>` output and just `iter.collect()` (so a new vector each time, as it's currently done).
[...]

PR #725, opened 2023-08-12
More `size_hint` and `count` methods

Philippe-Cholet:
As I intend to improve the `permutations` algorithm (related to #722), I first wanted to simplify a few things:
- `CompleteStateRemaining` basically stands for `Option<usize>` without its usuability. A doc comment seems enough.
- `prefill` is way clearer to see if there are enough values.
- `CompleteState::count` method rather than an one-off function.
- `0..=x` instead of `0..(x+1)` ; one more fold.

Then I encountered a TODO that I fixed:
It makes sense to have a `LazyBuffer::size_hint` method which greatly helps in fixing `Permutations::size_hint` and therefore correctly allocate the memory for codes like `(0..n).permutations(k).map(...).collect_vec()`.
I think we need to be able to apply a function to both bounds of a `SizeHint`. Here, it avoids repetition and therefore readability.
Then permutations' size hints are tested.

### DIFF
--- a/src/combinations.rs
+++ b/src/combinations.rs
@@ -1,5 +1,5 @@
 use std::fmt;
-use std::iter::FusedIterator;
+use std::iter::{Fuse, FusedIterator};
 
 use super::lazy_buffer::LazyBuffer;
 use super::size_hint::{self, SizeHint};
@@ -61,7 +61,7 @@ impl<I: Iterator> Combinations<I> {
 
     /// Returns a reference to the source iterator.
     #[inline]
-    pub(crate) fn src(&self) -> &I { &self.pool.it }
+    pub(crate) fn src(&self) -> &Fuse<I> { &self.pool.it }
 
     /// Resets this `Combinations` back to an initial state for combinations of length
     /// `k` over the same pool data source. If `k` is larger than the current length
--- a/src/combinations.rs
+++ b/src/combinations.rs
@@ -2,6 +2,7 @@ use std::fmt;
 use std::iter::FusedIterator;
 
 use super::lazy_buffer::LazyBuffer;
+use super::size_hint::{self, SizeHint};
 use alloc::vec::Vec;
 
 /// An iterator to iterate through all the `k`-length combinations in an iterator.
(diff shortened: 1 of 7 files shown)


Philippe-Cholet:
Size hints for `combinations_with_replacement`.
While I'm on it, I will do the same for `combinations`.
Currently, CI fails with "_error: failed to get `syn` as a dependency of package `wasm-bindgen-macro-support v0.2.87`_"

Philippe-Cholet:
I thought I was done but I can implement `count` methods for `permutations`, `combinations[_with_replacement]` and probably `powerset`.

Philippe-Cholet:
I just noticed that a few `.fold(Some(...), function_that_add_or_multiply)` could become `.map(func).sum/product()` once the MSRV (currently 1.36.0) is 1.37.0 or more.

phimuemue:
Would you mind doing the lazy buffer in a separate PR?

Philippe-Cholet:
All depends on that but ok.

phimuemue:
I don’t want to force you to do anything, but I see that the PRs are generally good, but broad. Having smaller PRs helps tremendously with reviewing, and I hope that we can merge them faster this way.

Philippe-Cholet:
I'm willing to learn how to do things in any better way. The first commit here is big because of I previously started from an outdated master branch, realized that later on and because I was unsure what to do with git, I started again and it resulted in a big commit I did not have the heart to split. But I agree that my first commit is broad, I was a bit overzealous.

phimuemue:
> I'm willing to learn how to do things in any better way. The first commit here is big because of I previously started from an outdated master branch, realized that later on and because I was unsure what to do with git, I started again and it resulted in a big commit I did not have the heart to split. But I agree that my first commit is broad, I was a bit overzealous.
> EDIT: New PR #727

No worries! I really want to include your ideas, but do not have much time at hand, so I try to streamline the review process (admittedly: streamline it on my side) - reviewing (very) small-bite chunks separately is usually easier than looking reviewing the "sum of the chunks" at once.

Philippe-Cholet:
Well I'm closing this, it became too broad but it will make 4 or 5 nice smaller PR. Let's consider this as a draft.

PR #727, opened 2023-08-15
Fuse the iterator in `LazyBuffer`

Philippe-Cholet:
Fuse the iterator in `LazyBuffer`.
@phimuemue Do I push the new `LazyBuffer::size_hint` here?
EDIT: It could also have a "fill" method `pub fn fill(&mut self) { self.buffer.extend(self.it.by_ref()) }` that would later be better than `while self.get_next() {}`. I did not created it before but would have a use for it basically to then know the real length in "count" methods.

### DIFF
--- a/src/combinations.rs
+++ b/src/combinations.rs
@@ -1,5 +1,5 @@
 use std::fmt;
-use std::iter::FusedIterator;
+use std::iter::{Fuse, FusedIterator};
 
 use super::lazy_buffer::LazyBuffer;
 use alloc::vec::Vec;
@@ -54,7 +54,7 @@ impl<I: Iterator> Combinations<I> {
 
     /// Returns a reference to the source iterator.
     #[inline]
-    pub(crate) fn src(&self) -> &I { &self.pool.it }
+    pub(crate) fn src(&self) -> &Fuse<I> { &self.pool.it }
 
     /// Resets this `Combinations` back to an initial state for combinations of length
     /// `k` over the same pool data source. If `k` is larger than the current length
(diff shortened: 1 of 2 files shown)


src/combinations.rs
```diff
@@ -54,7 +54,7 @@ impl<I: Iterator> Combinations<I> {
 
     /// Returns a reference to the source iterator.
     #[inline]
-    pub(crate) fn src(&self) -> &I { &self.pool.it }
+    pub(crate) fn src(&self) -> &Fuse<I> { &self.pool.it }
```
phimuemue, line 57: This function is only used in `Powerset::size_hint`, where we ask for `src().size_hint` - I think that using the size_hint of the fused iterator is actually what we want there. Can you confirm this?

I guess that `Powerset::size_hint` should - in the future - also redirect to `LazyBuffer::size_hint`, no? Then we could get rid of this `src` exposing the innards of `LazyBuffer` entirely

phimuemue (approved):
Hi @Philippe-Cholet! Thanks *a lot* for your engagement.

Having this as a separate PR simplifies review a lot. Could you shed some light on my question regarding the `Powerset::size_hint`? I'd merge this if you agree with both.

src/combinations.rs
Philippe-Cholet: Yes entirely, I confirm that `Powerset::size_hint` wants the not yet written `LazyBuffer::size_hint` from the wrapped `Combinations`.
`fn src` might become `fn pool(&self) -> &LazyBuffer<I>` later in order for `Powerset::size_hint` to `let src_total = self.combs.pool().size_hint();` instead of doing the job of `LazyBuffer::size_hint`.

PR #728, opened 2023-08-16
Do not update max key/value pair on equal elements.

ttsugriy:
This avoids the need to use negation and avoids max key/value pair updates for equal elements.

Also added a test coverage gap.


scottmcm:
You might have realized this since you closed the PR, but <https://doc.rust-lang.org/std/cmp/fn.max.html> intentionally returns the second if they're `Ordering::Equal` (and `min` returns the first), so it's probably intentional the the max is the last one in the iterator if they're all equivalent.

ttsugriy:
yes, sorry for the hassle @scottmcm and thank you for working on such an awesome project!

PR #729, opened 2023-08-16
combinations: `count` and `size_hint`

Philippe-Cholet:
The indices we track (to give combinations in the lexicographic order) can help us count the remaining combinations.

@phimuemue That's all for combinations. Each commit is as small as possible. Similar work for other iterators to come, but this should be the biggest as it prepares work for `powerset` and `combinations_with_replacement`.

### DIFF
--- a/src/combinations.rs
+++ b/src/combinations.rs
@@ -185,6 +185,9 @@ fn remaining_for(n: usize, first: bool, indices: &[usize]) -> Option<usize> {
         indices
             .iter()
             .enumerate()
+            // TODO: Once the MSRV hits 1.37.0, we can sum options instead:
+            // .map(|(i, n0)| checked_binomial(n - 1 - *n0, k - i))
+            // .sum()
             .fold(Some(0), |sum, (i, n0)| {
                 sum.and_then(|s| s.checked_add(checked_binomial(n - 1 - *n0, k - i)?))
             })
--- a/src/combinations.rs
+++ b/src/combinations.rs
@@ -158,7 +158,9 @@ pub(crate) fn checked_binomial(mut n: usize, mut k: usize) -> Option<usize> {
 /// For a given size `n`, return the count of remaining combinations or None if it would overflow.
 fn remaining_for(n: usize, first: bool, indices: &[usize]) -> Option<usize> {
     let k = indices.len();
-    if first {
+    if n < k {
+        Some(0)
+    } else if first {
         checked_binomial(n, k)
     } else {
         indices
--- a/src/combinations.rs
+++ b/src/combinations.rs
@@ -77,21 +77,6 @@ impl<I: Iterator> Combinations<I> {
             self.pool.prefill(k);
         }
     }
-
-    /// For a given size `n`, return the count of remaining elements or None if it would overflow.
-    fn remaining_for(&self, n: usize) -> Option<usize> {
-        let k = self.k();
-        if self.first {
-            checked_binomial(n, k)
-        } else {
-            self.indices
-                .iter()
-                .enumerate()
-                .fold(Some(0), |sum, (k0, n0)| {
-                    sum.and_then(|s| s.checked_add(checked_binomial(n - 1 - *n0, k - k0)?))
-                })
-        }
-    }
 }
 
 impl<I> Iterator for Combinations<I>
@@ -138,14 +123,15 @@ impl<I> Iterator for Combinations<I>
 
     fn size_hint(&self) -> (usize, Option<usize>) {
         let (mut low, mut upp) = self.pool.size_hint();
-        low = self.remaining_for(low).unwrap_or(usize::MAX);
-        upp = upp.and_then(|upp| self.remaining_for(upp));
+        low = remaining_for(low, self.first, &self.indices).unwrap_or(usize::MAX);
+        upp = upp.and_then(|upp| remaining_for(upp, self.first, &self.indices));
         (low, upp)
     }
 
-    fn count(mut self) -> usize {
-        self.pool.fill();
-        self.remaining_for(self.n()).expect("Iterator count greater than usize::MAX")
+    fn count(self) -> usize {
+        let Self { indices, pool, first } = self;
+        let n = pool.len() + pool.it.count();
+        remaining_for(n, first, &indices).expect("Iterator count greater than usize::MAX")
     }
 }
 
@@ -166,4 +152,19 @@ pub(crate) fn checked_binomial(mut n: usize, mut k: usize) -> Option<usize> {
         n -= 1;
     }
     Some(c)
+}
+
+/// For a given size `n`, return the count of remaining combinations or None if it would overflow.
+fn remaining_for(n: usize, first: bool, indices: &[usize]) -> Option<usize> {
+    let k = indices.len();
+    if first {
+        checked_binomial(n, k)
+    } else {
+        indices
+            .iter()
+            .enumerate()
+            .fold(Some(0), |sum, (k0, n0)| {
+                sum.and_then(|s| s.checked_add(checked_binomial(n - 1 - *n0, k - k0)?))
+            })
+    }
 }
--- a/src/combinations.rs
+++ b/src/combinations.rs
@@ -130,8 +130,9 @@ impl<I> Iterator for Combinations<I>
 
     fn count(self) -> usize {
         let Self { indices, pool, first } = self;
+        // TODO: make `pool.it` private
         let n = pool.len() + pool.it.count();
-        remaining_for(n, first, &indices).expect("Iterator count greater than usize::MAX")
+        remaining_for(n, first, &indices).unwrap()
     }
 }
 
--- a/src/combinations.rs
+++ b/src/combinations.rs
@@ -135,6 +135,11 @@ impl<I> Iterator for Combinations<I>
         // Create result vector based on the indices
         Some(self.indices.iter().map(|i| self.pool[*i].clone()).collect())
     }
+
+    fn count(mut self) -> usize {
+        self.pool.fill();
+        self.remaining_for(self.n()).expect("Iterator count greater than usize::MAX")
+    }
 }
 
 impl<I> FusedIterator for Combinations<I>
--- a/src/combinations.rs
+++ b/src/combinations.rs
@@ -136,6 +136,13 @@ impl<I> Iterator for Combinations<I>
         Some(self.indices.iter().map(|i| self.pool[*i].clone()).collect())
     }
 
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let (mut low, mut upp) = self.pool.size_hint();
+        low = self.remaining_for(low).unwrap_or(usize::MAX);
+        upp = upp.and_then(|upp| self.remaining_for(upp));
+        (low, upp)
+    }
+
     fn count(mut self) -> usize {
         self.pool.fill();
         self.remaining_for(self.n()).expect("Iterator count greater than usize::MAX")
--- a/src/combinations.rs
+++ b/src/combinations.rs
@@ -126,3 +126,17 @@ impl<I> FusedIterator for Combinations<I>
     where I: Iterator,
           I::Item: Clone
 {}
+
+pub(crate) fn checked_binomial(mut n: usize, mut k: usize) -> Option<usize> {
+    if n < k {
+        return Some(0);
+    }
+    // `factorial(n) / factorial(n - k) / factorial(k)` but trying to avoid it overflows:
+    k = (n - k).min(k); // symmetry
+    let mut c = 1;
+    for i in 1..=k {
+        c = (c / i).checked_mul(n)?.checked_add((c % i).checked_mul(n)? / i)?;
+        n -= 1;
+    }
+    Some(c)
+}
--- a/src/combinations.rs
+++ b/src/combinations.rs
@@ -156,6 +156,25 @@ pub(crate) fn checked_binomial(mut n: usize, mut k: usize) -> Option<usize> {
     Some(c)
 }
 
+#[test]
+fn test_checked_binomial() {
+    // With the first row: [1, 0, 0, ...] and the first column full of 1s, we check
+    // row by row the recurrence relation of binomials (which is an equivalent definition).
+    // For n >= 1 and k >= 1 we have:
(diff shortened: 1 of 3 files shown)


src/lazy_buffer.rs
```diff
@@ -23,6 +25,10 @@ where
         self.buffer.len()
     }
 
+    pub fn size_hint(&self) -> SizeHint {
+        size_hint::add_scalar(self.it.size_hint(), self.len())
```
phimuemue, line 29: Looks reasonable.

src/combinations.rs
```diff
@@ -126,3 +126,17 @@ impl<I> FusedIterator for Combinations<I>
...
+    }
+    // `factorial(n) / factorial(n - k) / factorial(k)` but trying to avoid it overflows:
+    k = (n - k).min(k); // symmetry
+    let mut c = 1;
+    for i in 1..=k {
+        c = (c / i).checked_mul(n)?.checked_add((c % i).checked_mul(n)? / i)?;
+        n -= 1;
+    }
+    Some(c)
+}
```
phimuemue, line 169: Could we at least exhausively test this implementation? (I.e. generate a lookup-table and check the results?)

Or even implement this via a lookup table? (Not urgent imho.)

src/combinations.rs
```diff
@@ -135,6 +135,11 @@ impl<I> Iterator for Combinations<I>
         // Create result vector based on the indices
         Some(self.indices.iter().map(|i| self.pool[*i].clone()).collect())
     }
+
+    fn count(mut self) -> usize {
+        self.pool.fill();
```
phimuemue, line 147: Do we really need to do `self.pool.fill`? Couldn't we just `let n = self.pool.len()+self.pool.iter.count()` (avoiding copying of elements) and then call `self.remaining_for(n)`?

As a consequence: Do we need `LazyBuffer::fill`?

src/combinations.rs
```diff
@@ -77,6 +77,21 @@ impl<I: Iterator> Combinations<I> {
             self.pool.prefill(k);
         }
     }
+
+    /// For a given size `n`, return the count of remaining elements or None if it would overflow.
+    fn remaining_for(&self, n: usize) -> Option<usize> {
+        let k = self.k();
+        if self.first {
+            checked_binomial(n, k)
```
phimuemue, line 85: I *think* I understand this one.

src/combinations.rs
```diff
@@ -77,6 +77,21 @@ impl<I: Iterator> Combinations<I> {
...
+        } else {
+            self.indices
+                .iter()
+                .enumerate()
+                .fold(Some(0), |sum, (k0, n0)| {
+                    sum.and_then(|s| s.checked_add(checked_binomial(n - 1 - *n0, k - k0)?))
+                })
```
phimuemue, line 92: But I'm pretty sure I *don't trivially* understand this one. Can you shed some light on this or can anyone co-review?

src/combinations.rs
```diff
@@ -77,6 +77,21 @@ impl<I: Iterator> Combinations<I> {
             self.pool.prefill(k);
         }
     }
+
+    /// For a given size `n`, return the count of remaining elements or None if it would overflow.
+    fn remaining_for(&self, n: usize) -> Option<usize> {
```
phimuemue, line 82: Should this really be a method of `Combinations` and accept `n` (as `n` is given by `Combinations::n`)?

phimuemue (changes requested):
Hi @Philippe-Cholet I tried to review, and I have some questions.

I think the most difficult part for me is to understand `remaining_for`.

src/combinations.rs
Philippe-Cholet: Yes, I just was not sure how. I mean that I don't think we want it public so not in the tests folder. So a doctest or a `#[test]` function which is not a thing we can find in the entire library, maybe there is a reason.

phimuemue: Maybe a hardwired two-dimensional array of `u128` with a reference from where we took it, and then simply iterate over all pairs of supported `k`/`n` and assert that the `usize::try_from(lookuptable_holding_u128[n][k])` equals the result from the function?

src/combinations.rs
Philippe-Cholet: Well `Combinations::size_hint` also calls it on `low` and `upp` so not just `Combinations::n`.

src/combinations.rs
phimuemue: Couldn't `remaining_for` also be a free function (possibly accepting more arguments as required)?

src/combinations.rs
Philippe-Cholet: Well **this is definitely the hard part**, I'm not gonna say it's trivial, even for me.
The `indices` we track are basically the the k-combinations of `0..n`.
I worked on an example, let's say k=3 and n=5 (big enough to see things properly without having too many elements).
The indices are then (one line per combination):

    0 1 2
    0 1 3
    0 1 4
    0 2 3
    0 2 4
    0 3 4
    1 2 3
    1 2 4
    1 3 4
    2 3 4

There are `binomial(5, 3)==10` elements: **6** elements starting with 0 (`binomial(4,2)` to choose 2 elements from `1..5`), **3** with 1 (`binomial(3,2)` to choose 2 elements from `2..5`), **1** with 2 (`binomial(2,2)` to choose 2 elements from `3..5`).

The binomials (rows for n, columns for k, both starting at 0).

    1
    1 1
    1 2 *1*
    1 3 *3* 1
    1 4 *6* 4 1
    1 5 10 *10* 5 1

Now, let's say we have `indices = 0 1 4` and we want to count the remaining ones. So
- the only one starting with 2 `2 3 4`
- the ones starting with 1 (`1 2 3, 1 2 4, 1 3 4` start with 1 and the rest: choose 2 from 2..5 so `binomial(5-2, 2)`)
- and then for the ones starting with 0, some other binomial**s**:
  - the ones for 3 in 2nd position: `0 3 4` (`binomial(1, 1)` as the only choice is 4)
  - the ones for 2 in 2nd position: `0 2 3, 0 2 4` (`binomial(2, 1)` to choose 1 elements from `3..5`)

So we have a sum of four binomials while we sum `k=3` binomials in the code so you could be concerned. But there is a trick (binomials are fun). See the above triangle where the column 1 3 6 sums into 10, that's a property of the binomials that simplifies the sum into the one in the code. For the example above: `binomial(1,1)+binomial(2,1)==binomial(3,2)`

That's some light, not some hard proof I would give to a math teacher but this is why I put a test where I check the count and the size hint at every step of it.
**PS:** I hope I did not make some mistake in this big comment...

src/combinations.rs
Philippe-Cholet: Well that's a possibility: `fn remaining_for(n: usize, k: usize, first: bool, indices: &[usize]) -> Option<usize>`
I don't think it would be elegant but we could discard `fill` I guess. I don't think it's a big problem to fill the buffer because if we did generate a few combinations (`n-k+1`, really small compared to the number of combinations in most cases), it would be filled anyway. But it's possible.

src/combinations.rs
Philippe-Cholet: Sure. Something like this below inside `combinations.rs`?

    #[test]
    fn test_checked_binomial() {
        let table: [[u128; N]; K] = [[1, ...], ...]; // N K constants.
        for n in 0..N {
            for k in 0..K {
                assert_eq!(checked_binomial(n, k), usize::try_from(lookuptable_holding_u128[n][k]).ok());
            }
        }
    }

src/combinations.rs
Philippe-Cholet: Done in https://github.com/rust-itertools/itertools/pull/729/commits/2ab0ff199293b68f1e8ff38db373c439dac24222

src/combinations.rs
```diff
@@ -120,9 +120,51 @@ impl<I> Iterator for Combinations<I>
...
+    for i in 1..=k {
+        c = (c / i).checked_mul(n)?.checked_add((c % i).checked_mul(n)? / i)?;
+        n -= 1;
+    }
+    Some(c)
+}
+
+/// For a given size `n`, return the count of remaining combinations or None if it would overflow.
+fn remaining_for(n: usize, first: bool, indices: &[usize]) -> Option<usize> {
+    let k = indices.len();
```
phimuemue, line 159: This only works if `n>=k`, right? Please add an `assert(n>=k)`.

src/combinations.rs
```diff
@@ -120,9 +120,51 @@ impl<I> Iterator for Combinations<I>
...
+    if first {
+        checked_binomial(n, k)
+    } else {
+        indices
+            .iter()
+            .enumerate()
+            .fold(Some(0), |sum, (k0, n0)| {
+                sum.and_then(|s| s.checked_add(checked_binomial(n - 1 - *n0, k - k0)?))
+            })
```
phimuemue, line 168: Thanks for your explanations.

Please add these links in the comments for the next reader:
* https://en.wikipedia.org/wiki/Combinatorial_number_system
* http://www.site.uottawa.ca/~lucia/courses/5165-09/GenCombObj.pdf

And maybe the following explanation within "The combinations generated after the current one can be counted by counting as follows:
* The subsequent combinations that differ in indices[0]: If subsequent combinations differ in indices[0], then their value for indices[0] must be at least 1 greater than the current indices[0]. As indices is strictly monotonically sorted, this means we can effectively choose k values from (n - 1 - indices[0]), leading to binomial(n-1-indices[0], k) possibilities.
* The subsequent combinations with same indices[0], but differing indices[1]: Here we can choose k-1 values from (n-1-indices[1]) values, leading to binomial(n-1-indices[1], k-1) possibilities.
* (...)
* The subsequent combinations with same indices[0..=i], but differing indices[i]: Here we can choose k-i values from (n-1-indices[i]) values: binomial(n-1-indices[i], k-i).
Since subsequent combinations can in any index, we must sum up the aforementioned binomial coefficients."

Please rename `k0` to `i` so that it matches the comment, and comment that `n0` resembles `indices[i]`.

tests/test_std.rs
```diff
@@ -909,6 +909,21 @@ fn combinations_zero() {
     it::assert_equal((0..0).combinations(0), vec![vec![]]);
 }
 
+#[test]
+fn combinations_range_count() {
+    for n in 0..6 {
+        for k in 0..=n {
+            let len = (n - k + 1..=n).product::<usize>() / (1..=k).product::<usize>();
```
phimuemue, line 916: Please `assert_eq(len, it.count)` and `assert_eq(len, it.size_hint().0)` and `assert_eq(Some(len), it.size_hint().1)`.

tests/test_std.rs
```diff
@@ -909,6 +909,21 @@ fn combinations_zero() {
     it::assert_equal((0..0).combinations(0), vec![vec![]]);
 }
 
+#[test]
+fn combinations_range_count() {
+    for n in 0..6 {
+        for k in 0..=n {
```
phimuemue, line 915: Can we loop until 10?

tests/test_std.rs
```diff
@@ -909,6 +909,21 @@ fn combinations_zero() {
...
+            let len = (n - k + 1..=n).product::<usize>() / (1..=k).product::<usize>();
+            let mut it = (0..n).combinations(k);
+            for count in (0..=len).rev() {
+                assert_eq!(it.size_hint(), (count, Some(count)));
+                assert_eq!(it.clone().count(), count);
+                assert_eq!(it.next().is_none(), count == 0);
```
phimuemue, line 921: Please do not call `it.next()` (i.e. side-effects) in the `assert`. Introduce an auxiliary variable.

tests/test_std.rs
```diff
@@ -909,6 +909,21 @@ fn combinations_zero() {
...
+            }
+        }
```
phimuemue, line 923: Please add `let should_be_none = it.next(); assert(should_be_none.is_none())` after the loops.

src/combinations.rs
```diff
@@ -120,9 +120,51 @@ impl<I> Iterator for Combinations<I>
...
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let (mut low, mut upp) = self.pool.size_hint();
+        low = remaining_for(low, self.first, &self.indices).unwrap_or(usize::MAX);
+        upp = upp.and_then(|upp| remaining_for(upp, self.first, &self.indices));
+        (low, upp)
+    }
+
+    fn count(self) -> usize {
+        let Self { indices, pool, first } = self;
+        let n = pool.len() + pool.it.count();
```
phimuemue, line 133: Please add a comment: `// TODO make pool.it private`

phimuemue (changes requested):
Thanks for the update. We should comment the sources for the complex algorithms and add some tests, then I think we're good to go.

src/combinations.rs
Philippe-Cholet: Runs fast for a limit of 100 (0.01 second I think, 65% of the last row overflow). I made it run on 10000 to see though (20-30 seconds, and like 98% overflow).
Maybe you have another meaning of exhaustive testing than me but for small k, it won't overflow for a long long time (`6_074_001_001` would be the smaller `n` for which `(0..n).combinations(2).count()==n*(n-1)/2` would overflow for a fixed `k=2`). So we have to have a limit on `n`. 100 might seem small but already 65% of the row has already overflowed.

src/combinations.rs
```diff
@@ -120,9 +120,51 @@ impl<I> Iterator for Combinations<I>
...
+        remaining_for(n, first, &indices).expect("Iterator count greater than usize::MAX")
```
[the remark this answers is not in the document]
Philippe-Cholet: I used this because permutations used it (same message). But ok.

src/combinations.rs
Philippe-Cholet: Good catch but the algorithm currently accepts `n` to be smaller than `k` so I decided to return `Some(0)` when `n < k`.

src/combinations.rs
Philippe-Cholet: Oh my! I missed that wikipedia link. Done!

Philippe-Cholet:
All changes are done (with a limit of 500 for `test_checked_binomial`).
And I added a TODO about "sum of options" (possible for a MSRV of 1.37, currently 1.36) that would really make `remaining_for` more readable.

PR #730, opened 2023-08-16
Simplify kmerge_by.

ttsugriy:
`Vec`'s collect uses size hint under the hood, but it's more concise and idiomatic.


Philippe-Cholet:
Except `FilterMap::size_hint` has a lower bound of 0 (see [this](https://doc.rust-lang.org/src/core/iter/adapters/filter_map.rs.html#125)).
We have insight about the predicate that `FilterMap` simply can not have.

ttsugriy:
ah, I totally missed this, @Philippe-Cholet You're absolutely right.

PR #736, opened 2023-08-24
Merge `Merge/MergeBy/MergeJoinBy` implementations (v2)

### LINKED ISSUE #701: Merge `MergeBy` and `MergeJoinBy`

Philippe-Cholet:
Cleaner alternative to the experimental/messy #711. Closes #701.
@phimuemue I would additionally rename `InternalMergeJoinBy` to `MergeBy`, making `MergeJoinBy` an alias of `MergeBy` instead. But I would like your opinion on this first.
I recently found a way to not add a fourth type to `MergeJoinBy` (added to the old PR) but another difference is that I do not wrap with `MergeFuncT` anymore.

### DIFF
--- a/src/merge_join.rs
+++ b/src/merge_join.rs
@@ -109,7 +109,6 @@ impl<L, R, T, F: FnMut(&L, &R) -> T> FuncLR<L, R> for F {
 }
 
 pub trait OrderingOrBool<L, R> {
-    type Out;
     type MergeResult;
     fn left(left: L) -> Self::MergeResult;
     fn right(right: R) -> Self::MergeResult;
@@ -121,7 +120,6 @@ pub trait OrderingOrBool<L, R> {
 }
 
 impl<L, R, F: FnMut(&L, &R) -> Ordering> OrderingOrBool<L, R> for MergeFuncLR<F, Ordering> {
-    type Out = Ordering;
     type MergeResult = EitherOrBoth<L, R>;
     fn left(left: L) -> Self::MergeResult {
         EitherOrBoth::Left(left)
@@ -149,7 +147,6 @@ impl<L, R, F: FnMut(&L, &R) -> Ordering> OrderingOrBool<L, R> for MergeFuncLR<F,
 }
 
 impl<L, R, F: FnMut(&L, &R) -> bool> OrderingOrBool<L, R> for MergeFuncLR<F, bool> {
-    type Out = bool;
     type MergeResult = Either<L, R>;
     fn left(left: L) -> Self::MergeResult {
         Either::Left(left)
@@ -171,7 +168,6 @@ impl<L, R, F: FnMut(&L, &R) -> bool> OrderingOrBool<L, R> for MergeFuncLR<F, boo
 }
 
 impl<T, F: FnMut(&T, &T) -> bool> OrderingOrBool<T, T> for F {
-    type Out = bool;
     type MergeResult = T;
     fn left(left: T) -> Self::MergeResult {
         left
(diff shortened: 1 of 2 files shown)


src/merge_join.rs
```diff
@@ -57,97 +47,20 @@ pub fn merge<I, J>(i: I, j: J) -> Merge<<I as IntoIterator>::IntoIter, <J as Int
...
-{
-    debug_fmt_fields!(MergeBy, a, b);
-}
-
-impl<T, F: FnMut(&T, &T)->bool> MergePredicate<T> for F {
-    fn merge_pred(&mut self, a: &T, b: &T) -> bool {
-        self(a, b)
-    }
-}
+pub type MergeBy<I, J, F> = InternalMergeJoinBy<I, J, F>;
```
jswrenn, line 50: Yes, I think you should rename `InternalMergeJoinBy` to `MergeBy` and ditch this type alias.

jswrenn (changes requested):

Philippe-Cholet:
@phimuemue What do you think of this new version?
`MergeJoinBy` do not have a fourth type here, as you thought possible (the small `FuncLR` trait that make it work seems reasonable to me).

Philippe-Cholet:
Rebased after recent #751.

src/merge_join.rs
```diff
@@ -171,46 +83,54 @@ where
...
+pub trait FuncLR<L, R> {
+    type T;
+}
+
+impl<L, R, T, F: FnMut(&L, &R) -> T> FuncLR<L, R> for F {
+    type T = T;
 }
 
 pub trait OrderingOrBool<L, R> {
+    type Out;
```
jswrenn, line 112:
```suggestion
```
I believe this associated type can be removed without issue. At least, I just tried removing it, and `cargo check` succeeded.

jswrenn (changes requested):

src/merge_join.rs
Philippe-Cholet: Well, it is a shock to me but you are right.

jswrenn (approved):

Philippe-Cholet:
That feature was quite a ride for me. 🎉

@jswrenn Note that you are mixing different PRs with the "next" milestone.

Philippe-Cholet:
Just that old PRs of 0.11 are with the new ones, which I noticed and found weird but if it's not an issue then case closed.

PR #737, opened 2023-08-26
Count combinations with replacement

Philippe-Cholet:
@phimuemue
Maybe a comment in `remaining_for` to suggest to see the big comment in `combinations.rs`.
I have a test `combinations_with_replacement_inexact_size_hints` if you want.

### DIFF
--- a/src/combinations_with_replacement.rs
+++ b/src/combinations_with_replacement.rs
@@ -101,6 +101,13 @@ where
             None => None,
         }
     }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let (mut low, mut upp) = self.pool.size_hint();
+        low = remaining_for(low, self.first, &self.indices).unwrap_or(usize::MAX);
+        upp = upp.and_then(|upp| remaining_for(upp, self.first, &self.indices));
+        (low, upp)
+    }
 }
 
 impl<I> FusedIterator for CombinationsWithReplacement<I>
--- a/src/combinations_with_replacement.rs
+++ b/src/combinations_with_replacement.rs
@@ -108,6 +108,12 @@ where
         upp = upp.and_then(|upp| remaining_for(upp, self.first, &self.indices));
         (low, upp)
     }
+
+    fn count(self) -> usize {
+        let Self { indices, pool, first } = self;
+        let n = pool.count();
+        remaining_for(n, first, &indices).unwrap()
+    }
 }
 
 impl<I> FusedIterator for CombinationsWithReplacement<I>
--- a/src/combinations_with_replacement.rs
+++ b/src/combinations_with_replacement.rs
@@ -124,6 +124,10 @@ where
 
 /// For a given size `n`, return the count of remaining combinations with replacement or None if it would overflow.
 fn remaining_for(n: usize, first: bool, indices: &[usize]) -> Option<usize> {
+    // With a "stars and bars" representation, choose k values with replacement from n values is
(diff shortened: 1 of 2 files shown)


Philippe-Cholet:
@phimuemue What do you think?

src/combinations_with_replacement.rs
```diff
@@ -107,3 +108,22 @@ where
...
+        indices
+            .iter()
+            .enumerate()
+            // TODO: Once the MSRV hits 1.37.0, we can sum options instead:
+            // .map(|(i, n0)| count(n - 1 - *n0, k - i))
+            // .sum()
+            .fold(Some(0), |sum, (i, n0)| {
+                sum.and_then(|s| s.checked_add(count(n - 1 - *n0, k - i)?))
+            })
+    }
```
phimuemue, line 141: I *believe* you that this is correct, but I have a hard time comprehending this without comments. Is it the "stars and sticks" thing that leads to `binomial(n+k-1, k)`?

Could you comment the loop similar to what we did in the other cases?

In addition: Could `n+k` overflow? And if so, shouldn't we return `None` here instead of `panic`king?

src/combinations_with_replacement.rs
```diff
@@ -101,6 +101,13 @@ where
             None => None,
         }
     }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let (mut low, mut upp) = self.pool.size_hint();
+        low = remaining_for(low, self.first, &self.indices).unwrap_or(usize::MAX);
+        upp = upp.and_then(|upp| remaining_for(upp, self.first, &self.indices));
+        (low, upp)
+    }
```
phimuemue, line 110: Nicely in line with `Combinations::size_hint`.

src/combinations_with_replacement.rs
```diff
@@ -108,6 +108,12 @@ where
         upp = upp.and_then(|upp| remaining_for(upp, self.first, &self.indices));
         (low, upp)
     }
+
+    fn count(self) -> usize {
+        let Self { indices, pool, first } = self;
+        let n = pool.count();
+        remaining_for(n, first, &indices).unwrap()
+    }
```
phimuemue, line 116: Nicely in line with `Combinations::count`.

src/combinations_with_replacement.rs
```diff
@@ -25,7 +25,7 @@ where
     I: Iterator + fmt::Debug,
     I::Item: fmt::Debug + Clone,
 {
-    debug_fmt_fields!(Combinations, indices, pool, first);
+    debug_fmt_fields!(CombinationsWithReplacement, indices, pool, first);
```
phimuemue, line 28: Good catch, thanks!

phimuemue (changes requested):
Hi @Philippe-Cholet, thanks again!

Please comment how `remaining_for` works - I *believe* that you did your homework there, but I'd really appreciate if you could clear things up for the reader.

src/combinations_with_replacement.rs
Philippe-Cholet: I chose the expression "stars and bars" from wikipedia, expression I did not know. I added a comment.
Good catch `n+k` overflowing, handled.
And I copied/updated the previous comment about the loop.

phimuemue:
bors r+

Thanks @Philippe-Cholet.

One thing: I'm constantly annoyed by the `first` field. Do you happen to know if there's a way to avoid this? (No need to work on it, I was just interested if you have insights into that.)

PR #738, opened 2023-08-26
Update rust/itertools versions in the doc

Philippe-Cholet:
Rust 1.36 according to `Cargo.toml`.
And the fixed `html_root_url` updates all html links we might encounter. I personally do not use it very much but being redirect of `0.8` felt weird.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -42,8 +42,8 @@
 //!
 //! ## Rust Version
 //!
-//! This version of itertools requires Rust 1.32 or later.
-#![doc(html_root_url="https://docs.rs/itertools/0.8/")]
+//! This version of itertools requires Rust 1.36 or later.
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -42,8 +42,8 @@
 //!
 //! ## Rust Version
 //!
-//! This version of itertools requires Rust 1.32 or later.
-#![doc(html_root_url="https://docs.rs/itertools/0.8/")]
+//! This version of itertools requires Rust 1.36 or later.
+#![doc(html_root_url="https://docs.rs/itertools/0.11/")]
```
jswrenn, line 46: I wonder why we have this line at all. I've never written this in any of my other crates. Perhaps it's just a vestige of the past that can be removed without issue?

PR #739, opened 2023-08-27
Complete `Permutations::size_hint`

Philippe-Cholet:
@phimuemue
This series of size hint/count improvements ends where it started: with a TODO I wanted to fix.
The end user will mostly enjoy that `(0..n).permutations(k).map(...).collect_vec()` will now allocate to the resulting vector in _one go_ (`PermutationState::StartUnknownLen` case).

In the first commit, you probably will want me to unwrap but `panic!(message)` is more similar to `expect` than `unwrap`. Which is why I previously used `expect` too.
The 2nd commit (about `enough_vals`) is off topic but I really don't see why it would not be a (minor) improvement.
I have a test `permutations_inexact_size_hints` ready if you want (similar to `combinations_inexact_size_hints`).

### DIFF
--- a/src/permutations.rs
+++ b/src/permutations.rs
@@ -138,12 +138,14 @@ where
     fn size_hint(&self) -> (usize, Option<usize>) {
         match self.state {
             PermutationState::StartUnknownLen { k } => {
+                // At the beginning, there are `n!/(n-k)!` items to come (see `remaining`) but `n` might be unknown.
                 let (mut low, mut upp) = self.vals.size_hint();
                 low = CompleteState::Start { n: low, k }.remaining().unwrap_or(usize::MAX);
                 upp = upp.and_then(|n| CompleteState::Start { n, k }.remaining());
                 (low, upp)
             }
             PermutationState::OngoingUnknownLen { k, min_n } => {
+                // Same as `StartUnknownLen` minus the `prev_iteration_count` previously generated items.
                 let prev_iteration_count = min_n - k + 1;
                 let (mut low, mut upp) = self.vals.size_hint();
                 low = CompleteState::Start { n: low, k }
--- a/src/permutations.rs
+++ b/src/permutations.rs
@@ -137,7 +137,12 @@ where
 
     fn size_hint(&self) -> (usize, Option<usize>) {
         match self.state {
-            PermutationState::StartUnknownLen { .. } |
+            PermutationState::StartUnknownLen { k } => {
+                let (mut low, mut upp) = self.vals.size_hint();
+                low = CompleteState::Start { n: low, k }.remaining().unwrap_or(usize::MAX);
+                upp = upp.and_then(|n| CompleteState::Start { n, k }.remaining());
+                (low, upp)
+            }
             PermutationState::OngoingUnknownLen { .. } => (0, None), // TODO can we improve this lower bound?
             PermutationState::Complete(ref state) => match state.remaining() {
                 Some(count) => (count, Some(count)),
--- a/src/permutations.rs
+++ b/src/permutations.rs
@@ -137,21 +137,18 @@ where
     }
 
     fn size_hint(&self) -> SizeHint {
+        let at_start = |k| {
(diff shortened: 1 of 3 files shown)


tests/test_std.rs
```diff
@@ -986,6 +986,32 @@ fn permutations_zero() {
...
+                assert!(elem.is_some());
+                assert_eq!(count, it.clone().count());
+                assert_eq!(count, it.size_hint().0);
+                assert_eq!(Some(count), it.size_hint().1);
+            }
+            let should_be_none = it.next();
+            assert!(should_be_none.is_none());
+        }
+    }
+}
```
jswrenn, line 1013: Is it possible to test an overflowing case by taking permutations from an `usize::MAX` stream of `()`? If such a test would take prohibitively long to run, we don't need it.

src/permutations.rs
```diff
@@ -153,11 +137,29 @@ where
...
+                (low, upp)
+            }
             PermutationState::Complete(ref state) => match state.remaining() {
-                CompleteStateRemaining::Known(count) => (count, Some(count)),
-                CompleteStateRemaining::Overflow => (::std::usize::MAX, None)
+                Some(count) => (count, Some(count)),
+                None => (::std::usize::MAX, None)
             }
             PermutationState::Empty => (0, Some(0))
         }
```
jswrenn, line 165: Your individual commit messages have helpful commentary about each of the branches here. Could you include some of those explanations as comments?

jswrenn (changes requested):
Thank you for this!

src/permutations.rs
Philippe-Cholet: Done in https://github.com/rust-itertools/itertools/pull/739/commits/95536d6413cb340a7c0c90ca5b3925bfa198ba24

tests/test_std.rs
Philippe-Cholet: I'm not sure to understand "_an `usize::MAX` stream of `()`_". Could you be more specific (a small code snippet maybe) about what I should test (assuming it's fast enough)?

src/permutations.rs
```diff
@@ -47,11 +47,6 @@ enum CompleteState {
     }
 }
 
-enum CompleteStateRemaining {
-    Known(usize),
-    Overflow,
-}
-
```
phimuemue, line 54: Good idea.

src/permutations.rs
```diff
@@ -143,7 +143,20 @@ where
...
+                let prev_iteration_count = min_n - k + 1;
+                let (mut low, mut upp) = self.vals.size_hint();
+                low = CompleteState::Start { n: low, k }
+                    .remaining()
+                    .unwrap_or(usize::MAX)
+                    .saturating_sub(prev_iteration_count);
+                upp = upp.and_then(|n| {
+                    CompleteState::Start { n, k }
+                        .remaining()
+                        .map(|count| count.saturating_sub(prev_iteration_count))
```
phimuemue, line 158: Can we express this via `size_hint::sub_scalar` and share code for `OngoingUnknownLen and `StartUnknownLen`?

tests/test_std.rs
phimuemue: Maybe `std::iter::repeat(()).combinations(2)`.

phimuemue (changes requested):
Thanks @Philippe-Cholet.

tests/test_std.rs
phimuemue: Maybe `std::iter::repeat(()).combinations(2)`.

src/permutations.rs
Philippe-Cholet: I forgot about `size_hint` module again, that's what's needed yes! Shared code done too.

tests/test_std.rs
Philippe-Cholet: Done.

jswrenn: Yes, this is great! Thank you!

jswrenn (approved):

PR #745, opened 2023-09-06
Do not sum options

Philippe-Cholet:
A mistake on my part, discovered in time as code is still commented out. Note that it passed tests.
It only returns None if the iterator yields None. But it would overflow whenever an addition does.

Do not product options either, as multiplications could overflow the same way. That's how I found out.

I guess `checked_sum` and `checked_product` methods would be nice.
EDIT: ~Would you be interested?~ Related: https://github.com/rust-lang/rust/pull/95485


phimuemue:
Nicely spotted @Philippe-Cholet.

Possibly indicates that we should really include the overflowing parts in tests.

bors r+

PR #752, opened 2023-09-18
Implement ExactSizeIterator for (Circular)TupleWindows

### LINKED ISSUE #660: Implement ExactSizeIterator for CircularTupleWindows
`.circular_tuple_windows()` will produce the same amount of elements as the upstream iterator, so I believe the implementation should be pretty simple.

Speaking about the practical example, I want to convert a vector of 2D points to pairs of normalized vectors.

```rust
  let vecs = shape
    .into_iter()
    .circular_tuple_windows::<(_, _)>()
    .map(|((x1, y1), (x2, y2)): (Point, Point)| (x2 - x1, y2 - y1) as Point)
    .circular_tuple_windows(); //  trait `ExactSizeIterator` is not satisfied
```

Philippe-Cholet:
Closes #660 
Implement `ExactSizeIterator` ~at the cost of storing/updating the length~ without any additional cost. I just flatten `struct Take<I> { iter: I, n: usize }` from `iter: Take<TupleWindows<Cycle<I>, T>>` into `iter: TupleWindows<Cycle<I>, T>, len: usize`.
It has the same length as the input iterator.

### DIFF
--- a/src/tuple_impl.rs
+++ b/src/tuple_impl.rs
@@ -184,6 +184,12 @@ where
         }
         None
     }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        // At definition, `T::num_items() - 1` are collected
+        // so each remaining item in `iter` will lead to an item.
+        self.iter.size_hint()
+    }
 }
 
 impl<I, T> FusedIterator for TupleWindows<I, T>
--- a/src/tuple_impl.rs
+++ b/src/tuple_impl.rs
@@ -3,7 +3,6 @@
 use std::iter::Cycle;
 use std::iter::Fuse;
 use std::iter::FusedIterator;
-use std::iter::Take;
 use std::marker::PhantomData;
 
 // `HomogeneousTuple` is a public facade for `TupleCollect`, allowing
@@ -208,7 +207,7 @@ where
     I: Iterator<Item = T::Item> + Clone,
     T: TupleCollect + Clone,
 {
-    iter: Take<TupleWindows<Cycle<I>, T>>,
+    iter: TupleWindows<Cycle<I>, T>,
     len: usize,
     phantom_data: PhantomData<T>,
 }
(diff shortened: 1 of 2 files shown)


src/tuple_impl.rs
```diff
@@ -236,8 +238,29 @@ where
     type Item = T;
 
     fn next(&mut self) -> Option<Self::Item> {
+        self.len = self.len.saturating_sub(1);
         self.iter.next()
     }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        (self.len, Some(self.len))
```
jswrenn, line 246: Rather than add an additional field, why not just forward the underlying implementation of `size_hint`. 
```suggestion
        self.iter.size_hint()
```

jswrenn (changes requested):

src/tuple_impl.rs
Philippe-Cholet: The problem is the type of the `iter` field: `Take<TupleWindows<Cycle<I>, T>>`.
Even if `Take::size_hint` does a good job, `TupleWindows` does not even have access to the original iterator as it's wrapped by `Cycle` and `TupleWindows` won't store the length at definition as it does not require `I` to be `ExactSizeIterator`.
I guess we could rewrite `CircularTupleWindows` entirely to not rely on such wrapped type, but I'm not very keen on that to save a field.

Philippe-Cholet: Ahh.. maybe we can ensure that `TupleWindows<Cycle<I>, T>` has an overflowing size hint then `Take` would successfully do the job.

Philippe-Cholet: As I feared, it's not possible in some cases. `TupleWindows` would have fewer items.

    fn size_hint(&self) -> (usize, Option<usize>) { // TupleWindows
        size_hint::sub_scalar(self.iter.size_hint(), T::num_items() - 1)
    }
    // Might still be valuable to commit thought.

The test below would fail

    [(); usize::MAX].iter().circular_tuple_windows::<(_, _, _, _)>().len() == usize::MAX

However it's not a problem, I'm gonna unwrap `Take` type as it's just `struct Take<I> { iter: I, n: usize }`.

jswrenn (approved):

PR #753, opened 2023-09-19
Fix ignored doctests

Philippe-Cholet:
[Doctest attribute](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#attributes) | What it means
----------------- | -------------
text         | Not code.
ignore       | It is ignored by default but might run, e.g. `cargo test -- --ignored`.
should_panic | It must compile but panic at runtime.
no_run       | It must compile but is not run.
compile_fail | It must not compile.
edition2018  | Only runs on the given edition.

Personally, I annotate with "ignore" when it's slow and I don't want to run it every time.
Otherwise, I think another attribute should be used.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2373,7 +2373,9 @@ pub trait Itertools: Iterator {
     /// For example the sequence *Ok(1), Ok(2), Ok(3)* will result in a
     /// computation like this:
     ///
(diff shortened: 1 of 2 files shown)


jswrenn (approved):
Thanks!

PR #754, opened 2023-09-19
use 'cargo-hack' to run MSRV check without dev dependencies

danieleades:
decouple the MSRV check from the dev dependencies using `cargo-hack`

this allows dev dependencies to be updated without impacting downstream users.

note that the MSRV used in this PR was found using the following command-

`cargo msrv -- cargo no-dev-deps check`

(which requires installing `cargo-msrv` and `cargo-no-dev-deps`)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,11 +9,9 @@ on:
 
 jobs:
   check:
-    name: check
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        build: [msrv, stable]
         features:
           [
             "",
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -9,36 +9,29 @@ on:
...
-      - uses: dtolnay/rust-toolchain@master
-        with:
-          toolchain: ${{ matrix.rust }}
+      - uses: dtolnay/rust-toolchain@stable
       - run: cargo check ${{ matrix.features }}
 
+  msrv:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: taiki-e/install-action@cargo-hack
```
jswrenn, line 30: Thanks! I didn't know `cargo hack` could do this! 

Just add the `actions/checkout` dependency before this, and I think CI will pass.

jswrenn (changes requested):

.github/workflows/ci.yml
```diff
@@ -9,36 +9,32 @@ on:
...
-      - uses: dtolnay/rust-toolchain@master
-        with:
-          toolchain: ${{ matrix.rust }}
+      - uses: dtolnay/rust-toolchain@stable
       - run: cargo check ${{ matrix.features }}
 
+  msrv:
+    runs-on: ubuntu-latest
+    env:
+      CARGO_NET_GIT_FETCH_WITH_CLI: true
```
danieleades, line 30: the 1.43.1 version of cargo does not appear to work in CI without this line. This should be reviewed on the next MSRV bump

Cargo.toml
```diff
@@ -16,7 +16,7 @@ exclude = ["/bors.toml"]
 
 edition = "2018"
 
-rust-version = "1.36.0"
+rust-version = "1.43.1"
```
jswrenn, line 19: Can you mention that we're bumping MSRV in the commit message? I just want to make sure I don't miss it when I'm updating our CHANGELOG.

jswrenn (changes requested):

Cargo.toml
danieleades: yeah no worries. How do you want me to word it? Given that this commit doesn't actually change the MSRV, it just documents it correctly (ie. it would not have compiled with the 1.36.0 toolchain even before this change)

Philippe-Cholet: There is also the module documentation of "lib.rs" that mention the MSRV to update.

danieleades: done. Updated the docs in a new commit. Do you want to do a squash merge and choose your own commit message?

jswrenn: Heh, I've apparently turned off squash merging. I'll need to fiddle with that.

In the meantime, could you set the commit message to "Document and test bumped MSRV."

Philippe-Cholet:
I don't know much about CI configuration yet so I can only approve the documentation change about MSRV.

Philippe-Cholet (approved):

jswrenn (approved):

PR #740, opened 2023-08-27
clippy

mightyiam:
Co-authored-by: warren2k <846021+warren2k@users.noreply.github.com>

PR #759, opened 2023-09-21
Address clippy lints

danieleades:


benches/bench1.rs
```diff
@@ -474,10 +474,8 @@ fn equal(c: &mut Criterion) {
 fn merge_default(c: &mut Criterion) {
     let mut data1 = vec![0; 1024];
     let mut data2 = vec![0; 800];
-    let mut x = 0;
-    for (_, elt) in data1.iter_mut().enumerate() {
```
danieleades, line 478: is there a reason this pattern was used in the bench tests?

Philippe-Cholet: I don't know about this pattern. It definitely seems weird. Maybe due to something removed from itertools, I have no clue.
But you seem to address the same problems as the recent #740 . Sometimes in different ways.

Philippe-Cholet:
#740 which addresses clippy lints is now merged so this is sadly duplicated work. Thanks anyway.

PR #760, opened 2023-09-23
Minimize features

Philippe-Cholet:
Somehow, I looked at `try_collect` and ticked seeing that it needs `use_alloc`. It sure does in most common cases but we can collect to `arrayvec` for example.
Because of that, I looked at each adaptor/method to find another unrequired features and found only the ones based on `extrema_set` which only needs to allocate to vectors.

### DIFF
--- a/src/extrema_set.rs
+++ b/src/extrema_set.rs
@@ -1,3 +1,5 @@
+#![cfg(feature = "use_alloc")]
+use alloc::{vec, vec::Vec};
 use std::cmp::Ordering;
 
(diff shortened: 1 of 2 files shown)


Philippe-Cholet:
@jswrenn This seems straightforward to me. What do you think?
I guess I'm reserved about expanding `try_collect` usage (to "no std") because of the nightly version in the std.
But I don't see any problem about the "`extrema_set`" methods.

jswrenn (approved):
This looks good to me!

PR #761, opened 2023-09-27
Implement `ExactSizeIterator` for `Tuples`

Philippe-Cholet:
I was hunting missing `size_hint` specializations for a few weeks now and from my notes, `Tuples` is apparently the last one.
(Apart from `TupleCombinations` but I'm quite unsure it's reasonably possible.)

### DIFF
--- a/src/tuple_impl.rs
+++ b/src/tuple_impl.rs
@@ -105,6 +105,14 @@ where
     fn next(&mut self) -> Option<Self::Item> {
         T::collect_from_iter(&mut self.iter, &mut self.buf)
     }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let buf_len = T::buffer_len(&self.buf);
+        let (mut low, mut hi) = self.iter.size_hint();
+        low = add_then_div(low, buf_len, T::num_items()).unwrap_or(usize::MAX);
+        hi = hi.and_then(|elt| add_then_div(elt, buf_len, T::num_items()));
+        (low, hi)
+    }
 }
 
 /// `(n + a) / d` avoiding overflow when possible, returns `None` if it overflows.
--- a/src/tuple_impl.rs
+++ b/src/tuple_impl.rs
@@ -107,6 +107,12 @@ where
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/tuple_impl.rs
```diff
@@ -105,6 +105,27 @@ where
...
         T::collect_from_iter(&mut self.iter, &mut self.buf)
     }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let buf_len = T::buffer_len(&self.buf);
+        let (mut low, mut hi) = self.iter.size_hint();
+        low = add_then_div(low, buf_len, T::num_items()).unwrap_or(usize::MAX);
+        hi = hi.and_then(|elt| add_then_div(elt, buf_len, T::num_items()));
+        (low, hi)
+    }
```
jswrenn, line 115: This seems fine, let's just add some comments for future passerby: 
```suggestion
    fn size_hint(&self) -> (usize, Option<usize>) {
        // The number of elts we've drawn from the underlying iterator, but have
        // not yet produced as a tuple.
        let buffered = T::buffer_len(&self.buf);
        // To that, we must add the size estimates of the underlying iterator.
        let (mut unbuffered_lo, mut unbuffered_hi) = self.iter.size_hint();
        // The total low estimate is the sum of the already-buffered elements,
        // plus the low estimate of remaining unbuffered elements, divided by
        // the tuple size.
        let total_lo = add_then_div(unbuffered_lo, buffered, T::num_items()).unwrap_or(usize::MAX);
        // And likewise for the total high estimate, but using the high estimate
        // of the remaining unbuffered elements.
        let total_hi = unbuffered_hi.and_then(|hi| add_then_div(hi, buffered, T::num_items()));
        (total_lo, total_hi)
    }
```

jswrenn (changes requested):

src/tuple_impl.rs
jswrenn: (I haven't actually tested the above snippet, so apologies if there's a silly error in the renaming.

Philippe-Cholet: I thought you did. It hopefully passes CI as is.

Issue #755, opened 2023-09-19
Audit for missing `for_each` implementations. 


jswrenn:
Implementing `for_each` for our iterators allows us to provide more performant internal iteration. We provide it in some cases, but not all. I'd like us to take three steps towards resolving this:
1. Take an inventory of whether our adapters provide `for_each`.
2. Implement `for_each` for missing adapters. 
3. (Potentially!) Implement a clippy lint to prevent regressions.

Happy to provide mentorship for any of these items.

scottmcm:
Note that the vast majority of the time, you should be overriding `fold`, not `for_each`.

The default `for_each` just calls `fold` with a trivial `()` accumulator, which is easily optimized out (since in the default Rust calling convention, it's not even passed).

Thus the only reason to override `for_each` *in addition* to `fold` is if you can optimize something by knowing that the accumulation is stateless, but I can't off-hand recall any case where I've seen that to be the case.

jswrenn:
Whoops, yes, of course. I'll modify the issue accordingly.

Philippe-Cholet:
First, find `fold` implementations is easy.

file | Adaptors with specialized `fold`
--- | -----------------------------
adaptors/coalesce.rs | CoalesceBy
adaptors/map.rs | MapSpecialCase
adaptors/mod.rs | PutBack, Product, FilterOk, FilterMapOk, Update
cons_tuples_impl.rs | ConsTuples
intersperse.rs | IntersperseWith
process_results_impl.rs | ProcessResults

Other adaptors and other iterators do not have `fold` specializations.

phimuemue:
Once we do this, we should test all of them via `test_specializations`.

Owen-CH-Leung:
I'm interested to contribute to this crate and I digged a bit into the source code. I think one potential candidate to override `fold` is `Combinations`. What do you guys think ?

I can create a PR in coming few days to implement this.

phimuemue:
> I'm interested to contribute to this crate and I digged a bit into the source code. I think one potential candidate to override `fold` is `Combinations`.
> 
> https://github.com/rust-itertools/itertools/blob/master/src/combinations.rs#L100
> 
> What do you guys think ?
> 
> I can create a PR in coming few days to implement this.

@Philippe-Cholet I think you're the most knowledgeable there. Could you mentor/review this?

Philippe-Cholet:
[...]

PR #762, opened 2023-09-27
Implement custom fold for ZipSlice

Owen-CH-Leung:
As per discussion in #755 , this PR implements custom `fold` logic for `ZipSlices<T, U>` and create benchmark test to measure the performance gain compared with the default `fold`. 

Running the `zipdot i32 zipslices fold` benchmark using the default `fold` logic, the time was `time:   [42.749 ns 43.163 ns 43.782 ns]`

Running the `zipdot i32 zipslices fold` benchmark using the custom `fold` logic, the time was `time:   [42.523 ns 42.621 ns 42.737 ns]`

### DIFF
--- a/benches/bench1.rs
+++ b/benches/bench1.rs
@@ -169,6 +169,20 @@ fn zipdot_i32_zipslices(c: &mut Criterion) {
     });
 }
 
+fn zipdot_i32_zipslices_fold(c: &mut Criterion) {
+    let xs = vec![2; 1024];
+    let ys = vec![2; 768];
+    let xs = black_box(xs);
+    let ys = black_box(ys);
+    // output is 3072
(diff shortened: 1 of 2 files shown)


Philippe-Cholet:
I was gonna say I was surprised by the `unsafe` use, especially for such time win but ZipSlices is not part of the itertools library, it's only part of benches.

jswrenn (commented):
I appreciate the help on #755. However, as @Philippe-Cholet suggests, our focus is on improving the performance of user-facing iterators. Specializing this particular `fold` adds maintenance cost without any performance gain (even internally).

Owen-CH-Leung:
Ahh yesss I misread. I thought it's under `src` when I was revising it. Thanks both for your comments. I'll continue to see if custom `fold` can be implemented in other iterators

PR #763, opened 2023-09-27
`TupleCombinations`: `size_hint` and `count`

Philippe-Cholet:
This was definitely simpler to implement than it was for `Combinations` (with vector items). The only difficulty is that it's recursive with a macro.

According to my notes, all iterators that should have a `size_hint` specialization now have. I like that.

The `count_ident` macro is moved to be available library-wide.

### DIFF
--- a/src/adaptors/mod.rs
+++ b/src/adaptors/mod.rs
@@ -631,6 +631,10 @@ where
     fn size_hint(&self) -> SizeHint {
         self.iter.size_hint()
     }
+
+    fn count(self) -> usize {
+        self.iter.count()
+    }
 }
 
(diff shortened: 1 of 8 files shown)


Philippe-Cholet:
@jswrenn Ooops... I missed that `checked_binomial` in `combinations.rs` is behind the `use_alloc` feature. Where should I move it? In `adaptors/mod.rs` I guess as it seems to be the easiest move?

jswrenn:
Yes, in adaptors/mod.rs is fine.

jswrenn (changes requested):
For consistency, could you add a specialization test for this to `tests/specializations.rs`?

jswrenn (approved):

PR #772, opened 2023-10-02
Specialize `WithPosition::fold`

Philippe-Cholet:
Each call to `f` is done with a `Position` variant known at compile time so I think it might be optimized away in some cases?

    Without specialization: [5.3231 µs 5.3487 µs 5.3806 µs]
    With specialization:    [2.6974 µs 2.7042 µs 2.7132 µs]

We should mention #755 in PRs that specialize `fold` (such as this one) to help keep track of the advancement of the TODO list I'm gonna make there.

### DIFF
--- a/src/with_position.rs
+++ b/src/with_position.rs
@@ -37,7 +37,7 @@ where
 /// Indicates the position of this element in the iterator results.
 ///
 /// See [`.with_position()`](crate::Itertools::with_position) for more information.
-#[derive(Copy, Clone, Debug, PartialEq)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
 pub enum Position {
     /// This is the first element.
     First,
--- a/src/with_position.rs
+++ b/src/with_position.rs
@@ -81,6 +81,33 @@ impl<I: Iterator> Iterator for WithPosition<I> {
     fn size_hint(&self) -> (usize, Option<usize>) {
         self.peekable.size_hint()
     }
+
+    fn fold<B, F>(mut self, mut init: B, mut f: F) -> B
+    where
+        F: FnMut(B, Self::Item) -> B,
+    {
+        if let Some(mut head) = self.peekable.next() {
+            if !self.handled_first {
+                // The current head is `First` or `Only`,
+                // it depends if there is another item or not.
+                match self.peekable.next() {
+                    Some(second) => {
+                        let first = std::mem::replace(&mut head, second);
+                        init = f(init, (Position::First, first));
+                    }
+                    None => return f(init, (Position::Only, head)),
+                }
+            }
+            // Have seen the first item, and there's something left.
+            init = self.peekable.fold(init, |acc, mut item| {
+                std::mem::swap(&mut head, &mut item);
+                f(acc, (Position::Middle, item))
+            });
+            // The "head" is now the last item.
+            init = f(init, (Position::Last, head));
+        }
+        init
+    }
 }
 
 impl<I> ExactSizeIterator for WithPosition<I> where I: ExactSizeIterator {}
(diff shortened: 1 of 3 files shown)


src/with_position.rs
```diff
@@ -81,6 +81,34 @@ impl<I: Iterator> Iterator for WithPosition<I> {
...
         self.peekable.size_hint()
     }
+
+    fn fold<B, F>(mut self, mut init: B, mut f: F) -> B
+    where
+        F: FnMut(B, Self::Item) -> B,
+    {
+        if !self.handled_first {
+            // Haven't seen the first item yet, and there might be one to give.
+            self.handled_first = true;
```
phimuemue, line 91: Do we actually need this? `self` is consumed and `handled_first` seems unused afterwards.

src/with_position.rs
```diff
@@ -81,6 +81,34 @@ impl<I: Iterator> Iterator for WithPosition<I> {
...
+        }
+        if let Some(mut head) = self.peekable.next() {
+            // Have seen the first item, and there's something left.
+            init = self.peekable.fold(init, |acc, mut item| {
+                std::mem::swap(&mut head, &mut item);
+                f(acc, (Position::Middle, item))
+            });
+            // The "head" is now the last item.
+            init = f(init, (Position::Last, head));
+        }
```
phimuemue, line 109: Shouldn't this whole thing only be relevant in the `Some(_)` case of the preceeding `match self.peekable.peek` (which would then become `match self.peekable.next`).

phimuemue (commented):
Hi @Philippe-Cholet! Nice idea.

src/with_position.rs
```diff
@@ -81,6 +81,34 @@ impl<I: Iterator> Iterator for WithPosition<I> {
...
+            if let Some(first) = self.peekable.next() {
+                // Peek to see if this is also the last item,
+                // in which case tag it as `Only`.
+                match self.peekable.peek() {
+                    Some(_) => init = f(init, (Position::First, first)),
+                    None => return f(init, (Position::Only, first)),
+                }
+            }
+        }
+        if let Some(mut head) = self.peekable.next() {
```
phimuemue, line 101: I'm somehow bothered because we call `peekable.next` here, even though we called `peekable.peek` some lines before. I'm not sure if we can easily avoid this (without convoluting control flow too much)...

Maybe something like this (only a draft - didn't test it):
````
    if let Some(head) = if !handled_first {
        if let Some(first) = self.peekable.next() {
            match self.peekable.next() {
                Some(head) => {
                    init = {
                        f(init, (Position::First, first));
                        Some(head)
                    }
                }
                None => return f(init, (Position::Only, first)),
            }
        }
    } else {
        self.peekable.next()
    } {
        // Have seen the first item, and there's something left.
        init = self.peekable.fold(init, |acc, mut item| {
            std::mem::swap(&mut head, &mut item);
            f(acc, (Position::Middle, item))
        });
        // The "head" is now the last item.
        init = f(init, (Position::Last, head));
    }
````

Philippe-Cholet:
@phimuemue 
I do not update `handled_first` field anymore, as it's consumed.
I avoid to `peek` entirely, the bench become even faster.
And rebased on master.

    [5.4031 µs 5.4195 µs 5.4353 µs] // without specialization
    [563.64 ns 565.92 ns 568.61 ns] // with new specialization

Philippe-Cholet:
Just adding that I initially tried to avoid `std::mem::swap` with `(init, head) = self.peekable.fold((init, head), ...);` which worked for me but not for the msrv CI test. It did not like assigning to a tuple. The performance was not affected.

phimuemue:
Nicely done, thanks.

As an aside: Does the mail address from your commits work?

phimuemue (approved):

Philippe-Cholet:
@phimuemue I did not check the mail address for quite some time, but I definitely did this time.

PR #775, opened 2023-10-05
Specialize `TupleCombinations::fold`

Philippe-Cholet:
Related to #755.

    [8.2569 ms 8.2733 ms 8.2907 ms]             // Benchmark `TupleCombinations::fold`
    [3.2749 ms 3.2875 ms 3.3007 ms]             // `TupleCombinations::fold`
    [3.3237 ms 3.3439 ms 3.3661 ms] (same perf) // `Tuple1Combinations::fold`
    [237.52 µs 238.81 µs 240.59 µs]             // `Tuple*Combinations::fold` (macro)

This is so much faster that I'm wandering if I'm missing something here!

I previously added the specialization test: https://github.com/rust-itertools/itertools/commit/a9aaeb4cbacee3f4e83df56b3bb45d35d9167fa9

I'm unsure we can **fold the `while let` loop** as we need to clone `iter` at every step of it.
Both attempts below successfully pass the tests but are not faster.

Maybe there is a solution involving `std::cell::???` but I'm not familiar enough with this, are you?
I tried with `itertools::rciter` but it was 60% slower (and `Rc` means the `use_alloc` feature). And with a basic `RcIter::fold` specialization, I even had an (expected) error at runtime.

    let iter = $crate::rciter(iter);
    iter.clone().fold(init, |acc, z| {
        let iter_ref = iter.rciter.borrow();
        <$P<I>>::from(iter_ref.clone())
            .map(|($($X),*,)| (z.clone(), $($X),*))
            .fold(acc, &mut f)
    })

I eventually found a way to rely on `iter.fold` using `unsafe` only (to hold `&mut I` and `&I` at the same time) but it's the same performance as the current `fold` specialization.
I'm not familiar with unsafe Rust so it might just be dumb in the first place.

    let iter_mut = &mut iter;
    let iter_ref: &I = unsafe {
        std::mem::transmute_copy(&iter_mut)
    };
    iter_mut.fold(init, |acc, z| {
        let c: $P<I> = iter_ref.clone().into();
        c.map(|($($X),*,)| (z.clone(), $($X),*))
            .fold(acc, &mut f)
    })

### DIFF
--- a/src/adaptors/mod.rs
+++ b/src/adaptors/mod.rs
@@ -736,14 +736,14 @@ macro_rules! impl_tuple_combination {
             type Item = (A, $(ignore_ident!($X, A)),*);
 
             fn next(&mut self) -> Option<Self::Item> {
-                if let Some(($($X),*,)) = self.c.next() {
+                if let Some(($($X,)*)) = self.c.next() {
                     let z = self.item.clone().unwrap();
                     Some((z, $($X),*))
                 } else {
                     self.item = self.iter.next();
                     self.item.clone().and_then(|z| {
                         self.c = self.iter.clone().into();
-                        self.c.next().map(|($($X),*,)| (z, $($X),*))
+                        self.c.next().map(|($($X,)*)| (z, $($X),*))
                     })
                 }
             }
@@ -769,13 +769,13 @@ macro_rules! impl_tuple_combination {
                 let Self { c, item, mut iter } = self;
                 if let Some(z) = item.as_ref() {
                     init = c
-                        .map(|($($X),*,)| (z.clone(), $($X),*))
+                        .map(|($($X,)*)| (z.clone(), $($X),*))
                         .fold(init, &mut f);
                 }
                 while let Some(z) = iter.next() {
                     let c: $P<I> = iter.clone().into();
                     init = c
-                        .map(|($($X),*,)| (z.clone(), $($X),*))
+                        .map(|($($X,)*)| (z.clone(), $($X),*))
                         .fold(init, &mut f);
                 }
                 init
--- a/src/adaptors/mod.rs
+++ b/src/adaptors/mod.rs
@@ -691,6 +691,13 @@ impl<I: Iterator> Iterator for Tuple1Combination<I> {
     fn count(self) -> usize {
         self.iter.count()
     }
+
+    fn fold<B, F>(self, init: B, f: F) -> B
+    where
+        F: FnMut(B, Self::Item) -> B,
+    {
+        self.iter.map(|x| (x,)).fold(init, f)
+    }
 }
 
 impl<I: Iterator> HasCombination<I> for (I::Item,) {
--- a/src/adaptors/mod.rs
+++ b/src/adaptors/mod.rs
@@ -650,6 +650,13 @@ where
     fn count(self) -> usize {
         self.iter.count()
     }
+
+    fn fold<B, F>(self, init: B, f: F) -> B
+    where
+        F: FnMut(B, Self::Item) -> B,
+    {
+        self.iter.fold(init, f)
+    }
 }
 
 impl<I, T> FusedIterator for TupleCombinations<I, T>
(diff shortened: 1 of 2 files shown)


src/adaptors/mod.rs
```diff
@@ -761,6 +761,26 @@ macro_rules! impl_tuple_combination {
...
             }
+
+            fn fold<B, F>(self, mut init: B, mut f: F) -> B
+            where
+                F: FnMut(B, Self::Item) -> B,
+            {
+                let Self { c, item, mut iter } = self;
+                init = c
+                    .map(|($($X),*,)| {
+                        let z = item.clone().unwrap();
```
phimuemue, line 772: Would it make sense / improve performance to `unwrap` only once, possibly by doing the first iteration explicitly, then saying `let z = item.as_ref().unwrap()`, and `z.clone()` within the loop?

src/adaptors/mod.rs
```diff
@@ -761,6 +761,26 @@ macro_rules! impl_tuple_combination {
...
                 checked_binomial(n, K).unwrap() + self.c.count()
             }
+
+            fn fold<B, F>(self, mut init: B, mut f: F) -> B
+            where
+                F: FnMut(B, Self::Item) -> B,
+            {
+                let Self { c, item, mut iter } = self;
+                init = c
+                    .map(|($($X),*,)| {
```
phimuemue, line 771: Do we need `$($X),*,`? Could we use the simpler `$($X,)*`?

src/adaptors/mod.rs
```diff
@@ -761,6 +761,26 @@ macro_rules! impl_tuple_combination {
...
+                        let z = item.clone().unwrap();
+                        (z, $($X),*)
+                    })
+                    .fold(init, &mut f);
+                while let Some(z) = iter.next() {
+                    let c: $P<I> = iter.clone().into();
+                    init = c
+                        .map(|($($X),*,)| (z.clone(), $($X),*))
```
phimuemue, line 779: Do we need `$($X),*,`? Could we use the simpler `$($X,)*`?

phimuemue (approved):
Hi @Philippe-Cholet, sorry for the delay. Nice that's is already faster. Feel free to merge if you want.

Is this specialization already tested in our specialization tests?

src/adaptors/mod.rs
Philippe-Cholet: In `impl_tuple_combination!(Tuple2Combination Tuple1Combination; a);`, `($($X,)*)` expands to `(a)` which is not the wanted tuple: `(a,)`.

Philippe-Cholet:
As mentionned in the too big message, the specialization test was previously added (by me).
I'm unsure I can merge this myself though. I wanted a feedback first anyway, thanks for it.

src/adaptors/mod.rs
Philippe-Cholet: It seemed like a good idea, but a test panicked on `unwrap` if done outside of `map`. I guess `item` might be none when `c` is empty (and therefore `unwrap` inside `map` is not called).

Philippe-Cholet: @phimuemue Well I could avoid `unwrap` entirely (it passes tests as expected):

    if let Some(z) = item.as_ref() {
        init = c
            .map(|($($X),*,)| {
                (z.clone(), $($X),*)
            })
            .fold(init, &mut f);
    }

src/adaptors/mod.rs
phimuemue: Maybe I'm wrong, but we already use `$($X,)*` in `size_hint`. I think it expands to `(a,)`.

Philippe-Cholet: To be sure I removed the comma and had the compiler error I thought I would have.
In `size_hint`, commas are only given to `count_ident!` macro, which does not want a tuple so it's not `(... ,)` wrapped.

Philippe-Cholet: Honestly, I think the `count_ident` macro should work without any comma too. It does not need those commas.

Philippe-Cholet: `$($X,)*` would be simpler and should have the same result.

Philippe-Cholet: There were four places where `($($X,)*)` is better than `($($X),*,)`, committed.

PR #776, opened 2023-10-06
Remove Clone requirement from unique

gilhooleyd:
This trait requirement appears to no longer be needed.

PR #777, opened 2023-10-06
Add Clone trait to Unique

gilhooleyd:
Using a Unique iterator requires the Clone trait, so require this trait when creating the Unique object.

See the discussion in https://github.com/rust-itertools/itertools/pull/776

### DIFF
--- a/src/unique_impl.rs
+++ b/src/unique_impl.rs
@@ -165,14 +165,18 @@ where
 /// See [`.unique()`](crate::Itertools::unique) for more information.
 #[derive(Clone)]
 #[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
-pub struct Unique<I: Iterator> {
+pub struct Unique<I>
+where
+    I: Iterator,
+    I::Item: Eq + Hash + Clone,
+{
     iter: UniqueBy<I, I::Item, ()>,
 }
 
 impl<I> fmt::Debug for Unique<I>
 where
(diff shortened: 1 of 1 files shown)


Philippe-Cholet:
I was not gonna suggest to make a test for it but I see your point.
I would only suggest to keep things as simple as possible with just `#[...] struct Item(u32);`.
And format to pass CI.

Philippe-Cholet:
Is the `unique` test not enough?!

gilhooleyd:
Thanks for both of your help! I think I've removed the test with the new CL.

I added the test originally when I was trying to remove `Clone`, and the struct I created didn't implement Clone.
You're right that now that Clone is required, this does the exact same thing as the existing tests.

I wish this functionality didn't require Clone, but I see now that it's necessary to do the in-place iteration while removing dups. Thank you again for your time!

jswrenn (approved):

jswrenn:
Ah, wait, could you also add it to the `struct` definition?

jswrenn:
Like so:
```rust
#[derive(Clone)]
pub struct Unique<I: Iterator>
where
    I: Iterator,
    I::Item: Eq + Hash + Clone,
{
    iter: UniqueBy<I, I::Item, ()>,
}
```

gilhooleyd:
Done! Apologies for the churn

Philippe-Cholet:
It's not a problem but could you squash these commits into one simple commit?

Issue #756, opened 2023-09-19
Write `CONTRIBUTING.md`


jswrenn:
Itertools needs a `CONTRIBUTING.md`. Key highlights should include:
- Picking the right receiver. (see #710)
- Testing (especially property-based testing).
- How to implement and robustly test `size_hint`.

A complete guide should also address:
- When to implement `DoubleEndedIterator`
- When to implement `ExactSizeIterator`
- When to implement `FusedIterator`.

(Anything else?)

Happy to provide mentorship on this.

Philippe-Cholet:
I recently contributed about each of those points (except `DoubleEndedIterator`) and there are key points I would have liked to know early on so I think I can help here. _I'm starting a draft:_ https://github.com/Philippe-Cholet/itertools/compare/master...contributing
Mentorship welcomed.

Philippe-Cholet:
- I would like some feedback on my [draft](https://github.com/Philippe-Cholet/itertools/blob/contributing/CONTRIBUTING.md) of "CONTRIBUTING.md".
- However, even if I am familiar with all kind of tests we do here, I did not write anything about testing yet as it's still unclear to me what should be suggested exactly. I think some clarification would be useful.
- After seeing "specialization tests" a few days ago, I had ideas to extend this further. I'd like to know if such thing would be of interest (we would discuss this further in another issue):

`tests/specializations.rs` definitely seems promising (I really like the `Unspecialized` struct) but all apply some "closure" **only on a fresh iterator** and _not on an iterator that has yielded elements_. A buggy iterator could behave differently if those were applied later on.
I've experimented quite a bit (and eventually ditched macros, for sanity) to check `size_hint/count/last/nth/fold/collect` (as most of our specialized methods are those) at every step of iteration (at the start and after every `next`). And it works fine but is slower as you might expect, so I added the possibility to limit this to a few steps.
Usage ([example](https://github.com/rust-itertools/itertools/commit/8c74f3331442b89c47185110e06cc5b0d9b12171)) would remain nice but my implementation is definitely way more verbose (working branch: https://github.com/rust-itertools/itertools/compare/master...Philippe-Cholet:itertools:test-specializations).

jswrenn:
Can you file a draft PR for CONTRIBUTING.md?

I'm *very* interested in seeing our specialization tests improved further!

PR #782, opened 2023-10-14
Test specializations

phimuemue:
We have more and more specializations, but so far they are not vastly tested for iterators that have already been advanced. This PR alters `test_specializations`.

### DIFF
--- a/tests/specializations.rs
+++ b/tests/specializations.rs
@@ -24,12 +24,17 @@ where
 {
     macro_rules! check_specialized {
         ($src:expr, |$it:pat| $closure:expr) => {
-            let $it = $src.clone();
-            let v1 = $closure;
-            let $it = Unspecialized($src.clone());
-            let v2 = $closure;
-            assert_eq!(v1, v2);
-        };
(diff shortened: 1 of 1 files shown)


Philippe-Cholet:
I worked some time ago on a (maybe too complex) way to test specializations (draft: https://github.com/rust-itertools/itertools/issues/756#issuecomment-1739905965) before I split focus on other things.
Here, 5 as limit might be too little, at least fior some iterators, I would suggest up to 10.

phimuemue:
@Philippe-Cholet Can you merge this? I‘ll raise the limit to ten when I have time.

Side note: I just filled this as a PR to have the CI/Tests/Checks running, but it seems, I cannot merge myself anymore (as it was the case back then with bors). Am I doing something wrong?

Philippe-Cholet (approved):

Philippe-Cholet:
All the maintainer part is new to me but I guess we can not merge a PR (except jswrenn I suppose as he is an owner) especially if it has a commit we wrote without explicit approval, something like that. I think it would have work if you approved your own changes.

PR #785, opened 2023-10-20
GroupMap: add fold_with (2) more generic init

phimuemue:
Hi, this completes #778 

@Philippe-Cholet Could you please approve this? Apparently I cannot simply push anymore to the repo.

@jswrenn I ran into this problem twice now. Am I doing something wrong? Is there a way to allow fixups like this for maintainers?

### DIFF
--- a/src/grouping_map.rs
+++ b/src/grouping_map.rs
@@ -138,7 +138,7 @@ where
     ///
     /// let lookup = (1..=7)
     ///     .into_grouping_map_by(|&n| n % 3)
-    ///     .fold_with(|_key| Default::default(), |Accumulator { acc }, _key, val| {
+    ///     .fold_with(|_key, _val| Default::default(), |Accumulator { acc }, _key, val| {
     ///         let acc = acc + val;
     ///         Accumulator { acc }
     ///      });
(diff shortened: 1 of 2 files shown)


tamird:
It seems a bit strange to accept the value since init is only called for the first key-value pair in a group. The value that the closure will be called with is not guaranteed to be deterministic AFAIK.

Philippe-Cholet:
@tamird I agree it's weird to add the `value` parameter, I'm not fond of it. But I guess it's not a big pain to ignore it by adding `,_` to the closure. And maybe there is an usage for it.

@phimuemue Can't you approve them yourself? In the "Files changed" page: click "Review changes" button, check "Approve" box, valid by clicking "Submit review" button. That's what I would do. I'd like to know where you are blocked if you are.

phimuemue:
@Philippe-Cholet I already tried this. "Approve" is disabled (seems I cannot review my own PRs).

phimuemue:
> It seems a bit strange to accept the value since init is only called for the first key-value pair in a group. The value that the closure will be called with is not guaranteed to be deterministic AFAIK.

True, but `fold_with` is merely a shorthand that avoids `Option`s. `val` is deterministic: the iterator's first item associated with its key.

Philippe-Cholet (approved):

Philippe-Cholet:
@phimuemue I previously successfully merged two PRs (by approving) that I did not create but where I committed to. I guess the key point is that you created the PR.

Issue #766, opened 2023-09-28
Improve and generalize specialization testing utilities.


jswrenn:
This issue tracks two complimentary directions for improving our specialization tests:
- @Philippe-Cholet noted that our specialization tests do not currently test partially-consumed iterators.
- We should extend our specialization testing pattern to benchmarking, so both each specialized (and unspecialized) method are tested in a [benchmark group](https://docs.rs/criterion/latest/criterion/struct.BenchmarkGroup.html).

Philippe-Cholet:
I'm definitely gonna work on improving specialization tests (as [mentioned](https://github.com/rust-itertools/itertools/issues/756#issuecomment-1739905965)) and benchmarks.

Philippe-Cholet:
After working on `combinations` & co which produce a lot of elements, a "struggle" I sometimes encounter when writing quickcheck tests is to have small enough iterators on which apply our itertools adaptors (otherwise tests would be too slow).
The current iterators we rely on are mostly the custom `Iter<u8, Inexact(default) or Exact>` and `Vec<u8>`. We can call `.take(...)` on `Iter` (but eventually lose the unfused element at the end) and truncate vectors.
But it surely does not reduce the numbers of tests. Let's say it tests against vectors of lengths 0..=128 and we truncate each to 10 inside the test then it tests against a _lot_ of iterators of length 10, mostly unnecessarily. I think the problem is that the truncation is done too late to reduce the number of tests.

So I think it would be valuable to have a way to generate small iterators, **especially for specialization tests** _which I expect to be slower if we want to test against more than fresh iterators_.
Even without const generics `Iter<T, const SIZE_LIMIT: usize>`, I'm sure we could have `Iter<T, LIMIT=NoLimit>` or `Iter<T, Max10>`. And similarly wraps vectors into `struct LittleVec<T, LIMIT=Max10>(Vec<T>);`.
Maybe you have a better idea.

This is merely a though I had, I did not work on it yet.

jswrenn:
The simplest thing we can do, right now, is to `return`, rather than truncate. That won't totally discard the test, but it will avoid most of the computational overhead of running it.

To totally discard a test, see this guidance in the quickcheck readme: https://github.com/BurntSushi/quickcheck#discarding-test-results-or-properties-are-polymorphic

Philippe-Cholet:
Thanks! ...I guess I simply forgot about `TestResult`, autocompletion would have then saved me. 😥

[...]

PR #786, opened 2023-10-20
Macro `bench_specializations`

Philippe-Cholet:
Related to #766 
Macro to be able to benchmark iterator methods.

Currently, there is only one example in the used macro (put none would lead to a compiler failure).
The expression `$iterator` is supposed to be fast, negligeable compared to the rest. Otherwise, we could use `bencher.iter_batched(|| { $iterator }, |mut it| { ... }, ...)`.

### DIFF
--- a/benches/specializations.rs
+++ b/benches/specializations.rs
@@ -1,4 +1,5 @@
 use criterion::black_box;
+use itertools::iproduct;
 use itertools::Itertools;
 
 /// Create multiple functions each defining a benchmark group about iterator methods.
@@ -118,6 +119,38 @@ macro_rules! bench_specializations {
 // Example: To bench only `ZipLongest::fold`, you can do
 //     cargo bench --bench specializations zip_longest/fold
 bench_specializations! {
+    cartesian_product {
+        {
+            let v = black_box(vec![0; 16]);
+        }
+        iproduct!(&v, &v, &v)
+    }
+    multi_cartesian_product {
+        {
+            let vs = black_box([0; 3].map(|_| vec![0; 16]));
+        }
+        vs.iter().multi_cartesian_product()
+    }
+    tuple_combinations {
+        {
+            let v = black_box((0..64).collect_vec());
+        }
+        v.iter().tuple_combinations::<(_, _, _, _)>()
+    }
+    while_some {
+        {}
+        (0..)
+            .map(black_box)
+            .map(|i| char::from_digit(i, 16))
+            .while_some()
+    }
+    with_position {
+        ExactSizeIterator
+        {
+            let v = black_box((0..10240).collect_vec());
+        }
+        v.iter().with_position()
+    }
     zip_longest {
         DoubleEndedIterator
         ExactSizeIterator
--- a/benches/specializations.rs
+++ b/benches/specializations.rs
@@ -100,6 +100,19 @@ macro_rules! bench_specializations {
                 black_box(x);
             }
         }));
+        $group.bench_function("nth_back", |bencher| bencher.iter(|| {
+            for start in 0_usize..10 {
+                for n in 0..10 {
+                    let mut it = $iterator;
+                    if let Some(s) = start.checked_sub(1) {
+                        black_box(it.nth_back(s));
+                    }
+                    while let Some(x) = it.nth_back(n) {
+                        black_box(x);
+                    }
+                }
+            }
+        }));
         $group.bench_function("rfold", |bencher| bencher.iter(|| {
             $iterator.rfold((), |(), x| {
                 black_box(x);
--- a/benches/specializations.rs
+++ b/benches/specializations.rs
@@ -1,0 +1,130 @@
+use criterion::black_box;
+use itertools::Itertools;
+
+/// Create multiple functions each defining a benchmark group about iterator methods.
+///
+/// Each created group has functions with the following ids:
+///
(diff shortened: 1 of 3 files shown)


benches/specializations.rs
```diff
@@ -0,0 +1,130 @@
...
+        }));
+        $group.bench_function("collect", |bencher| bencher.iter(|| {
+            $iterator.collect::<Vec<_>>()
+        }));
+        $group.bench_function("fold", |bencher| bencher.iter(|| {
+            $iterator.fold((), |(), x| {
+                black_box(x);
+            })
+        }));
+    };
```
phimuemue, line 93: Should this also incllude `all` (such as the specialization tests)?

benches/specializations.rs
```diff
@@ -0,0 +1,130 @@
...
+}
+
+// Example: To bench only `ZipLongest::fold`, you can do
+//     cargo bench --bench specializations zip_longest/fold
+bench_specializations! {
+    zip_longest {
+        DoubleEndedIterator
+        ExactSizeIterator
```
phimuemue, line 123: Neat idea! Maybe we should do this in specialization tests, too.

benches/specializations.rs
```diff
@@ -0,0 +1,130 @@
...
+        $group.bench_function("last", |bencher| bencher.iter(|| {
+            $iterator.last()
+        }));
+        $group.bench_function("nth", |bencher| bencher.iter(|| {
+            for start in 0_usize..10 {
+                for n in 0..10 {
+                    let mut it = $iterator;
+                    if let Some(s) = start.checked_sub(1) {
+                        black_box(it.nth(s));
+                    }
```
phimuemue, line 78: This is written slightly different than the construction of `bench_first_its` (where we explicitly compare against 0), although it seems to do the same. Is there a reason for the difference?

phimuemue (commented):
Hi @Philippe-Cholet, thanks for this. I'm inclined to merge this, but I have some questions:

* There's `$iterator` and `$first_its`. When do we use which? Could we unify (and, thus, potentially simplify code)?
* (Not directly related to this PR:) What's our general benchmarking story? Are benchmarks to be run within our CI or before we `cargo publish` a new version to ensure we have no regressions?

Please let me know if I you want to amend something or if I should merge this right away.

phimuemue (approved):

benches/specializations.rs
Philippe-Cholet: No, it just was written at different times, and even if I saw that, I did not change it because I don't think it really matters. But if you want one instead of the other, just tell me.

benches/specializations.rs
Philippe-Cholet: I listed the currently missing (stable) methods:

Iterator:
- Depends on `fold` by default: `for_each partition reduce max min max_by_key max_by min_by_key min_by`.
- Depends on `try_fold` by default: `all any find find_map position rposition cmp partial_cmp eq ne lt le gt ge`.

DoubleEndedIterator:
- nth_back
- Depends on `try_rfold` by default: rfind.

I'm inclined to add `nth_back` (not specialized yet in our crate) and not add the others as they either rely on `fold` or `try_fold` (and `try_rfold`). Obviously, we can change our mind later.

benches/specializations.rs
Philippe-Cholet: I like it too.
`ExactSizeIterator` might be unnecessary since we are I think unlikely to specialize `len` but it was easy to add once `DoubleEndedIterator` was added.

Philippe-Cholet:
`$first_its` is an array of length 1000. The first item is `$iterator`, the second is roughly `{ let mut it = $iterator; it.nth(0); it }` and so on (see `bench_first_its` for its real definition). It's meant to be immutably borrowed and is useful to benchmark `size_hint` (and the similar `len`) on not just a fresh iterator but on the first 1000 steps, it's useful because it does not measure `nth` calls as it would likely be way slower than `size_hint`. I agree it's weird at first sight.
`$iterator` is meant to be used in all other (most) cases. And I avoided any `clone` to not assume that `$iterator` is clonable.

I don't know what's "our general benchmarking story". I assume we simply use them to decide if we are improving things or if it would a perf regression in pull requests like the recent and various `fold` specializations. jswrenn might know more.

I changed the macro syntax a lot. What do you think?

Before merging, I would like @jswrenn to give a little feedback on it as he wrote the initial issue and might have a different idea in mind?

Plus, I'm discovering a lot of "criterion" parts here. What do you think about my first comment (copied below)? I think it's decent to assume `$iterator` to be fast (as allocations should be done in the initialization block), right?

> The expression `$iterator` is supposed to be fast, negligeable compared to the rest. Otherwise, we could use `bencher.iter_batched(|| { $iterator }, |mut it| { ... }, ...)`.

Philippe-Cholet:
@phimuemue I'm also a bit unsure about `()` use in `[r]fold` specializations. Maybe the compiler would optimize it away in some cases?

Philippe-Cholet:
Now that there are no opened PR about "fold" specializations, I moved benchmarks of `fold` specializations out of "bench1.rs" to put them in the new "specializations.rs".
There are no other iterator method specializations in "bench1.rs".

I should probably take time to create other "specialization benchmarks" to avoid work in future PRs. But it will be another PR though.

Philippe-Cholet:
@jswrenn As you opened #766 (to improve benchmarking), I'd like to know if this is similar to what you had in mind.

PR #787, opened 2023-10-20
docs: fixup docs of `tree_fold1`

RobWalt:
I was just reading through the docs when I noticed something odd. It's probably just a typo:

There is a recommendation to use `reduce` instead of `tree_fold1` when `f` is associative. Since `tree_fold1` won't really work as expected (unless you know what you're doing) in non-associative cases, this would imply that we don't recommend the use of `tree_fold1` at all?

I just took a guess and adjusted that part of the docs to what it most likely was supposed to be. Feel free to close this PR if I'm completely wrong here 🙈

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2499,7 +2499,8 @@ pub trait Itertools: Iterator {
     /// └─f─f─f─f─f─f
     /// ```
     ///
-    /// If `f` is associative, prefer the normal [`Iterator::reduce`] instead.
+    /// If `f` is non-associative, prefer the normal [`Iterator::reduce`] instead to prevent
+    /// unexpected behavior unless you know what you're doing.
     ///
     /// ```
     /// use itertools::Itertools;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2499,8 +2499,18 @@ pub trait Itertools: Iterator {
     /// └─f─f─f─f─f─f
     /// ```
     ///
-    /// If `f` is non-associative, prefer the normal [`Iterator::reduce`] instead to prevent
-    /// unexpected behavior unless you know what you're doing.
+    /// If `f` is associative you should also decide carefully:
+    ///
+    /// - if `f` is a trivial operation like `u32::wrapping_add`, prefer the normal
+    /// [`Iterator::reduce`] instead since it will most likely result in the generation of simpler
+    /// code because the compiler is able to optimize it
+    /// - otherwise if `f` is non-trivial like `format!`, you should use `tree_fold1` since it
+    /// reduces the number of operations from `O(n)` to `O(ln(n))`
+    ///
+    /// Here "non-trivial" means:
+    ///
+    /// - any allocating operation
+    /// - any function that is a composition of many operations
     ///
     /// ```
     /// use itertools::Itertools;


Philippe-Cholet:
@scottmcm I believe you wrote `Itertools::tree_fold1`, can you shed some light here?

RobWalt:
Ok yeah, I see what you mean. Thanks for the explanation. I think I missed some things before. Let's take a closer look at all the possible cases:

- associative:
  - `f` is simple: e.g. cases like `u32::wrapping_add`. I guess you wanted to say that this case is so simple that it's better to use the more simplistic `reduce` since it's probably better for optimizations, code gen etc, right? Some evidence that this is true would be cool
    - I can try to benchmark this a little to shine some light on it
  - `f` is complex, e.g. the format case. Here we recommend the use of `tree_fold1` since it reduces the amount of required operations from `n` to `ln2(n)`. I thought this is the main use case of `tree_fold1` and this is where my confusions and "improvements" of the docs come from
    -  As mentioned, this is imo the main use case and should be highlighted. Writing
        > If `f` is associative, prefer the normal [`Iterator::reduce`] instead.
      
        seems misleading.
- non-associative:
  - in general: e.g. the `i32::sub` case from the docs. `tree_fold1` will most likely produce different results than `reduce` in an "unintuitive" way and this is why it shouldn't be recommended in this case
  - in special cases: e.g. your example with `f32::add`. If you really know what you're doing and if you can estimate the data you're dealing with then it might make sense that you use it in this case. I think your example is a bit constructed in a way since the result of the `f32::add` operations also depends on the order of the elements in the array and not only on the folding function used. So yeah, in a way `tree_fold1` can produce a more exact result, but you need to ensure that the data is ordered so that it works out. Take this example where we shuffle the array first https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=31ea48fe708bf3778358c48a930963b5

Both of the non-associative cases should be neglacted in my opinion.

RobWalt:
Just did a very basic test on compiler explorer and both `reduce` and `tree_fold1` generate the exact same machine code for the simple case of `u32::add` as can be seen here.

https://rust.godbolt.org/z/b7K5EenEE

RobWalt:
I guess my understanding of all of this is just lacking then. I'll just close the issue. Thanks a lot for the extra explanation, that was helpful! And sorry for the needless discussion

RobWalt:
Thanks for the positive feedback :) Ok, then let me know which parts I should put the highlights on! The different use cases of `tree_fold1` and `reduce` based on the complexity of `f`?

RobWalt:
@scottmcm I adjusted the docs once again. Does this come closer to an acceptable explanation? 😅

RobWalt:
bump to not forget this issue as it's also not that big of an issue 😅

jswrenn (approved):

TomFryersMidsummer:
Some of this doesn't seem right to me, and the docs now seem a bit confusing, if not inaccurate. Specifically:

> if `f` is non-trivial like `format!`, you should use `tree_fold1` since it reduces the number of operations from `O(n)` to `O(ln(n))`

Both `reduce` and `tree_fold1` call `f` <var>n</var> − 1 times, which is <var>O</var>(<var>n</var>).

I think what this is trying to get at is that if `f` is <var>O</var>(<var>n</var>), the iterator elements are of similar sizes*, and size(`f(x, y))` = size(`x`) + size(`y`), then `reduce(f)` is <var>O</var>(<var>n</var><sup>2</sup>) and `tree_fold1` is <var>O</var>(<var>n</var> log <var>n</var>).

`format!` does satisfy these condition, but it may not be the best example. `format!("{x}{y}")` and `format!("{x}foo{y}")` are the only obvious non-trivial associative cases, and these are `concat` and `join`.

\* Not necessary, but sufficient. `reduce` gets faster if you put the largest items at the end.

RobWalt:
I'm sorry if I made the docs less correct. I can look into this when I find the time again. Otherwise feel free to open a PR yourself

PR #792, opened 2023-11-03
Test the laziness of our iterators

Philippe-Cholet:
Related to #791
Add a test to:
- track iterators that are not lazy yet ;
- help track missing `must_use` attributes with very little discipline.

It's a long list of iterators and it has some edge cases. The error message help us understand how unlazy there are.

When `let _ = ...;` is not necessary, it indicates that a `must_use` attribute is missing.

CI will fail for some time.

PR #793, opened 2023-11-03
Make `Permutations` lazy

Philippe-Cholet:
Related to #792

To make the `Permutations` adaptor lazy, the operation "prefill the lazy buffer" is now done when handling the initial `Start` state rather than at definition.

### DIFF
--- a/src/permutations.rs
+++ b/src/permutations.rs
@@ -49,18 +49,10 @@ where
 }
 
 pub fn permutations<I: Iterator>(iter: I, k: usize) -> Permutations<I> {
-    let mut vals = LazyBuffer::new(iter);
-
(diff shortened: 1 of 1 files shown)


jswrenn (changes requested):
This looks great! Can you just add a test to prevent us regressing on this?

For instance, create an iterator whose `next` call panics, then create a Permutations around it; doing so shouldn't panic.

Philippe-Cholet:
That's what #792 is for. It helped me make a list of iterators that are not lazy yet and will ensure we don't regress on that front once merged. Make a test now would be soon a duplicate.

jswrenn (approved):

PR #795, opened 2023-11-05
Make `Combinations` lazy

Philippe-Cholet:
Related to #792

To make the `Combinations` adaptor lazy, the operation "prefill the lazy buffer" is now done when handling the first element rather than at definition.

However, the value returned by the `n` method is changed when called on a fresh `it.combinations(k)` since the "prefill" operation was delayed. I'm perplex on why the method `n` was made public in the first place as the value is subject to change (said in its documentation). But it is public and this change is therefore a (small) breaking one, right?


jswrenn (approved):
Having given this a few days of thought, I don't think this qualifies as a breaking change. Yes, the observable behavior changes. But that's also true even without this method, and we haven't been considering our other lazy-ification efforts to be breaking changes (rather, they're really bugfixes).

PR #797, opened 2023-11-06
Make `Intersperse[With]` lazy

Philippe-Cholet:
Related to #792 
To make `Intersperse[With]` adaptors lazy, I use a nested option which is what `core::iter::Peekable` uses internally.
Maybe we should have benchmark about `next` to see if it implies a performance cost. In that case, I'll wait for #786.


src/intersperse.rs
```diff
@@ -54,18 +54,17 @@ where
 {
     element: ElemF,
     iter: Fuse<I>,
-    peek: Option<I::Item>,
+    peek: Option<Option<I::Item>>,
```
jswrenn, line 57: With the added nesting, could you just write a comment here about how the value of `peek` should be interpreted? (`None`, `Some(None)`, `Some(Some(...))`?

jswrenn (changes requested):

jswrenn (approved):

PR #799, opened 2023-11-12
More specialization tests

Philippe-Cholet:
- To do the next point, `test_specializations` need clonable iterators which helped find iterators that should be clonable but were not yet. 3 very small commits.
- I do not want to add specialization tests one by one in future pull requests so this is **_mostly done_**! 1 big but straightforward commit.
- For two tests I previously wrote, I now discard inputs (in the quickcheck-way) instead of truncating them. 1 small commit.

Now about **_mostly done_**... Not all of our iterators are tested here:
- `group_by`
- `chunks`
- `tee`
- `rciter` 
- `test_specializations` need clonable iterators but `&mut iterator` is not clonable, Therefore, we can't clone `PeekingTakeWhile` and `TakeWhileRef`. We need to find a way to test them too.

Finally a **bug**... **`multi_cartesian_product`** is not tested here because specializations of `last` and `count` **FAIL** sometimes, and I don't want this pull request to be blocked by it! It requires further investigation and will have a dedicated pull request.
**EDIT:** Apparently, the iterator is not fused even when all child iterators are, and this was the [intended behavior](https://github.com/rust-itertools/itertools/pull/235#issuecomment-347604290), that's confusing.

### DIFF
--- a/src/adaptors/mod.rs
+++ b/src/adaptors/mod.rs
@@ -935,6 +935,7 @@ where
 ///
 /// See [`.filter_map_ok()`](crate::Itertools::filter_map_ok) for more information.
 #[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
+#[derive(Clone)]
 pub struct FilterMapOk<I, F> {
     iter: I,
     f: F,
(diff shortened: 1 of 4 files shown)


jswrenn (approved):
This looks good to me. You can go ahead and merge this as-is, or you can add a test for `multi_cartesian_product` and mark it as `#[ignore]` (so we don't forget about it, but it also doesn't fail CI).

PR #800, opened 2023-11-13
Make `Product` lazy

Philippe-Cholet:
Related to #792.
To make the `Product` adaptor lazy, I use a nested option, just like I did in #797.
Then the method `cartesian_product` and the macro `iproduct` will be lazy.

### DIFF
--- a/src/adaptors/mod.rs
+++ b/src/adaptors/mod.rs
@@ -308,7 +308,7 @@ where
     I: Iterator,
 {
     a: I,
-    a_cur: Option<I::Item>,
+    a_cur: Option<Option<I::Item>>,
     b: J,
     b_orig: J,
 }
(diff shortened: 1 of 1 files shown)


src/adaptors/mod.rs
```diff
@@ -308,22 +308,25 @@ where
     I: Iterator,
 {
     a: I,
-    a_cur: Option<I::Item>,
+    /// `a_cur` is None while no item have been taken out of `a` (at definition).
```
jswrenn, line 311:
```suggestion
    /// `a_cur` is `None` while no item have been taken out of `a` (at definition).
```

jswrenn (approved):

src/adaptors/mod.rs
Philippe-Cholet: Commit updated.

jswrenn (approved):

Issue #710, opened 2023-07-02
Ownership issues for `Itertools` methods


Philippe-Cholet:
Following the logic of my previous comment (_https://github.com/rust-itertools/itertools/issues/709#issuecomment-1616526337_), I was worried about ownership for other methods, here's what I found:

It consumes the iterator completely, it should take by value, right?
Once called, the iterator still exists but is always empty, then it's dropped later.

    fn join(&mut self, sep: &str) -> String;

Transfer ownership to another iterator: it should take by value (then people can call `.by_ref()` if they want).

    fn take_while_inclusive... // subject of #709
    fn fold_while<B, F>(&mut self, init: B, mut f: F) -> FoldWhile<B>;
    fn peeking_take_while<F>(&mut self, accept: F) -> PeekingTakeWhile<Self, F>;
    // This one below seems weird (and super old), just `.by_ref().take_while(accept)` shortcut ? Then it's ok!
    fn take_while_ref<F>(&mut self, accept: F) -> TakeWhileRef<Self, F>;

Methods that consume only part (in some case) of the iterator (should mutably borrow, people might want to consider remaining elements). But it's **not a real issue** as people can call `.by_ref()`.

    fn collect_tuple<T>(mut self) -> Option<T>;
    fn try_collect<T, U, E>(self) -> Result<U, E>; // will be deprecated anyway once it's stabilized in Iterator
    fn process_results<F, T, E, R>(self, processor: F) -> Result<R, E>;

The rest of the methods follow the "std Iterator logic" described in my previous comment.

PR #806, opened 2023-11-23
More specialization benchmarks

Philippe-Cholet:
I do not want us to add specialization benchmarks one by one in future pull requests so this is _**mostly done**_!
Not all of our iterators are benchmarked here:

- `group_by`, `chunks`, `tee`, `rciter` just like in #799.
- `peeking_take_while` and `take_while_ref` do not take ownership (see #710) of the iterator they adapt and it's therefore problematic to add them to the `bench_specializations` macro.
- `process_results` does not give an iterator but process one and we can't create a `ProcessResults` object outside of the library.
- `MapForGrouping` used for `GroupingMapBy` but it's strictly internal to the library, we will need to benchmark a (non-iterator) method of `GroupingMapBy` to benchmark the future `MapForGrouping::fold`.
- `unfold`, `iterate`: infinite iterators, some benchmarks would not end. We can avoid running them but maybe we should handle them differently later.

After this, we will be able to specialize most `fold` methods (see #755) faster because their specialization tests/benchmarks are written and we will just need to write the specialization in a commit, run tests, benchmark it before/after and review the PR.

### DIFF
--- a/benches/specializations.rs
+++ b/benches/specializations.rs
@@ -1,3 +1,5 @@
+#![allow(unstable_name_collisions)]
+
 use criterion::black_box;
 use itertools::iproduct;
 use itertools::Itertools;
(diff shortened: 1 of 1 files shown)


jswrenn (approved):
This is excellent!

jswrenn:
Go ahead and click `Merge` if you're ready to do so.

PR #807, opened 2023-11-25
Test double ended specializations

Philippe-Cholet:
We don't have any `rfold/nth_back` specialization yet but it will come as I intend to do `rfold` specializations alongside `fold` ones (benchmarks are ready too).

Apart from `rciter` for which I did not add any specialization test, all our double ended iterators are included here.

### DIFF
--- a/tests/specializations.rs
+++ b/tests/specializations.rs
@@ -86,6 +86,44 @@ where
     }
 }
 
+fn test_double_ended_specializations<I>(it: &I)
+where
+    I::Item: Eq + Debug + Clone,
+    I: DoubleEndedIterator + Clone,
+{
+    macro_rules! check_specialized {
+        ($src:expr, |$it:pat| $closure:expr) => {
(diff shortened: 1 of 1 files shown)


phimuemue (approved):
Thanks, looks good to me.

A thought: Should we make `test_specializations` a macro that by default checks `Iterator` and `DoubledEndedIterator` (and all other specialized things), and only omits checks them if actively opted-out?

jswrenn (approved):

Philippe-Cholet:
@phimuemue So opt-in or opt-out?
If tomorrow we implement `DoubleEndedIterator` on some existing iterator, then in both cases, we have to update the test here. Therefore both are not ideal so I think it's merely a preference, and I prefer functions/traits over macro when possible.

I sure would prefer to merge those two functions in whatever form that detect if it needs to be done or not (I mean, checking if each of them should have a double ended test was not really quick to do), here and for benchmarks. But I don't see a way to do it.

PR #816, opened 2023-12-08
`Positions` based on `Enumerate`

Philippe-Cholet:
Following https://github.com/rust-itertools/itertools/pull/813#issuecomment-1847585114, with this PR, `Positions` now uses `Enumerate` internally.

- `next` and `next_back` are now based on `find_map` instead of "while-next" loops. It would be useful performance-wise for iterators specializing `try_[r]fold`. I intend to do the same for other while-next loops (where benchmarks show times decrease, up to -20%).
- Stable benchmarks here.
- Doing this, I first made a silly mistake. I therefore add a test that would have helped me, and fix a little error in the documentation (it could alternatively use `filter_map` to avoid `*` copy).

### DIFF
--- a/src/adaptors/mod.rs
+++ b/src/adaptors/mod.rs
@@ -16,7 +16,7 @@ pub use self::multi_product::*;
 
 use crate::size_hint::{self, SizeHint};
 use std::fmt;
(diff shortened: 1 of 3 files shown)


phimuemue (approved):
Thank you, @Philippe-Cholet

PR #822, opened 2023-12-15
Document order guarantees for combinations

### LINKED ISSUE #819: Can the order of Combinations be relied on?
Does `itertools::Combinations` makes any guarantees about the *order* in which it yields combinations?

The docs provide this example:

```rust
use itertools::Itertools;

let it = (1..5).combinations(3);
itertools::assert_equal(it, vec![
    vec![1, 2, 3],
    vec![1, 2, 4],
    vec![1, 3, 4],
    vec![2, 3, 4],
]);
```

Which at a glance is constructing the combinations by "indexing" the iterator `(1..5)` with lexicographically sorted indices `[0, 1, 2]`, `[0, 1, 3]`, `[0, 2, 3]`, and `[1, 2, 3]`.

I am wondering if this is something I can rely on, or if it is an implementation detail that should not be relied on?

ryanavella:
Fixes #819 by documenting the order guarantees for Itertools adapters.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1591,6 +1591,11 @@ pub trait Itertools: Iterator {
     /// let it: TupleCombinations<Range<u32>, (u32, u32, u32)> = (1..5).tuple_combinations();
     /// itertools::assert_equal(it, vec![(1, 2, 3), (1, 2, 4), (1, 3, 4), (2, 3, 4)]);
     /// ```
+    ///
+    /// # Guarantees
+    ///
+    /// If the adapted iterator is deterministic,
+    /// this iterator adapter yields items in a reliable order.
     fn tuple_combinations<T>(self) -> TupleCombinations<Self, T>
     where
         Self: Sized + Clone,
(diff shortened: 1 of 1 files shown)


Philippe-Cholet:
CI failed because rustfmt does not accept trailing spaces.

The question is if we want this guarantee to be library wide or limited to some adaptors like `Combinations`. I'll let @jswrenn decide on that.

ryanavella:
> The question is if we want this guarantee to be library wide or limited to some adaptors like Combinations. I'll let @jswrenn decide on that.

Ah yes, I hope I didn't misunderstand the conclusion from the other issue. I originally wanted a guarantee for Combination-like adapters, but I interpreted your comments as saying we don't intend to break iteration order for any adapters.

I could add a blanket "unless otherwise noted" clause, if there are specific adapters we'd like to leave open to further optimizations?

jswrenn:
Sorry for the misunderstanding. I don't think we can guarantee order for all adaptors, but I am okay guaranteeing it for the combinations adaptors.

ryanavella:
Thank you for the clarification, I've updated this PR so that the guarantee is only shown under `Itertools::combinations` and `Itertools::tuple_combinations`. Let me know if all looks good to you.

Philippe-Cholet:
@jswrenn What do you think of his formulation?

jswrenn (changes requested):
Whoops, I had a draft review I thought I had posted. Thanks for the ping. It's just some minor nits.

Philippe-Cholet:
@ryanavella If you could commit jswrenn' suggestions and squash into a single commit, I would gladly merge this.

Philippe-Cholet (approved):

Philippe-Cholet:
@jswrenn My approval is not enough to get this merged.

jswrenn (approved):

PR #827, opened 2023-12-20
use self

adamnemecek:


Philippe-Cholet:
From memory, we support a minimum version of Rust of 1.43.1 (see `Cargo.toml` four an accurate version) where `Option::zip` and `let-else` are not supported yet among other things, or I would use them myself.
`Vec::new()` vs `vec![]`, I'm not sure it's worth your time. And what's the benefit? Shorter code.

The repo is formatted with rustfmt so I'm not sure what you are doing with your `fmt` commit.

About using `Self` I agree but I think there is a PR about clippy lints that surely address this, a PR that we should somehow merge, even if only partially (in it, it suggests the integratation of clippy to our CI which might be tricky).

jswrenn:
I don't think we're likely to accept this. Large PRs are hard to review and need to have value proportionate to their review effort. The difference here is cosmetic. For this, I can imagine a maintainer denying clippy's `use_self` lint, then running `cargo clippy --fix`.

PR #831, opened 2023-12-28
Fix spelling error in documentation.

sgtatham:
The primary trait in this crate is called `Itertools`, not `IterTools`.

(This actually confused me - by sheer bad luck, `exactly_one` was the first itertools method I ever tried to use, so that docs comment was the very first one I found in my search, so I spelled it `IterTools` in my source code, and got a compile error!)

### DIFF
--- a/src/exactly_one_err.rs
+++ b/src/exactly_one_err.rs
@@ -8,7 +8,7 @@ use either::Either;
 
 use crate::size_hint;
 
-/// Iterator returned for the error case of `IterTools::exactly_one()`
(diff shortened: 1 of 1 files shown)


jswrenn (approved):

Philippe-Cholet:
@sgtatham Nicely spotted, thanks. I ran some command line to find other spelling errors just in case, found none.

Huge fan of your [puzzle collection](https://www.chiark.greenend.org.uk/~sgtatham/puzzles/) by the way (unless it's not yours).

sgtatham:
Thank you! (Yes, it is mine)

PR #834, opened 2024-01-03
Add documentation for `multi_cartesian_product` when iterator is empty

YizhePKU:
Document the correct behavior of nullary cartesian product.

[The bug](#337) was submitted in 2019 and #603 fixes it just fine. Please just merge both this PR and PR #603 and call it a day. I understand that review time is scarce, but having automated tests pass should be good enough for this kind of change.

PR #835, opened 2024-01-03
Rewrite `MultiProduct`

### LINKED ISSUE #337: multi_cartesian_product of zero iterators should produce one empty Vec

Philippe-Cholet:
Fixes #337. Closes #603 since this is an alternative.

#834 legitimately asked that we work on the bugfix #603. But after rebasing it, I saw that I could finally fuse the iterator (and therefore have a working specialization test). I thought of using #603 as a base but I eventually chose to make my own changes, starting over. I however credited @JakobDegen where it made sense.

I made commits as atomic as I could.

So this fuses the iterator which is a breaking change, which won't affect most people.

### DIFF
--- a/src/adaptors/multi_product.rs
+++ b/src/adaptors/multi_product.rs
@@ -102,10 +102,13 @@ where
     type Item = Vec<I::Item>;
 
     fn next(&mut self) -> Option<Self::Item> {
+        // This fuses the iterator.
         let inner = self.0.as_mut()?;
         match &mut inner.cur {
             Some(values) => {
                 debug_assert!(!inner.iters.is_empty());
+                // Find (from the right) a non-finished iterator and
+                // reset the finished ones encountered.
                 for (iter, item) in inner.iters.iter_mut().zip(values.iter_mut()).rev() {
                     if let Some(new) = iter.iter.next() {
                         *item = new;
@@ -136,9 +139,12 @@ where
 
     fn count(self) -> usize {
         match self.0 {
-            None => 0,
+            None => 0, // The cartesian product has ended.
             Some(MultiProductInner { iters, cur }) => {
                 if cur.is_none() {
+                    // The iterator is fresh so the count is the product of the length of each iterator:
+                    // - If one of them is empty, stop counting.
+                    // - Less `count()` calls than the general case.
                     iters
                         .into_iter()
                         .map(|iter| iter.iter_orig.count())
@@ -151,6 +157,7 @@ where
                         })
                         .unwrap_or_default()
                 } else {
+                    // The general case.
                     iters.into_iter().fold(0, |mut acc, iter| {
                         if acc != 0 {
                             acc *= iter.iter_orig.count();
@@ -164,7 +171,7 @@ where
 
     fn size_hint(&self) -> (usize, Option<usize>) {
         match &self.0 {
-            None => (0, Some(0)),
+            None => (0, Some(0)), // The cartesian product has ended.
             Some(MultiProductInner { iters, cur }) => {
                 if cur.is_none() {
                     iters
@@ -186,6 +193,7 @@ where
 
     fn last(self) -> Option<Self::Item> {
         let MultiProductInner { iters, cur } = self.0?;
+        // Collect the last item of each iterator of the product.
         if let Some(values) = cur {
             let mut count = iters.len();
             let last = iters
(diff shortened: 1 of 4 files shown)


phimuemue (approved):
Thank you @Philippe-Cholet .

We should do this.

jswrenn:
Thanks for this! I'll give it a review tomorrow!

Philippe-Cholet:
_**TL;DR** The choice I made in https://github.com/rust-itertools/itertools/pull/835/commits/db5d9062a1e399f86807f773615aa6c747be3275 does not seem to affect performance._

I just tried `cur: Option<I::Item>,` in each `MultiProductIter` (as before in "master") instead of `cur: Option<Vec<I::Item>>,` in `MultiProductInner` (as in this PR), benchmarked both and got similar performances (both a bit slower for `next` than in "master" +8%, but with bugfix and fused).

jswrenn (changes requested):
This looks good to me, but needs some more documentation. For starters, could you add some more documentation to the type definitions?

src/adaptors/multi_product.rs
```diff
@@ -171,81 +102,114 @@ where
...
+                        .unwrap_or_default()
+                } else {
+                    iters.into_iter().fold(0, |mut acc, iter| {
+                        if acc != 0 {
+                            acc *= iter.iter_orig.count();
+                        }
+                        acc + iter.iter.count()
+                    })
+                }
+            }
```
jswrenn, line 159: This is rather complicated logic, and if you have the time to provide comments for it (and the other specialized methods), that would be much appreciated. I'd love to see `multi_product.rs` get to the point where it's obvious how everything fits together. I'd suggest some comments here, myself, but I don't have a great understanding of how this works. I don't think there are many other places in itertools where our methods peek through three or four layers of nested types. 

This *is* a breaking change, so we can be a bit more patient in merging it.

jswrenn (approved):

src/adaptors/multi_product.rs
Philippe-Cholet: Indeed, the `try_fold(...).unwrap_or...` is basically an optimized `.product()` and deserves a comment. I'm gonna add some comments where I think it's valuable.

src/adaptors/multi_product.rs
```diff
@@ -136,9 +139,12 @@ where
 
     fn count(self) -> usize {
         match self.0 {
-            None => 0,
+            None => 0, // The cartesian product has ended.
```
phimuemue, line 142: Would it make sense to have an `enum Something { MultiProductInner(iters, cur), CartesianProductHasEnded }` instead of an `Option` here?

src/adaptors/multi_product.rs
```diff
@@ -122,10 +123,11 @@ where
             // Only the first time.
             None => {
```
phimuemue, line 127: Analogously, would it make sense to replace this `Option` by `enum CurrentValues { NotYetPopulated, Populated(Vec<_>) } `?

src/adaptors/multi_product.rs
Philippe-Cholet: It makes sense and the other too. But it's just an option with customized names and without the methods and `?` shortcut.
On "permutations", I remember doing the opposite change and really simplify things with it.
In `next`, there is `let inner = self.0.as_mut()?;` that would be quite complicated with such change.

Philippe-Cholet: Honestly I hesitated to add a comment here as the `None` variant for `.0` is already documented.

jswrenn: A compromise:
```rust
pub use {
    Option as State,
    Option::None as ProductEnded,
    Option::Some as ProductInProgress,
};
```

Philippe-Cholet:
I rebased on master and added 3 commits to comment "state and values" with code.

jswrenn (approved):

phimuemue (approved):
Thanks

Philippe-Cholet:
I merely rebased (no conflict) to see if the new CI still accepts this.
Do we merge this (and #853) now or in the next release?

jswrenn:
I'll cut a release with our non-breaking changes first, probably tomorrow.

Philippe-Cholet:
Now that "0.12.1" is released, I renamed the associated milestone and created a new one.
I think it's time to merge this and _approve_/merge #853 as well, right?

PR #837, opened 2024-01-10
Faster `find_position`

Philippe-Cholet:
`find` usually rely on `try_fold` so this change should improve performance for iterators specializing `find/try_fold`.

Credits to #618, I also noticed https://github.com/phimuemue/rust-itertools/tree/simplify_find_position


phimuemue (approved):
Thank you.

Issue #264, opened 2018-03-03
Feature: n-dimensional product over the same iterator


lucasem:
So that code like this:

```rust
iproduct!(0..4, 0..4, 0..4)
```

becomes the more succinct:

```rust
iproduct!(0..4; 3)
```

This would mandate vectors instead of tuples for the items.
It would help for when the dimension isn't known until runtime:

```rust
fn truth_table(dim: usize) → Box<Iterator<Item = Vec<bool>>> {
    Box::new(iproduct!([false, true].iter(); dim))
}
```

burdges:
Is this so common as to warrant an `iproduct!` like convenience macro?  It's easy enough to do this now with `std::iter::repeat([fasle,  true].iter()).take(dim).multi_cartesian_product()` right?

lucasem:
Ah that's a nice solution, thanks! The `multi_cartesian_product` docs are coarse to read through.

Having this feature in the `iproduct!` macro could still be nice, but I don't stand by it as strongly after seeing your approach:

```rust
fn truth_table(dim: usize) → Box<Iterator<Item = Vec<bool>>> {
    Box::new(iter::repeat(vec![false, true]).take(dim).multi_cartesian_product()
}
```

bluss:
I wince a bit at all of this, since any product / flat_map or other "nested loop" iterator will have performance or code gen problems in Rust. But if that is not of primary concern, then it's ok.  Demo of the difference is in https://github.com/bluss/rust-itertools/pull/222#issue-142970031 (where you can see that fold/for_each, representing internal iteration, don't have to have the problem.)

bluss:
I don't mind adding that syntax to the macro, with the caveats that it has to be feasible to implement and that one thinks over what the duplicate syntax/ownership/cloning semantic is for the repetition.

bluss:
Maybe I should throw in something really ridiculous and then we can think of what it should result in:

```rust
let mut i = 0;
iproduct!( { i += 1; i..10 }; 3);
```

Note that this might not be so ridiculous, the `i += 1` can be replaced with a mutating method and its return value.

tobz1000:
`multi_cartesian_product` will have an order-of-magnitude worse performance than `iproduct`, so if the number of repeats is known at compile time (and small), and the code is at all performance-sensitive, `iproduct` should be preferred. And as @bluss says, a nested for-loop will be faster still.

[...]

PR #841, opened 2024-01-10
Add macro `iunpack`

A4-Tacks:
Add it, simply and quickly unpack iterators to patterns.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -394,6 +394,293 @@ macro_rules! chain {
     };
 }
 
+#[macro_export]
+/// Use pattern unpack iterator
+///
+/// The expression after the equal sign must implement [`IntoIterator`].\
+/// The final pattern may be followed by a trailing comma.
+///
+/// Use else to handle iterator length mismatch or pattern mismatch
+///
+/// - Use `*name` pattern any elements to [`Vec`],
+///   use [`DoubleEndedIterator`] pattern end elements.
+/// - Use `*name: Type` pattern any elements collect into impl [`FromIterator`].
+/// - Use `*=iter` pattern start and end elements, do not check iter is stopped.
+///   Internally use the given variable name to store iterator for future use.
+/// - Use `**name` like `*name`, but use [`Iterator`]
+///
+/// [`FromIterator`]: std::iter::FromIterator
+/// [`Iterator`]: std::iter::Iterator
+/// [`IntoIterator`]: std::iter::IntoIterator
+/// [`DoubleEndedIterator`]: std::iter::DoubleEndedIterator
+/// [`Vec`]: std::vec::Vec
+///
+/// # Examples
+///
+/// Sized iterator
+/// ```
+/// # use itertools::{assert_equal, iunpack};
+/// assert_eq!(loop {
+///     iunpack!(a, b, c, d, e = 0..5; else panic!());
+///     break (a, b, c, d, e);
+/// }, (0, 1, 2, 3, 4));
+///
+/// assert_eq!(loop {
+///     iunpack!(a, b, c, d, e = 0..3; else(err) { break err });
+///     panic!();
+/// }, 3); // fail, not enough values
+///
+/// assert_eq!(loop {
+///     iunpack!(a, b, c, d, e = 0..7; else(err) { break err });
+///     panic!();
+/// }, 5); // fail, too many values
+/// ```
+///
+/// Any size iterator
+/// ```
+/// # use itertools::{assert_equal, iunpack};
+/// # use std::collections::HashSet;
+/// assert_eq!(loop {
+///     iunpack!(a, b, *c, d, e = 0..8; else panic!());
+///     break (a, b, c, d, e);
+/// }, (0, 1, vec![2, 3, 4, 5], 6, 7));
+///
+/// assert_eq!(loop {
+///     iunpack!(a, b, *=c, d, e = 0..8; else panic!());
+///     break (a, b, c.collect::<Vec<_>>(), d, e);
+/// }, (0, 1, vec![2, 3, 4, 5], 6, 7));
+///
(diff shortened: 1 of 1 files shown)


phimuemue:
Hi there, this is some impressive macro-programming.

But from a user's perspective: Is this *really* better than a manual `collect_tuple` (and its cousins), or - in the special cases - manually collecting the iterator into separate variables? `iunpack` requires the user to learn yet another mini-language (see the `*`, `**`, `*c=`, etc). We could justify this if the macro covered *all* use cases - but e.g. can it e.g. unpack into a HashSet, but only values that match the pattern `1 | 2`? And if so, how easy would it be to specify this?

I think itertools should provide (relatively) fool-proof and intuitive helpers, and I don't think `iunpack` fits this category. So my gut feeling would be to decline this PR.

A4-Tacks:
> Hi there, this is some impressive macro-programming.
> 
> But from a user's perspective: Is this _really_ better than a manual `collect_tuple` (and its cousins), or - in the special cases - manually collecting the iterator into separate variables? `iunpack` requires the user to learn yet another mini-language (see the `*`, `**`, `*c=`, etc). We could justify this if the macro covered _all_ use cases - but e.g. can it e.g. unpack into a HashSet, but only values that match the pattern `1 | 2`? And if so, how easy would it be to specify this?
> 
> I think itertools should provide (relatively) fool-proof and intuitive helpers, and I don't think `iunpack` fits this category. So my gut feeling would be to decline this PR.

I think it's easy to understand, and we have macro expand, right?
Its operation is like extending the slice pattern to iterators

A4-Tacks:
For example, it can unpack structures or verify the correctness of enumerations when unpacking iterators, which will be extremely convenient and can simplify many codes

A4-Tacks:
And this macro implementation can effectively handle iterators with variable numbers of elements for quick single branch matching at the beginning and end, without the need for lengthy code.
One of the powerful features of this macro is the use of buffers to rely solely on one-way iterators to match tails

Philippe-Cholet:
It seems powerful but quite unusual at first sight, and all those loop-break around the macro (but I guess it could be presented differently). I get that you understand everything here as you made it but it's not that easy. I would agree with phimuemue.

After looking at the internals, I need to say that we currently use stable Rust 1.43.1 and "let-else" is not available for us yet (and 1.65 is not for tomorrow).

Honestly after all the debate of `iproduct!(something; n)` (#264) that should be simple enough (which I intend to solve in a near future), this definitely seems on another level.

A4-Tacks:
> It seems powerful but quite unusual at first sight, and all those loop-break around the macro (but I guess it could be presented differently). I get that you understand everything here as you made it but it's not that easy. I would agree with phimuemue.
> 
> After looking at the internals, I need to say that we currently use stable Rust 1.43.1 and "let-else" is not available for us yet (and 1.65 is not for tomorrow).
> 
> Honestly after all the debate of `iproduct!(something; n)` (#264) that should be simple enough (which I intend to solve in a near future), this definitely seems on another level.

Regarding the issue of if-else, may it be possible to modify it to an internal code block?
I still hope it will be merged.

A4-Tacks:
> It seems powerful but quite unusual at first sight, and all those loop-break around the macro (but I guess it could be presented differently). I get that you understand everything here as you made it but it's not that easy. I would agree with phimuemue.
> 
> After looking at the internals, I need to say that we currently use stable Rust 1.43.1 and "let-else" is not available for us yet (and 1.65 is not for tomorrow).
> 
> Honestly after all the debate of `iproduct!(something; n)` (#264) that should be simple enough (which I intend to solve in a near future), this definitely seems on another level.

Having changed it to use if-let, is it available now?

A4-Tacks:
Can it run now? Can it be merged?

Philippe-Cholet:
I just let our CI run, it passed (thanks to the removed "let-else") which does not mean it'll be merged.

The "merge process" might be considered slow for new additions and especially for something that is not easy like this. But if we add something, we commit ourselves to maintain it.
I'd first like to know if the changes change @phimuemue's opinion on this as it seems a bit more friendly than initially.

phimuemue:
> I just let our CI run, it passed (thanks to the removed "let-else") which does not mean it'll be merged.
> 
> The "merge process" might be considered slow for new additions and especially for something that is not easy like this. But if we add something, we commit ourselves to maintain it. I'd first like to know if the changes change @phimuemue's opinion on this as it seems a bit more friendly than initially.

I hope I do not come across as rude, but this did not change my opinion. Imho this macro - while impressive - has downsides:
* It wraps trivial behavior (albeit in a non-trivial way). Sure, nice for code golf, but I think it not desperately needed in general.
* **Too specific:** Could I collect the first three elements into a HashMap if they match `(1|2)`, and collect the remaining ones into `b, c, d` if they are odd numbers, collect the rest into a `Vec`, and have a dedicated error depending on whether it fails to collect the `HashMap`, `b, c, d` or the `Vec`. No problem with manual collecting, but impossible/not straightforward with this macro.
 A slight change in requirements may thus require users to rewrite from `iunpack` to manual collect.
* It promotes an **own mini-language** that is not capable of expressing sufficiently many use cases, and I think collecting manually is clearer.
* It is imho **not straightforward** what this macro does. (Just from a skim, I see it introduces a `loop`, or it populates an array and uses `rotate_left` afterwards, while simply skipping the elements would be more performant. All of which is unexpected in my book.) 

I think this has its place in a code-golf/AoC-like crate, but not in itertools. But I won't veto if @Philippe-Cholet or @jswrenn wants to merge.

Philippe-Cholet:
I think we three agree that it is a nice and powerful macro that does not really fit our crate.
It should belong somewhere, maybe its own crate, as I do not know which crate it could be a good fit.

phimuemue:
Closing

PR #842, opened 2024-01-11
`iterate`: document a possible panic

### LINKED ISSUE #600: `Iterate` advances the source iterator one element ahead
Calling `Iterate::next` fetches "the next after next" element [here](https://github.com/rust-itertools/itertools/blob/master/src/sources.rs#L159) then returns the actual "next" element.  This behavior sometimes triggers unintended panics when overiteration is intended to be guarded by e.g. `TakeWhile`.

Philippe-Cholet:
Resolves #600. See https://github.com/rust-itertools/itertools/issues/600#issuecomment-1884539907 for why I don't see an alternative.

### DIFF
--- a/src/sources.rs
+++ b/src/sources.rs
@@ -177,7 +177,7 @@ where
 /// ```
 /// use itertools::iterate;
 ///
-/// itertools::assert_equal(iterate(1, |&i| i * 3).take(5), vec![1, 3, 9, 27, 81]);
+/// itertools::assert_equal(iterate(1, |i| i % 3 + 1).take(5), vec![1, 2, 3, 1, 2]);
 /// ```
 ///
 /// **Panics** if compute the next value does.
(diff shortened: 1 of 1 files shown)


phimuemue (approved):
I approve this, because having the docs is better than not having them.

But isn't the interesting part here that the "next next" value is computed in `next`? And: Doesn't `successors` potentially suffer the same problem (albeit able to circumvent it using `None`)? (see https://doc.rust-lang.org/beta/src/core/iter/sources/successors.rs.html#48)

Philippe-Cholet:
Without `Clone`-able values, `iterate` and `successors` are forced to compute it ahead-of-time while we still have access to the value, it's the trade-off: not cloned but ahead-of-time computation. But being able to return `None` in `successors`' closure seems a better API decision to me.

phimuemue (approved):

PR #846, opened 2024-01-13
pin the toolchain version used by clippy

danieleades:
see discussion in #618

- pin clippy version for consistent lints across all PRs
- modify MSRV action such that dependabot will bump version of clippy toolchain but not touch msrv

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ jobs:
           ]
     steps:
       - uses: actions/checkout@v4
-      - uses: dtolnay/rust-toolchain@stable
+      - uses: dtolnay/rust-toolchain@1.75.0
         with:
           components: clippy
       - run: RUSTFLAGS="--deny warnings" cargo clippy ${{ matrix.features }}
@@ -41,7 +41,9 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: taiki-e/install-action@cargo-no-dev-deps
-      - uses: dtolnay/rust-toolchain@1.43.1
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: 1.43.1
       - run: cargo no-dev-deps check
 
   test:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,9 +58,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: dtolnay/rust-toolchain@master
+      - uses: dtolnay/rust-toolchain@stable
         with:
-          toolchain: stable
           components: rustfmt
       - run: cargo fmt --check
 


Philippe-Cholet:
I was working on this but was kinda blocked (I mostly discover dependabot to be honest), so thanks.

What bug me about this kind of change is #758 that we did not want.

Are you saying that `with: toolchain: 1.43.1` will make dependabot ignore it? That sounds so simple as I was looking in another direction to solve this ([use "ignore" option](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#ignore)).

danieleades:
> I was working on this but was kinda blocked (I mostly discover dependabot to be honest), so thanks.
> 
> What bug me about this kind of change is #758 that we did not want.
> 
> Are you saying that `with: toolchain: 1.43.1` will make dependabot ignore it? That sounds so simple as I was looking in another direction to solve this ([use "ignore" option](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#ignore)).

You certainly can configure dependabot to ignore the whole action, but that would also prevent it from bumping the pinned clippy toolchain version.

Dependabot doesn't parse the toolchain config, only the action version, so this should work fine (though it's untested).

I can probably test it on a fork to confirm it works the way I expect

Philippe-Cholet:
Ok that seems good. Another thing that bothered me in #758 was that it suggested to update the version to "1.80.0" while I thought it would have suggested "1.74.0" (or whatever appropriate version at the time it wrote it).

danieleades:
> Ok that seems good. Another thing that bothered me in #758 was that it suggested to update the version to "1.80.0" while I thought it would have suggested "1.74.0" (or whatever appropriate version at the time it wrote it).

I missed that, that could be a problem.
Dependabot here is not doing anything clever to check the latest stable version, it's just checking the versions of the GitHub action that have been published. Perhaps these higher numbers are aliases for nightly releases?

Philippe-Cholet:
This is [intentional](https://github.com/dtolnay/rust-toolchain/issues/43) but it's convenient ahead-of-time PR.
I'm having a hard time finding a Rust repo pinning clippy (and using Dependabot).

mightyiam:
For my taste, this is too many things in one PR.

danieleades:
> I'm having a hard time finding a Rust repo pinning clippy (and using Dependabot).

pinning the version of linters is a general pattern, not limited to the Rust ecosystem

danieleades:
> For my taste, this is too many things in one PR.

```yml
      - uses: dtolnay/rust-toolchain@master
        with:
          toolchain: stable
          components: rustfmt
```

to

```yml
      - uses: dtolnay/rust-toolchain@stable
        with:
          components: rustfmt
```

is not strictly required, but trivial. All other changes are required.

danieleades:
> This is https://github.com/dtolnay/rust-toolchain/issues/43 but it's convenient ahead-of-time PR.

i think this might be deal-breaker unfortunately

Philippe-Cholet:
> i think this might be deal-breaker unfortunately

I agree. I think we are gonna stick with stable clippy and handle rare breakage as soon as it appears.

Philippe-Cholet:
Thank you anyway.

PR #847, opened 2024-01-13
add test coverage workflow

danieleades:


Philippe-Cholet:
After reading the [readme](https://crates.io/crates/cargo-llvm-cov), I learnt that nightly is required to support doctests (and without it, coverage would probably be quite lacking for non-adaptor `Itertools` methods, as I found out when experimenting with tarpaulin a while back).
Looks good otherwise.

danieleades:
> After reading the [readme](https://crates.io/crates/cargo-llvm-cov), I learnt that nightly is required to support doctests (and without it, coverage would probably be quite lacking for non-adaptor `Itertools` methods, as I found out when experimenting with tarpaulin a while back). Looks good otherwise.

i'm not sure what you mean? Nightly is required for the doctests coverage, which is why that's the toolchain used in the github action. Is that an issue?

Philippe-Cholet:
Not an issue but an observation that I found useful to share, that's all.
Looks good ~~otherwise~~.

jswrenn (commented):
Shouldn't the triggers for this be basically identical to our triggers for CI testing? https://github.com/rust-itertools/itertools/blob/f60fe7e9548ffb047420c3d9ec576aa792d8e9ca/.github/workflows/ci.yml#L3-L9

danieleades:
> Shouldn't the triggers for this be basically identical to our triggers for CI testing?
> 
> https://github.com/rust-itertools/itertools/blob/f60fe7e9548ffb047420c3d9ec576aa792d8e9ca/.github/workflows/ci.yml#L3-L9

sure, i can update with this. it will also need to run on main though, so would be something like (i guess)-

```yml
 on: 
   pull_request: 
     paths-ignore: 
       - "**.md" 
  push: 
    branches: [ master ]
     paths-ignore: 
       - "**.md" 
   merge_group: 
     paths-ignore: 
       - "**.md" 
 ```

that look right?

danieleades:
i've updated based on the discussion in this issue - https://github.com/matplotlib/napari-matplotlib/issues/155

Philippe-Cholet:
Since `merge_group` is not a trigger anymore (not in `on:` block), I am not sure `if: github.event_name != 'merge_group'` is useful but it sure is not a problem to have it.
Seems good to me.

danieleades:
> Since `merge_group` is not a trigger anymore (not in `on:` block), I am not sure `if: github.event_name != 'merge_group'` is useful but it sure is not a problem to have it. Seems good to me.

i'll happily remove it if it's redundant. I don't know have any experience with github merge queues

jswrenn (approved):
LGTM! If, for some reason, this isn't quite the right combination of triggers, we can revise in a follow-up PR. 

Thank you so much for this!

Philippe-Cholet:
https://github.com/rust-itertools/itertools/pull/835#issuecomment-1898863835 says ["Missing Base Commit"](https://docs.codecov.com/docs/error-reference#section-missing-head-commit). Maybe that merge this did not trigger an initial coverage report?

danieleades:
> [#835 (comment)](https://github.com/rust-itertools/itertools/pull/835#issuecomment-1898863835) says ["Missing Base Commit"](https://docs.codecov.com/docs/error-reference#section-missing-head-commit). Maybe that merge this did not trigger an initial coverage report?

looks like this coverage job ran on master but failed to upload - https://github.com/rust-itertools/itertools/actions/runs/7573294610/job/20625097590

Philippe-Cholet:
I searched a bit and found the same thing: https://github.com/codecov/codecov-action/issues/557#issuecomment-1224970469

danieleades:
the codecov token secret will need to be set up by a maintainer

Philippe-Cholet:
For the secret, I'm pretty sure an owner would be required, which means **jswrenn** (or **bluss**).
As a temporary fix, I triggered the workflow and got success: https://app.codecov.io/github/rust-itertools/itertools/tree/master/src

PR #774, opened 2023-10-04
Implement custom fold for zip longest

Owen-CH-Leung:
This PR is a continuation of the previous PR [here](https://github.com/rust-itertools/itertools/pull/764)

PR #849, opened 2024-01-13
Specialize `Interleave[Shortest]::fold`

Philippe-Cholet:
Related to #755

    cargo bench --bench specializations "interleave(_shortest)?/fold"

    interleave/fold            [2.1749 µs 2.1782 µs 2.1817 µs]
                               [754.64 ns 757.16 ns 760.51 ns]
                               [-66.092% -65.545% -65.107%]

    interleave_shortest/fold   [1.6414 µs 1.6460 µs 1.6511 µs]
                               [412.05 ns 413.60 ns 415.41 ns]
                               [-74.904% -74.580% -74.173%]

### DIFF
--- a/src/adaptors/mod.rs
+++ b/src/adaptors/mod.rs
@@ -195,6 +195,34 @@ where
         };
         (lower, upper)
     }
+
+    fn fold<B, F>(self, mut init: B, mut f: F) -> B
+    where
+        F: FnMut(B, Self::Item) -> B,
+    {
+        let Self {
+            mut it0,
+            mut it1,
+            phase,
+        } = self;
+        if phase {
+            match it1.next() {
+                Some(y) => init = f(init, y),
+                None => return init,
+            }
+        }
+        let res = it0.try_fold(init, |mut acc, x| {
+            acc = f(acc, x);
+            match it1.next() {
+                Some(y) => Ok(f(acc, y)),
+                None => Err(acc),
+            }
+        });
+        match res {
+            Ok(val) => val,
+            Err(val) => val,
+        }
+    }
 }
 
 impl<I, J> FusedIterator for InterleaveShortest<I, J>
--- a/src/adaptors/mod.rs
+++ b/src/adaptors/mod.rs
@@ -76,6 +76,30 @@ where
     fn size_hint(&self) -> (usize, Option<usize>) {
         size_hint::add(self.a.size_hint(), self.b.size_hint())
     }
+
+    fn fold<B, F>(self, mut init: B, mut f: F) -> B
+    where
+        F: FnMut(B, Self::Item) -> B,
+    {
+        let Self { mut a, mut b, flag } = self;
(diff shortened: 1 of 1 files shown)


src/adaptors/mod.rs
```diff
@@ -76,6 +76,28 @@ where
...
+        if flag {
+            if let Some(y) = b.next() {
+                init = f(init, y);
+            } else {
+                return a.fold(init, f);
+            }
+        }
+        init = a.fold(init, |mut acc, x| {
+            acc = f(acc, x);
+            if let Some(y) = b.next() {
```
phimuemue, line 94: I see we must check if ‘b.next‘ returns something. But if it returns None, couldn’t we exploit this by using try_fold and early-out, leaving the rest to a simple a.fold?

I could imagine that this might be better when b has very few elements compared to a, and I would hope that the compiler is able to generate the early-out without additional checks (it needs to check for None, anyway).

Philippe-Cholet: This makes me think of one the first specializations done in this large streak: https://github.com/rust-itertools/itertools/pull/774/files#diff-5423381b620ccb5a8ab55224d730c8729d676199d27701ac61043cd6773e4617R63-R66
There are probably some others that could possibly benefit from this idea (anytime there are two iterators I guess).

I'm not sure benchmarks would show improvements but it's definitely worth considering.

src/adaptors/mod.rs
```diff
@@ -171,6 +193,35 @@ where
...
+        if phase {
+            match it1.next() {
+                Some(y) => init = f(init, y),
+                None => return init,
+            }
+        }
+        let res = it0.try_fold(init, |mut acc, x| {
+            acc = f(acc, x);
+            match it1.next() {
+                Some(y) => acc = f(acc, y),
```
phimuemue, line 215: Shouldn’t this simply return Ok(f(acc, x))?

phimuemue (approved):
As always, LGTM.

Note: We should clean up these adaptors.

flag/phase should probably have the same name (maybe next_is_from_it1 or similar), and we may want to fuse the iterators also for the shortest-variant, or am I missing something?

src/adaptors/mod.rs
Philippe-Cholet: I noticed this as well a few minutes ago.

src/adaptors/mod.rs
Philippe-Cholet: Benchmark is now -65% instead of -75% initially but I pushed it as it seems to be a better decision for the general case.

Philippe-Cholet:
They could surely be a bit more similar (in another PR) at least the names. I'm pretty sure that fuse the iterators in the shortest-variant would be a breaking change for unfused iterators though.

jswrenn (approved):

phimuemue (approved):

PR #853, opened 2024-01-16
`ConsTuples`: remove `DoubleEndedIterator` implementation

### LINKED ISSUE #852: `ConsTuples::next_back` is plain wrong
I was re-considering specialize `ConsTuples::rfold` for #755 when I saw something obviously wrong:
https://github.com/rust-itertools/itertools/blob/f60fe7e9548ffb047420c3d9ec576aa792d8e9ca/src/cons_tuples_impl.rs#L26-L32

**`self.iter.next()` instead of `self.iter.next_back()` is one obvious bug here!**

`ConsTuples` is an internal detail of the `iproduct` macro and the cartesian product does not implement `DoubleEndedIterator` (and it can't as currently defined) so this is not used inside the crate (and won't).
No one noticed (in 8 years) such an obvious error so it's very unlikely used outside of itertools, right?

I suggest we _remove the implementation_ (breaking change?).
Or we could fix it, add a test (and specialize `rfold`).

Philippe-Cholet:
Fixes #852

### DIFF
--- a/src/cons_tuples_impl.rs
+++ b/src/cons_tuples_impl.rs
@@ -21,16 +21,6 @@ macro_rules! impl_cons_iter(
                 self.iter.fold(accum, move |acc, (($($B,)*), x)| f(acc, ($($B,)* x, )))
             }
         }
-
-        #[allow(non_snake_case)]
-        impl<X, Iter, $($B),*> DoubleEndedIterator for ConsTuples<Iter, (($($B,)*), X)>
-            where Iter: DoubleEndedIterator<Item = (($($B,)*), X)>,
-        {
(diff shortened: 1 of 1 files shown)


phimuemue:
@jswrenn  could you merge this (or veto if you think we should fix it)?

jswrenn (commented):
I think we should consider the removal of a trait impl to be a breaking change, even if it's the case that that impl did not behave correctly. There may be consumers who are happy with the buggy behavior, and who would be severely inconvenienced if they could no longer satisfy a `DoubleEndedIterator` trait bound in generic code.

jswrenn (approved):

PR #854, opened 2024-01-17
`ZipLongest::fold`: use `try_fold`?

Philippe-Cholet:
Based on https://github.com/rust-itertools/itertools/pull/849#discussion_r1453662745, `ZipLongest::fold` seems to be the only iterator that might benefit from this idea.

Benchmark specializations on slices is now +13% (still better than no `fold` specialization) but it might perform better in a more general case.

### DIFF
--- a/src/zip_longest.rs
+++ b/src/zip_longest.rs
@@ -54,17 +54,20 @@ where
     }
 
     #[inline]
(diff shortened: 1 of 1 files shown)


phimuemue (approved):
Thanks.

PR #856, opened 2024-01-19
allow small variation in coverage without returning a failure

danieleades:


Philippe-Cholet (commented):
The content of the file seems good to me.
I've seen use of `codecov.yml` but yours is `.codecov.yml`, is it sensible?

Philippe-Cholet (approved):

Philippe-Cholet:
For future readers, it's the same as https://github.com/PyO3/pyo3/pull/1142.

PR #857, opened 2024-01-20
Uniform field names for Interleave[Shortest]

phimuemue:
The two structs are similar, and this PR reflects this fact in their field names.


Philippe-Cholet (approved):
I see you applied your idea from https://github.com/rust-itertools/itertools/pull/849#pullrequestreview-1824116441, thanks.

PR #863, opened 2024-01-25
Prefer boxed slices over vectors internally

Philippe-Cholet:
While reading [This Week in Rust 531](https://this-week-in-rust.org/blog/2024/01/24/this-week-in-rust-531/) and specifically [Identifying Rust's `collect::<Vec<_>>()` memory leak footgun](https://blog.polybdenum.com/2024/01/17/identifying-the-collect-vec-memory-leak-footgun.html), I got the remainder that

> The proper data structure for fixed-length lists is `Box<[T]>` rather than `Vec<T>`.

So I went through our codebase and found three places to apply this: two are committed and `MultiProduct`. I thought I would wait #835 to be merged to do it there too but I reached one [limitation](https://doc.rust-lang.org/edition-guide/rust-2021/IntoIterator-for-arrays.html) of edition 2018 (in `last` specialization) so I'm not.

The other vectors we internally use do not have a fixed length. Such as `Combinations` that has a growing vector in order for `Powerset` to work too.
(_And I obviously do not want to break our iterators generating vectors when it could be boxed slices._)

### DIFF
--- a/src/combinations_with_replacement.rs
+++ b/src/combinations_with_replacement.rs
@@ -1,3 +1,4 @@
+use alloc::boxed::Box;
 use alloc::vec::Vec;
 use std::fmt;
 use std::iter::FusedIterator;
@@ -16,7 +17,7 @@ where
     I: Iterator,
     I::Item: Clone,
 {
-    indices: Vec<usize>,
+    indices: Box<[usize]>,
     pool: LazyBuffer<I>,
     first: bool,
 }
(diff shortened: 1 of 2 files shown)


phimuemue (approved):
Thanks, good catch. I think we should do this and trust that the additional `into_boxed_slice` does not need to reallocate, and, thus, can be optimized to a no-op.

Philippe-Cholet:
@phimuemue I noted that collect to `Box<[_]>` does `.collect::<Vec<_>>().into_boxed_slice()` in the std so if it reallocates then the std thinks it's the way.

`into_boxed_slice` does `shrink_to_fit` so I'm not sure I understand your "space consumption problem".

phimuemue:
> `into_boxed_slice` does `shrink_to_fit` so I'm not sure I understand your "space consumption problem".

I was thinking about the cases where we keep the `Vec`: Should this `shrink_to_fit` then? (We can also postpone this.)

Philippe-Cholet:
@phimuemue I think it has sense when we know it won't ever grow and we don't repeatedly shrink it. There is not that many cases in itertools: `kmerge_by` definition only I think.

PR #870, opened 2024-02-01
`iproduct`: generate tuples for simpler cases too

### LINKED ISSUE #868: iproduct rejects trailing comma
To reproduce:

```
use itertools::iproduct;

fn main() {
    for e in iproduct!(
        [1,2],
        [3,4],
    ) {
        println!("{e:?}");
    }
}
```

Expected output: it works (printing 4 lines of output).

Actual output:

```
error: unexpected end of macro invocation
   --> src/main.rs:6:15
    |
6   |         [3,4],
    |               ^ missing tokens in macro arguments
    |
note: while trying to match meta-variable `$K:expr`
   --> /playground/.cargo/registry/src/index.crates.io-6f17d22bba15001f/itertools-0.12.0/src/lib.rs:259:26
    |
259 |     ($I:expr, $J:expr, $($K:expr),+) => (
    |                          ^^^^^^^
```

`chain!` and `izip!` are fine.

Removing the trailing comma works:
```
use itertools::iproduct;

fn main() {
    for e in iproduct!(
        [1,2],
        [3,4]
    ) {
        println!("{e:?}");
    }
}
```

### LINKED ISSUE #869: iproduct should allow empty product (yielding unit once)
To reproduce:

```
use itertools::iproduct;

fn main() {
    for e in iproduct!(
    ) {
        println!("{e:?}");
    }
}
```

Expected output:
```
()
```

Actual output:
```
error: unexpected end of macro invocation
   --> src/main.rs:4:14
    |
4   |       for e in iproduct!(
    |  ______________^
5   | |     ) {
    | |_____^ missing tokens in macro arguments
    |
note: while trying to match `@`
   --> /playground/.cargo/registry/src/index.crates.io-6f17d22bba15001f/itertools-0.12.0/src/lib.rs:247:6
    |
247 |     (@flatten $I:expr,) => (
    |      ^
```

This wouldn't be particularly useful for humans, but macros that generate iproduct! invocations might produce this case sometimes.

Philippe-Cholet:
Fixes #868
Allow trailing commas is convenient when the macro is on multiple lines.

Fixes #869
It is more consistent to return tuples in every case.
It however is a breaking change.
Note the new case: the empty product generating a single unit tuple `()`, similar to the recent #835.

What do you think?
EDIT: 2nd option, we can fix the first one and close the second.

PS: I note that semver-checks does not see much breaking changes lately.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -250,10 +250,13 @@ macro_rules! iproduct {
         $crate::iproduct!(@flatten $crate::cons_tuples($crate::iproduct!($I, $J)), $($K,)*)
     );
     ($I:expr $(,)?) => (
-        $crate::__std_iter::IntoIterator::into_iter($I)
+        $crate::__std_iter::IntoIterator::into_iter($I).map(|elt| (elt,))
     );
     ($I:expr, $J:expr $(,)?) => (
-        $crate::Itertools::cartesian_product($crate::iproduct!($I), $crate::iproduct!($J))
+        $crate::Itertools::cartesian_product(
+            $crate::__std_iter::IntoIterator::into_iter($I),
+            $crate::__std_iter::IntoIterator::into_iter($J),
+        )
     );
     ($I:expr, $J:expr, $($K:expr),+ $(,)?) => (
         $crate::iproduct!(@flatten $crate::iproduct!($I, $J), $($K,)+)
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -249,6 +249,9 @@ macro_rules! iproduct {
     (@flatten $I:expr, $J:expr, $($K:expr,)*) => (
         $crate::iproduct!(@flatten $crate::cons_tuples($crate::iproduct!($I, $J)), $($K,)*)
     );
+    () => (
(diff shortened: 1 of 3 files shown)


jswrenn (approved):
I'm a little hesitant to make this breaking change. I appreciate the argument for consistency, but `(T,)` is a somewhat exotic type — unary tuples are basically never used in Rust. However, I concede that this *might* be useful in tandem with code that abstracts over tuple arity, e.g., with `frunk` or in the invocation of another macro. I also don't think we need to wait for someone to run into this particular edge case, since it's not particularly likely they'll also report it. Given this, I'm okay with approving this PR.

ijackson:
>  `(T,)` is a somewhat exotic type 

Indeed.  But `iproduct!( just_one_iterator )` is weird too.  In practice people won't write such code by hand very often.  Even when they do (eg when messing about trying to decide between code patterns, or commenting bits out, or something) providing the unary tuple will probably be more convenient as it'll involve less editing of the code that handles the output elements.

> Given this, I'm okay with approving this PR.

Thanks.

PR #871, opened 2024-02-07
Deprecate `unfold` for `std::iter::from_fn`

Philippe-Cholet:
Solves #223. Based on https://github.com/rust-itertools/itertools/pull/610#issuecomment-1145341531 that suggested to add a deprecation notice instead of removing it directly.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "itertools"
-version = "0.12.1"
+version = "0.13.0"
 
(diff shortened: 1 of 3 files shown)


Philippe-Cholet:
Ah "semver checks" is alive. I guess I should update the version to "0.13.0" in "README.md" and "Cargo.toml"?

jswrenn (approved):
Thanks! You've been absolutely killing it with triage lately, and the effort is much appreciated!

Issue #533, opened 2021-03-14
InterleaveShortest<I, J> is not actually fused


nox:
I was looking at the code and it turns out that `InterleaveShortest<I, J>` will resume iterating after a `None` if one of its iterators is not fused.

```rust
use itertools::Itertools;

fn not_fused() -> impl Iterator<Item = bool> {
    let mut last = Some(true);
    std::iter::from_fn(move || {
        match last {
            Some(b) => {
                last = None;
                Some(b)
            },
            None => {
                last = Some(true);
                None
            }
        }
    })
}

fn demonstrate_that_not_fused_is_indeed_not_fused() {
    let mut not_fused = not_fused();
    assert_eq!(not_fused.next(), Some(true));
    assert_eq!(not_fused.next(), None);
    assert_eq!(not_fused.next(), Some(true));
}

fn main() {
    demonstrate_that_not_fused_is_indeed_not_fused();

    let mut interleave_shortest = std::iter::repeat(false).interleave_shortest(not_fused());
    assert_eq!(interleave_shortest.next(), Some(false));
    assert_eq!(interleave_shortest.next(), Some(true));
    assert_eq!(interleave_shortest.next(), Some(false));
    assert_eq!(interleave_shortest.next(), None);

    // This fails, because `InterleaveShortest<I, J>` is not actually fused.
    assert_eq!(interleave_shortest.next(), None);
}
```

phimuemue:
This was apparently an intentional decision: https://github.com/rust-itertools/itertools/commit/beaf3955b51f5a5a2c5861d34ebd591736377b84

Not sure what to make of this... We should probably at least update documentation, and check if we behave consistently across itertools resp. the `std::iter` landscape.

jswrenn:
`InterleaveShortest` doesn't implement `FusedIterator`, so I wouldn't have expected it to be fused.

But, since it _already_ behaves as a fused iterator if its component iterators are fused, I imagine we could just add this impl:
```rust
impl<I, J> FusedIterator for InterleaveShortest<I, J>
where
    I: FusedIterator,
    J: FusedIterator
{}

jswrenn:
Ah, I see, the documentation prose claims it's fused.

Huh, I'm not sure what to make of this either.

nox:
IMO none of the iterators should be using fused iterators internally, but I concede that would be a breaking change. Why is `Interleave<I, J>` using `Fuse` internally for example?

bluss:
[...]

PR #872, opened 2024-02-08
Fuse iterators in InterleaveShortest

phimuemue:
In my book, `InterleaveShortest` should not produce more items than `Interleave`, but so far it did for unfused source iterators.

This PR brings `InterleaveShortest` in line with Interleave. The documentation already (incorrectly) stated that `InterleaveShortest` is fused.

Related to #533.


Philippe-Cholet:
I started an investigation ("blame" in the web pages of our code is even powerful than I thought):
In #31 it was added and was actually fused.
Then the commit that is interesting us here is https://github.com/rust-itertools/itertools/commit/beaf3955b51f5a5a2c5861d34ebd591736377b84 (by bluss).

phimuemue:
> I started an investigation ("blame" in the web pages of our code is even more powerful than I thought (for each line, we can have navigate its history)): In #31 it was added and was actually fused. Then the commit that is interesting us here is [beaf395](https://github.com/rust-itertools/itertools/commit/beaf3955b51f5a5a2c5861d34ebd591736377b84) (by bluss, he missed one comment in this "unfuse" commit that you found).

I'm fine with *either* unfused *or* fused, but I'd vote for uniform behavior. Not sure if we have a general guideline regarding fusedness, though.

Philippe-Cholet:
> I'm fine with _either_ unfused _or_ fused, but I'd vote for uniform behavior. Not sure if we have a general guideline regarding fusedness, though.

I agree.
I think it would be great to have a clear guideline about fusedness. And maybe better tests about (un)fusedness. I'm not sure all the iterators behave correctly in that regard (and have correct documentation).
It's not very high on my TODO list though as I'd rather triage/complete old issues/PRs and work on `fold` specializations and `maybe_/Try` to eventually move to use const-generics.

Philippe-Cholet:
I'm gonna make an issue about fusedness to see opened issues and PRs related to it, clarify the situation (like make a guideline) and eventually make a TODO list about all this.

PR #873, opened 2024-02-09
Specialize `MapForGrouping::fold`

Philippe-Cholet:
Related to #755

`MapForGrouping` is only internally used by `GroupingMapBy` and it only calls `for_each` on it (which in turn rely on `fold`). So I allow the clippy lint `missing_trait_methods` because specialize other methods is simply useless. (In case we run clippy with warning it.)

_I could replace `next` body by `unreachable!()` and it would still work fine._
Anyway, it will disappear from test coverage now that we have one ([see here](https://app.codecov.io/gh/rust-itertools/itertools/pull/873/blob/src/grouping_map.rs#L34)).

I previously wandered how to test and benchmark this:
- The current tests will test this.
- And I guess I could benchmark this `fold` specialization through some `into_grouping_map_by` benchmark but I simply don't think it's worth the time: no doubt `adapted_iterator.map.fold` is faster than the default `fold` calling `next` repeatedly. We just don't know how much faster.

Note that all `GroupingMapBy` methods ultimately rely on this `fold` so this specialization will improve performance for all its methods.

### DIFF
--- a/src/grouping_map.rs
+++ b/src/grouping_map.rs
@@ -22,6 +22,7 @@ impl<I, F> MapForGrouping<I, F> {
     }
 }
 
+#[allow(clippy::missing_trait_methods)]
 impl<K, V, I, F> Iterator for MapForGrouping<I, F>
 where
     I: Iterator<Item = V>,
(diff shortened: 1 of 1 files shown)


phimuemue (approved):
Thank you.

OT: It seems that `MapForGrouping` is (essentially) just a special cased [Map](https://doc.rust-lang.org/std/iter/struct.Map.html). IIRC we have some of these special-cased Maps/Folds (e.g. `MapOk`, `FilterOk`, `FilterMapOk`). I wonder if we could - once MSRV reaches 1.75 - [replace them by `Filter<InnerIterator, impl FnMut(InnerIterator::Item)->V>`](https://blog.rust-lang.org/2023/12/21/async-fn-rpit-in-traits.html#whats-stabilizing) or the like. Then, they would automatically inherit the specialized methods.

phimuemue:
After a second thought, I'm pretty sure `type MapForGrouping<I, F> = MapSpecialCase<I, MapForGroupingFn<F>>` would be cleaner.

Philippe-Cholet:
Seems cleaner indeed.

PR #874, opened 2024-02-10
update dependabot config

danieleades:
adds dependabot config for bumping cargo dependencies

this should generate PRs for criterion, quickcheck, and rand.

the quickcheck PR will fail CI since there are breaking changes, likely blocked on https://github.com/BurntSushi/quickcheck/issues/267

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,7 @@ updates:
   directory: "/"
   schedule:
     interval: daily
+- package-ecosystem: cargo
+  directory: "/"
+  schedule:
+    interval: daily


Philippe-Cholet (approved):

jswrenn:
What's the advantage of actively bumping dev-dependencies? Why shouldn't we specify the minimum version we actually need, and let the cargo resolver do its job?

Philippe-Cholet:
Just read our "Cargo.toml" and saw **either** is "1.0" when the latest version is "1.10.0". Maybe we should update it to "1" or "1.10" as dependency trees might have it in multiple versions.
About dev-dependencies, maybe it's not that important though?

jswrenn:
> Maybe we should update it to "1" or "1.10" as dependency trees might have it in multiple versions.

`either = "1"` is a shorthand for "either >= "1". Cargo will only resolve a single version per major-version train, so there's no risk here of applications getting compiled with both "1" and "1.10".

jswrenn:
See https://github.com/tokio-rs/tokio/pull/6335#issuecomment-1937061448, which sets a similar policy for tokio-rs. Unless we have good reason to change our current policy, I'd like to stick with it.

Philippe-Cholet:
There is nothing urgent about updating dev-dependencies.
And while I think dependabot could _in general_ be helpful about semver-major updates for our dependencies, we don't expect **either** (our only dependency) to ever go to "2.0".
And `either = "1.0"` actually means `either = ">=1.0.0, <2.0.0"` (cf [link](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html)) which I forgot.

Philippe-Cholet:
Your help on improving CI is very much appreciated, thanks!
However, it seems that this (and #875) should be closed. I'd promptly reopen if needed.

PR #876, opened 2024-02-12
type MapForGrouping<...> = MapSpecialCase<...>

phimuemue:
Follow-up to #873

### DIFF
--- a/src/grouping_map.rs
+++ b/src/grouping_map.rs
@@ -1,45 +1,33 @@
 #![cfg(feature = "use_std")]
 
-use crate::MinMaxResult;
+use crate::{
+    adaptors::map::{MapSpecialCase, MapSpecialCaseFn},
+    MinMaxResult,
+};
 use std::cmp::Ordering;
 use std::collections::HashMap;
-use std::fmt;
 use std::hash::Hash;
 use std::iter::Iterator;
 use std::ops::{Add, Mul};
 
 /// A wrapper to allow for an easy [`into_grouping_map_by`](crate::Itertools::into_grouping_map_by)
-#[derive(Clone)]
-pub struct MapForGrouping<I, F>(I, F);
+pub type MapForGrouping<I, F> = MapSpecialCase<I, GroupingMapFn<F>>;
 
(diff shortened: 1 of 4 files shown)


src/grouping_map.rs
```diff
@@ -1,45 +1,33 @@
...
 
 /// A wrapper to allow for an easy [`into_grouping_map_by`](crate::Itertools::into_grouping_map_by)
-#[derive(Clone)]
-pub struct MapForGrouping<I, F>(I, F);
+pub type MapForGrouping<I, F> = MapSpecialCase<I, GroupingMapFn<F>>;
 
-impl<I: fmt::Debug, F> fmt::Debug for MapForGrouping<I, F> {
-    debug_fmt_fields!(MapForGrouping, 0);
-}
+pub struct GroupingMapFn<F>(F);
```
Philippe-Cholet, line 16: It should implement `Debug` just like `MapSpecialCaseFnOk`.

src/grouping_map.rs
```diff
@@ -1,45 +1,33 @@
...
-    }
-
-    fn fold<B, G>(self, init: B, f: G) -> B
-    where
-        G: FnMut(B, Self::Item) -> B,
-    {
-        let mut key_mapper = self.1;
-        self.0.map(|val| (key_mapper(&val), val)).fold(init, f)
+impl<K, I: Iterator, F: FnMut(&I::Item) -> K> MapForGrouping<I, F> {
+    pub(crate) fn new(iter: I, key_mapper: F) -> Self {
```
Philippe-Cholet, line 26: It creates a method `new` for `MapSpecialCase` (since `MapForGrouping` aliases it). It's not in conflict with another today but maybe we should create a function `new_by` instead just like `new` below and therefore simplify `grouping_map::new(grouping_map::MapForGrouping::new(self, key_mapper))` to `grouping_map::new_by(self, key_mapper)` in "lib.rs".

Philippe-Cholet (changes requested):

src/grouping_map.rs
Philippe-Cholet: And derive Clone.

Philippe-Cholet (approved):

PR #879, opened 2024-02-18
Update crate keywords

Philippe-Cholet:
Now that `group_by` has been renamed to `chunk_by` in #866 I think we should just delete the crate keyword "group-by".
Note that on [crates.io](https://crates.io/keywords/group-by), this keyword only leads to "itertools" and "itertools-wild" so it does not seem useful at all.

_Should we delete another one as well?_


jswrenn (approved):
Sounds reasonable to me. I doubt "group-by" is even that common of a search term.

PR #884, opened 2024-02-26
Specialize TakeWhileRef fold

Owen-CH-Leung:
#755 

This PR introduces a custom fold method for `TakeWhileRef`. The optimization hinges on the `fold_while` logic:

Benchmark Results:

Default fold: time:   [53.963 ns 54.083 ns 54.209 ns]
Custom fold: time:   [53.210 ns 53.359 ns 53.511 ns]

### DIFF
--- a/src/adaptors/mod.rs
+++ b/src/adaptors/mod.rs
@@ -13,6 +13,8 @@ pub use self::map::{map_into, map_ok, MapInto, MapOk};
 pub use self::multi_product::*;
 
 use crate::size_hint::{self, SizeHint};
+use crate::FoldWhile::{Continue, Done};
+use crate::Itertools;
 use std::fmt;
 use std::iter::{Enumerate, FromIterator, Fuse, FusedIterator};
 use std::marker::PhantomData;
@@ -554,6 +556,30 @@ where
 
     fn size_hint(&self) -> (usize, Option<usize>) {
         (0, self.iter.size_hint().1)
+    }
+
+    fn fold<B, G>(mut self, init: B, mut g: G) -> B
+    where
+        G: FnMut(B, I::Item) -> B,
+    {
+        let acc = init;
+        let original_iter = self.iter.clone();
+
+        let result = self.iter.clone().fold_while(acc, |acc, x| {
+            if (self.f)(&x) {
+                Continue(g(acc, x))
+            } else {
+                Done(acc)
+            }
+        });
+
+        match result {
+            Done(acc) => {
+                *self.iter = original_iter;
+                acc
+            }
+            Continue(acc) => acc,
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


Philippe-Cholet:
@Owen-CH-Leung I'm just curious about why you closed this. I did not look into the details.

`TakeWhileRef` is a bit of a pain for me. Because of `&mut I`, it's not clonable so it's not tested yet in "tests/specializations.rs" and not benchmarked in "benches/specializations.rs".
Any improvement on that front would be appreciated.

Owen-CH-Leung:
> @Owen-CH-Leung I'm just curious about why you closed this. I did not look into the details.
> 
> `TakeWhileRef` is a bit of a pain for me. Because of `&mut I`, it's not clonable so it's not tested yet in "tests/specializations.rs" and not benchmarked in "benches/specializations.rs". Any improvement on that front would be appreciated.

Opps just wrong click. 

Yeah I was trying to use the `fold_while` implementation to implement fold for `TakeWhileRef`, but I think my changes will now leave the original iterator unchanged and thus it's failing one of the doc test `take_while_ref`. And I couldn't think of a way to keep the performance gain while also having the intended effect as illustrated in the `take_while_ref` doctest

src/adaptors/mod.rs
```diff
@@ -555,6 +557,30 @@ where
...
     }
+
+    fn fold<B, G>(mut self, init: B, mut g: G) -> B
+    where
+        G: FnMut(B, I::Item) -> B,
+    {
+        let acc = init;
+        let original_iter = self.iter.clone();
+
+        let result = self.iter.clone().fold_while(acc, |acc, x| {
```
Philippe-Cholet, line 568: Do not clone it a second time.

Owen-CH-Leung:
> `TakeWhileRef` is a bit of a pain for me. Because of `&mut I`, it's not clonable so it's not tested yet in "tests/specializations.rs" and not benchmarked in "benches/specializations.rs".
> Any improvement on that front would be appreciated.

I think the design `&mut I` is chosen such that the API `take_while_ref` can directly mutate the iterator without transferring the ownership or cloning the iterator. This will make the iterator more performant I believe. Are we open to changing the API such as making it take over the ownership of the iterator / creating interior mutability ?

Philippe-Cholet:
I don't think we are open of changing the API (see #710). Plus, as I did not play much with `take_while_ref`, I'm not sure to fully understand its purpose.

About testing, we currently have a `fn test_specializations<I>(it: &I) where ..., I: Clone;` but I think we could have a similar `fn test_specializations_no_iter_clone<I>(get_it: impl Fn() -> I) where ...;` where the function `get_it` could clone or do something else for unclonable iterators but that's only an idea for now: I tried applying it a bit but it's definitely not straightforward.

PR #885, opened 2024-02-27
`k_smallest` variants (2)

Philippe-Cholet:
Follow-up of the recent #654.

I wanted to only update the documentation as I did not want to burden **ejmount** with it.
But I missed a few things:
- Function bounds: we usually use `FnMut` and not `Fn`.
- `izip` in the test could truncate the result, replaced by our `assert_equal`.
- add a "generic test" similar to an existing one.

### DIFF
--- a/tests/test_std.rs
+++ b/tests/test_std.rs
@@ -508,34 +508,24 @@ qc::quickcheck! {
         let num_elements = min(k, m as _);
 
         // Compute the top and bottom k in various combinations
+        let sorted_smallest = sorted[..num_elements].iter().cloned();
         let smallest = v.iter().cloned().k_smallest(k);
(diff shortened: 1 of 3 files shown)


tests/test_std.rs
```diff
@@ -508,34 +508,24 @@ qc::quickcheck! {
...
-            smallest,
-            smallest_by,
-            smallest_by_key) {
-            assert_eq!(a,b);
-            assert_eq!(a,c);
-            assert_eq!(a,d);
-        }
+        it::assert_equal(smallest, sorted_smallest.clone());
+        it::assert_equal(smallest_by, sorted_smallest.clone());
+        it::assert_equal(smallest_by_key, sorted_smallest);
```
phimuemue, line 524: Good one.

phimuemue (approved):
Thanks.

PR #886, opened 2024-02-27
Loosen `Fn` bounds to `FnMut`

Philippe-Cholet:
In #885, I relaxed the `Fn` bounds to `FnMut` for the new variants of `k_smallest`.
It lead me to find out this is also possible for `diff_with` and `Itertools::into_group_map_by`.

I'm not sure why we would need `Fn` bounds in our crate.
clippy does not have a `disallowed_traits` lint or I would have added `Fn` to it.


jswrenn (approved):
Well spotted!

PR #887, opened 2024-02-28
Document the `use_alloc` feature

Philippe-Cholet:
I always found weird that this feature was never documented.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -37,7 +37,11 @@
 //! - `use_std`
 //!   - Enabled by default.
 //!   - Disable to compile itertools using `#![no_std]`. This disables
-//!     any items that depend on collections (like `chunk_by`, `unique`,
+//!     any item that depend on allocations (see the `use_alloc` feature)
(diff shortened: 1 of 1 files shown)


jswrenn (approved):

src/lib.rs
```diff
@@ -37,7 +37,11 @@
 //! - `use_std`
 //!   - Enabled by default.
 //!   - Disable to compile itertools using `#![no_std]`. This disables
-//!     any items that depend on collections (like `chunk_by`, `unique`,
+//!     any item that depend on allocations (see the `use_alloc` feature)
+//!     and hashes (like `unique`, `counts`, `into_grouping_map` and more).
```
jswrenn, line 41:
```suggestion
//!     and hash maps (like `unique`, `counts`, `into_grouping_map` and more).
```

jswrenn (changes requested):

src/lib.rs
Philippe-Cholet: There is also "hash sets" so I chose "hashes" to unify those two terms.

jswrenn (approved):

PR #888, opened 2024-02-28
`set_from`: use `zip`

Philippe-Cholet:
Use `try_fold` instead of a "breakable for-loop" as we usually do.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2218,15 +2218,10 @@ pub trait Itertools: Iterator {
         Self: Iterator<Item = &'a mut A>,
         J: IntoIterator<Item = A>,
     {
-        let mut count = 0;
-        for elt in from {
-            match self.next() {
-                None => break,
-                Some(ptr) => *ptr = elt,
-            }
-            count += 1;
-        }
-        count
+        from.into_iter()
+            .zip(self)
+            .map(|(new, old)| *old = new)
+            .count()
     }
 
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -2218,15 +2218,15 @@ pub trait Itertools: Iterator {
...
-        count
+        from.into_iter()
+            .try_fold(0, |count, elt| match self.next() {
+                Some(ptr) => {
+                    *ptr = elt;
+                    Ok(count + 1)
+                }
+                None => Err(count),
+            })
+            .unwrap_or_else(|count| count)
```
jswrenn, line 2229: I think this could be more simply expressed as:
```suggestion
        self.zip(from).map(|(old, new)| {
            *old = new;
        }).count()
```

jswrenn (approved):

jswrenn (changes requested):
Whoops, the previous review shouldn't have been an approval.

src/lib.rs
Philippe-Cholet: We currently get `from` elements before `self` ones so zipping that way would get one too many element from `self` if `from` is shorter.
And `zip.map.count` would rely on `std::iter::Zip::fold` which is unspecialized (IDK why).
And I prefer a simple `try_fold` to this chain.

jswrenn (approved):

PR #767, opened 2023-09-28
Create `CONTRIBUTING.md`

Philippe-Cholet:
Fixes #756 

As mentioned in the issue, I did not write anything related to testing yet, as I think we should clarify what's expected of contributors.

[Rendered here](https://github.com/Philippe-Cholet/itertools/blob/contributing/CONTRIBUTING.md)

PR #890, opened 2024-02-28
Remove badges in README.md

Philippe-Cholet:
I might have been overzealous on the badges here but I think it's nice, especially the MSRV one that does not need manual updates.

<details>
<summary>Rendered badges</summary>

[Documentation](https://docs.rs/itertools) [crates.io](https://crates.io/crates/itertools) [MSRV](https://github.com/rust-itertools/itertools/blob/master/Cargo.toml) [License](https://github.com/rust-itertools/itertools) [Build Status](https://github.com/rust-itertools/itertools/actions) [Code Coverage](https://codecov.io/github/rust-itertools/itertools) [Total Downloads](https://crates.io/crates/itertools) [Contributors](https://github.com/rust-itertools/itertools/graphs/contributors)
</details>

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,9 +4,6 @@ Extra iterator adaptors, functions and macros.
 
 Please read the [API documentation here](https://docs.rs/itertools/).
 
-[![build_status](https://github.com/rust-itertools/itertools/actions/workflows/ci.yml/badge.svg)](https://github.com/rust-itertools/itertools/actions)
-[![crates.io](https://img.shields.io/crates/v/itertools.svg)](https://crates.io/crates/itertools)
-
 How to use with Cargo:
 
 ```toml


jswrenn (commented):
I might be alone on this, but I'm not a fan of badges in READMEs, and would even prefer to see the two badges we currently have eliminated for several reasons.

Repos, ideally, are self-contained snapshots of the state of the project. Care should be taken to minimize references to outside data, or to ensure that outside references are highly durable. Furthermore, the job of a README is to succinctly convey the essential information for using a project — not to recreate UI elements that are already presented elsewhere. We additionally need to be mindful that whatever badges we put in our README are also mirrored on crates.io, so they should make sense both for a user who's browsing the code on GitHub, and for a user who's evaluating the crate on crates.io.

Our two current badges fail to meet this bar:
- Whether CI is passing on the master branch is completely irrelevant to >99% users of this crate, who are instead grabbing releases from crates.io. It also doesn't maintain accuracy as you checkout older snapshots of the repo — it stubbornly always reflects what's *currently* happening on master.
- The badge indicating our current release number on crates.io is irrelevant to anyone finding the crate on crates.io. It duplicates information present at the top of our `Cargo.toml`. It doesn't maintain accuracy as you check out older snapshots of the repo.

This same line of critique applies to most or all of the badges this PR adds:
- MSRV: Our MSRV is a property of our codebase at particular points in time. It should be documented in a static way.
- License: Github already renders this above the README. Crates.io also renders it prominently. We document it statically in our README, in our Cargo.toml, and in our LICENSE files. We don't need to duplicate this UI element.
- Code Coverage: Same critique as build status.
- Total Downloads: This isn't a property of our code base. It's also presented prominently in crates.io. We don't need to duplicate this UI element in our readme.
- Total Contributors: Github already provides this information on the homepage of the repo. We don't need to duplicate this UI element in our readme.

Philippe-Cholet:
Well this is quite argumented and I can agree.

About [Documentation](https://docs.rs/itertools) (which replaces a sentence, same link), I guess we can say the end user does not care that the "docs are passing" as long as he has the link to it (and that it works, which is done in CI).

So remove the two badges instead?

Philippe-Cholet:
@jswrenn
**phimuemue** liked your comment and my reponse on the "docs" badge so I assume he agrees.
I therefore removed all badges instead.
For me, it was instructive about how to create badges, and why it might not be a good idea.
You can merge or close this as you see fit.

Philippe-Cholet:
@jswrenn Because this is markdown only, according to your https://github.com/rust-itertools/itertools/pull/767#issuecomment-1932290547, you have to bypass branch protections to merge this.

PR #895, opened 2024-03-07
Deprecate `tree_fold1` for `tree_reduce`

Philippe-Cholet:
I think `tree_fold1` should have been deprecated for `tree_reduce` at the same time `fold1` was deprecated for `reduce`.

With this deprecation, we will be able to remove `tree_fold1` and `fold1` at the same time and close that chapter once and for all.

@scottmcm Do you have an objection on this?

**PS:** I wish I had thought of this but no, thanks [reddit](https://www.reddit.com/r/rust/comments/1830wnj/tree_foldl_should_be_tree_reduce/).


Philippe-Cholet:
In `GroupingMap`, `fold_first` seems to be named after `fold1`, is it?
If that's the case, should it be deprecated for `reduce` similarly to what I did here with `tree_`?

jswrenn (approved):
This seems reasonable to me.

Philippe-Cholet:
I'm okay with both. @jswrenn Feel free to merge or ask me to change the name.

jswrenn:
Let's stick with `tree_reduce`. We don't have any other `reduce*` methods, so sorting doesn't matter too much for us. If the standard library decides to stabilize such a method, the libs team might want to revisit the name.

Issue #844, opened 2024-01-12
About fallible iterators


Philippe-Cholet:
I just created the label "fallible-iterator" and labelled (only opened) related issues and pull requests where the iterator item is fallible: results mostly, options too.
It's one recurring topic, and a proper discussion might be useful.

---

To start, I think our adaptors of iterators of results should be `_ok`-suffixed.
`try_`-prefixed should be reserved to methods tied to the currently unstable `Try` trait.

jswrenn:
I'd like to take the approach described in https://github.com/rust-itertools/itertools/pull/814#discussion_r1439725414; i.e.:
- We privately fork the `Try` trait and implement it for `Option` and `Result`.
- We name our methods with the prefix `maybe_`.
- When `Try` is stabilized, we (hopefully) simply delete our fork of `Try` and depend on the standard library one.

With this approach, we can give users a cutting-edge experience without increasing our MSRV. And, by using the `maybe_` prefix, we can also freely introduce new methods without worrying of conflicting with future additions to the standard library.

Philippe-Cholet:
From https://github.com/rust-itertools/itertools/pull/814#discussion_r1439867108
> In libcore, `fold` is usually specialized using `try_fold` specializations alongside with `NeverShortCircuit` which surely won't be public forever so I thought that we would copy it at some point.

so I think it's natural to copy part (here `Try` related parts) of the std as we simply want to expand it.

I would largely prefer `try_`-prefixed names over `maybe_` but I reckon that avoid conflict with libcore is more important, with RFC2845 implemented or not.

jswrenn:
Yes, avoiding conflict with libcore is important, and it's why I'm recommending using the `maybe_` prefix. Without RFC2845 implemented, the conflict would cause a compile error. With RFC2845 implemented, our implementations would shadow the standard library's, but almost certainly be sub-optimal in implementation (e.g., for any `try_` methods stabilized before the `Try` trait, we can't delegate to the stdlib's `try_` methods in our own implementations, since we couldn't replicate the stdlib's `Try` bound).

Philippe-Cholet:
@phimuemue What do you think on this matter?

---

_What would be the inconvenients of using Rust nightly_ (for one feature only)? As we would be able to use the `Try` trait and specialize `try_fold`. (We would still have to copy some parts such as `NeverShortCircuit` but not much).
It would require to have the nightly toolchain installed (I currently don't), which might "intimidate" new contributors.
[...]

PR #897, opened 2024-03-08
Removed lifetimes in "process_results" in favour of using NonNull

w-utter:
There were some lifetime issues that I was running into trying to directly raise an `impl Iterator<Item = Result<T, E>>` to a `Result<impl Iterator<Item = T>, E>` saying something along the lines that the lifetimes were incompatible. However when changing to NonNull these errors went away, and works the same way as before.

This will coincide with a wrapper fn that handles doing `iterator.into_iter().process_results(|iter| iter.map(|item| item))` which I will work on shortly.

Just as a note, In the future this api should probably take in all T who implement Try just so that this would be possible with options and other user generated objects, but that is out of the scope of this pr.

### DIFF
--- a/src/process_results_impl.rs
+++ b/src/process_results_impl.rs
@@ -1,5 +1,6 @@
 #[cfg(doc)]
 use crate::Itertools;
+use core::ptr::NonNull;
 
 /// An iterator that produces only the `T` values as long as the
 /// inner iterator produces `Ok(T)`.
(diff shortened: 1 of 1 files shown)


Philippe-Cholet:
In #710, I said that `process_results` "should" take `&mut self` instead of `self`, in which case the lifetime you remove here is useful I think. I'm gonna think about this.

And as an iterator of results, you might want to take a look at #844.

w-utter:
Ah, wrote this a little hastily while trying to get `collect_into::<Result<Vec<T>>()` and similar to work without thinking of the ramifications of removing the associated lifetime. Currently the std library requires the iterator being collected into to implement extend which i thought could be bypassed by using this kind of adapter. Looking at some other material I realize that there are some other ways of doing it. Thanks for the explanation!

PR #899, opened 2024-03-08
New `Itertools::tail`

Philippe-Cholet:
Hi, after 80 merged PRs fixing, modifying, testing, benchmarking, specializing or whatever really... I just checked this is my first PR where I actually add a "_brand new_" feature myself. 🤣

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3144,6 +3144,57 @@ pub trait Itertools: Iterator {
         self.k_largest_by(k, k_smallest::key_to_cmp(key))
     }
 
+    /// Consumes the iterator and return an iterator of the last `n` elements.
+    ///
+    /// The iterator, if directly collected to a `Vec`, is converted
+    /// without any extra copying or allocation cost.
+    ///
+    /// ```
+    /// use itertools::{assert_equal, Itertools};
+    ///
+    /// let v = vec![5, 9, 8, 4, 2, 12, 0];
+    /// assert_equal(v.iter().tail(3), &[2, 12, 0]);
+    /// assert_equal(v.iter().tail(10), &v);
+    ///
+    /// assert_equal(v.iter().tail(1), v.iter().last());
+    ///
+    /// assert_equal((0..100).tail(10), 90..100);
+    /// ```
+    ///
+    /// For double ended iterators without side-effects, you might prefer
+    /// `.rev().take(n).rev()` to have a similar result (lazy and non-allocating)
+    /// without consuming the entire iterator.
+    #[cfg(feature = "use_alloc")]
+    fn tail(self, n: usize) -> VecIntoIter<Self::Item>
+    where
+        Self: Sized,
+    {
+        match n {
+            0 => {
+                self.last();
+                Vec::new()
+            }
+            1 => self.last().into_iter().collect(),
+            _ => {
+                let mut iter = self.fuse();
+                let mut data: Vec<_> = iter.by_ref().take(n).collect();
+                // Update `data` cyclically.
+                let idx = iter.fold(0, |i, val| {
+                    data[i] = val;
+                    if i + 1 == n {
+                        0
+                    } else {
+                        i + 1
+                    }
+                });
+                // Respect the insertion order.
+                data.rotate_left(idx);
+                data
+            }
+        }
+        .into_iter()
+    }
+
     /// Collect all iterator elements into one of two
     /// partitions. Unlike [`Iterator::partition`], each partition may
     /// have a distinct type.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3159,6 +3159,8 @@ pub trait Itertools: Iterator {
     /// assert_equal(v.iter().tail(1), v.iter().last());
     ///
     /// assert_equal((0..100).tail(10), 90..100);
+    ///
+    /// assert_equal((0..100).filter(|x| x % 3 == 0).tail(10), (72..100).step_by(3));
     /// ```
     ///
     /// For double ended iterators without side-effects, you might prefer
@@ -3176,7 +3178,9 @@ pub trait Itertools: Iterator {
             }
             1 => self.last().into_iter().collect(),
             _ => {
-                let mut iter = self.fuse();
+                // Skip the starting part of the iterator if possible.
+                let (low, _) = self.size_hint();
+                let mut iter = self.fuse().skip(low.saturating_sub(n));
                 let mut data: Vec<_> = iter.by_ref().take(n).collect();
                 // Update `data` cyclically.
                 let idx = iter.fold(0, |i, val| {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3190,6 +3190,7 @@ pub trait Itertools: Iterator {
                 let mut data: Vec<_> = iter.by_ref().take(n).collect();
                 // Update `data` cyclically.
                 let idx = iter.fold(0, |i, val| {
+                    debug_assert_eq!(data.len(), n);
                     data[i] = val;
                     if i + 1 == n {
                         0
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -55,7 +55,7 @@ extern crate core as std;
 extern crate alloc;
 
 #[cfg(feature = "use_alloc")]
-use alloc::{string::String, vec::Vec};
+use alloc::{collections::VecDeque, string::String, vec::Vec};
 
 pub use either::Either;
 
@@ -71,6 +71,8 @@ use std::fmt::Write;
 #[cfg(feature = "use_std")]
 use std::hash::Hash;
 use std::iter::{once, IntoIterator};
+#[cfg(feature = "use_alloc")]
+type VecDequeIntoIter<T> = alloc::collections::vec_deque::IntoIter<T>;
 #[cfg(feature = "use_alloc")]
 type VecIntoIter<T> = alloc::vec::IntoIter<T>;
 use std::iter::FromIterator;
@@ -3146,8 +3148,10 @@ pub trait Itertools: Iterator {
 
     /// Consumes the iterator and return an iterator of the last `n` elements.
     ///
-    /// The iterator, if directly collected to a `Vec`, is converted
+    /// The iterator, if directly collected to a `VecDeque`, is converted
     /// without any extra copying or allocation cost.
+    /// If directly collected to a `Vec`, it may need some data movement
+    /// but no re-allocation.
     ///
     /// ```
     /// use itertools::{assert_equal, Itertools};
@@ -3167,20 +3171,22 @@ pub trait Itertools: Iterator {
     /// `.rev().take(n).rev()` to have a similar result (lazy and non-allocating)
     /// without consuming the entire iterator.
     #[cfg(feature = "use_alloc")]
-    fn tail(self, n: usize) -> VecIntoIter<Self::Item>
+    fn tail(self, n: usize) -> VecDequeIntoIter<Self::Item>
     where
         Self: Sized,
     {
         match n {
             0 => {
                 self.last();
-                Vec::new()
+                VecDeque::new()
             }
             1 => self.last().into_iter().collect(),
             _ => {
                 // Skip the starting part of the iterator if possible.
                 let (low, _) = self.size_hint();
                 let mut iter = self.fuse().skip(low.saturating_sub(n));
+                // TODO: If VecDeque has a more efficient method than
+                // `.pop_front();.push_back(val)` in the future then maybe revisit this.
                 let mut data: Vec<_> = iter.by_ref().take(n).collect();
                 // Update `data` cyclically.
                 let idx = iter.fold(0, |i, val| {
@@ -3191,7 +3197,8 @@ pub trait Itertools: Iterator {
                         i + 1
                     }
                 });
-                // Respect the insertion order.
+                // Respect the insertion order, efficiently.
+                let mut data = VecDeque::from(data);
                 data.rotate_left(idx);
                 data
             }
--- a/src/k_smallest.rs
+++ b/src/k_smallest.rs
@@ -57,6 +57,7 @@ where
     }
 
     iter.for_each(|val| {
+        debug_assert_eq!(storage.len(), k);
         if is_less_than(&val, &storage[0]) {
             // Treating this as an push-and-pop saves having to write a sift-up implementation.
             // https://en.wikipedia.org/wiki/Binary_heap#Insert_then_extract
--- a/tests/quick.rs
+++ b/tests/quick.rs
@@ -1949,4 +1949,9 @@ quickcheck! {
             result_set.is_empty()
         }
     }
+
+    fn tail(v: Vec<i32>, n: u8) -> bool {
+        let n = n as usize;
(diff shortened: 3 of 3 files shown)


Philippe-Cholet:
I wonder if _libcore_ would want a `Iterator::tail<const N: usize>` (not as straightforward as this implementation in case the iterator generate less than `N` items but it's possible) in which case name my method `Itertools::tail(n: usize)` would clash.

src/lib.rs
```diff
@@ -3133,6 +3133,52 @@ pub trait Itertools: Iterator {
...
+    ///
+    /// let v = vec![5, 9, 8, 4, 2, 12, 0];
+    /// assert_equal(v.iter().tail(3), &[2, 12, 0]);
+    /// assert_equal(v.iter().tail(10), &v);
+    ///
+    /// assert_equal((0..100).tail(10), 90..100);
+    /// ```
+    ///
+    /// For double ended iterators without side-effects, you might prefer
+    /// `.rev().take(n).collect_vec().into_iter().rev()`
```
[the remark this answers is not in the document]
Philippe-Cholet: The idea was to have a very similar type (but it's not exactly the same: `Rev<VecIntoIter<_>>`) but yes I probably go with something simpler. Something like:

> you might prefer `.rev().take(n).rev()` to have a similar result (lazy and non-allocating) without consuming the entire iterator.

src/lib.rs
```diff
@@ -3133,6 +3133,52 @@ pub trait Itertools: Iterator {
...
+    fn tail(self, n: usize) -> VecIntoIter<Self::Item>
+    where
+        Self: Sized,
+    {
+        match n {
+            0 => {
+                self.last();
+                Vec::new()
+            }
+            1 => self.last().into_iter().collect(),
```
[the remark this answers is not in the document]
Philippe-Cholet: Yes the idea was that `last` could sometimes be faster than `fold` (and sometimes optimized away, like I suppose it could if I was interating on a slice).
And avoid `% 0` in the general case.
Case `n = 1`, I came up with it later as I was thinking about optimizing it, I'm not entirely sure it's worth it.

src/lib.rs
```diff
@@ -3133,6 +3133,52 @@ pub trait Itertools: Iterator {
...
+            _ => {
+                let mut iter = self.fuse();
+                let mut data: Vec<_> = iter.by_ref().take(n).collect();
+                // Update `data` cyclically.
+                let idx = iter.fold(0, |i, val| {
+                    data[i] = val;
+                    (i + 1) % n
```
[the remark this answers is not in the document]
Philippe-Cholet: I was wondering about this and I'm glad you have more insight on this than me.
I guess I'm gonna make a small benchmark.

src/lib.rs
```diff
@@ -3133,6 +3133,52 @@ pub trait Itertools: Iterator {
...
+                });
+                // Respect the insertion order.
+                data.rotate_left(idx);
```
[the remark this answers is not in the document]
Philippe-Cholet: This rotate the slice only once for the entire iteration.
But unless I don't know something, `VecDeque` would rotate the data multiple times during a long iteration to make some place on the back by pushing data on the front.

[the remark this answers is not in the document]
Philippe-Cholet: The size of the allocation of a `VecDeque` necessarily is more than `n` to have some place to amortize the rotation (and it's not contiguous).
While my solution can allocate exactly `n` (in some cases) and the rotation is a simple clean-up at the end.

src/lib.rs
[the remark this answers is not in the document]
Philippe-Cholet: `NonZeroUsize`: I'm not sure it's worth the complexity.

src/lib.rs
```diff
@@ -3133,6 +3133,52 @@ pub trait Itertools: Iterator {
...
+    /// assert_equal(v.iter().tail(10), &v);
+    ///
+    /// assert_equal((0..100).tail(10), 90..100);
+    /// ```
+    ///
+    /// For double ended iterators without side-effects, you might prefer
+    /// `.rev().take(n).collect_vec().into_iter().rev()`
+    /// to have the same result without consuming the entire iterator.
+    #[cfg(feature = "use_alloc")]
+    fn tail(self, n: usize) -> VecIntoIter<Self::Item>
```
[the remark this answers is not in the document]
Philippe-Cholet: We are not doing const-generics yet but I'm sometimes thinking about it.
```rust
match iter.next_chunk() { // I want this stabilized BTW.
    Ok(arr) => {
        /*update; rotate;*/
        arr.into_iter()
    }
    Err(arr_iter) => arr_iter,
}
```
All that to say it would return `core::array::IntoIter` and that `VecIntoIter` seems the right type and I would expect a simple `Vec` from `collect_...`.

I'm also wondering if `tail` is the best name but I did not find a better one.

src/lib.rs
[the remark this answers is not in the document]
Philippe-Cholet: Our MSRV is 1.43.1 (old). Would it make sense to keep my `Vec` implementation and switch to `VecDeque` when the MSRV reaches 1.67 ?

src/lib.rs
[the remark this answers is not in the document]
jswrenn: Agreed, we should stick to `usize`.

src/lib.rs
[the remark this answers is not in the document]
Philippe-Cholet: After basic benchmark, I got an improvement with `if i + 1 == n { 0 } else { i + 1 }` over `(i + 1) % n`.

src/lib.rs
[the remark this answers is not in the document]
Philippe-Cholet: I took a look at the new implementation, nice!

If we were to change `VecIntoIter` to `VecDequeIntoIter` once our MSRV is big enough, I suppose it would be a breaking change?!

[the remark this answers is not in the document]
Philippe-Cholet: Or directly use `VecDeque`, I guess it's simpler.

Philippe-Cholet:
https://github.com/rust-itertools/itertools/pull/899#discussion_r1525222651 Do we keep the `n = 1` case?
https://github.com/rust-itertools/itertools/pull/899#discussion_r1525200029 Is there a better name than `tail`?

src/lib.rs
```diff
@@ -3133,6 +3135,49 @@ pub trait Itertools: Iterator {
...
+                self.last();
+                VecDeque::new()
+            }
+            1 => self.last().into_iter().collect(),
+            _ => {
+                let mut iter = self.fuse();
+                let mut data: VecDeque<_> = iter.by_ref().take(n).collect();
+                iter.for_each(|value| {
+                    data.pop_front();
+                    data.push_back(value);
```
phimuemue, line 3173: I hoped that there would be a method that avoids the intermediate bookkeeping of pop_front/push_back (namely, updating `len`), but I did not find one, so that's fine with me.

However, @scottmcm's go-ahead would be appreciated here.

src/lib.rs
```diff
@@ -3133,6 +3135,49 @@ pub trait Itertools: Iterator {
...
+    fn tail(self, n: usize) -> VecDequeIntoIter<Self::Item>
+    where
+        Self: Sized,
+    {
+        match n {
+            0 => {
+                self.last();
+                VecDeque::new()
+            }
+            1 => self.last().into_iter().collect(),
```
phimuemue, line 3167: At first sight, I also found the special cases odd, but I think it's worth to exploit specialized `last` implementations.

phimuemue (approved):
Hi @Philippe-Cholet, looks reasonable.

I approve this so that you can merge, but I'd appreciate a second pair of eyes, maybe @scottmcm?

src/lib.rs
Philippe-Cholet: I executed the benchmark I wrote earlier (`v.iter().tail(10)` with `let v = (0..1024).collect_vec();`) and the `VecDeque` version is 4 times slower.
Not sure why without investigation but apart from updating `len`, it also checks `is_empty` and `is_full` while the `Vec` version may just be more straightforward.

Philippe-Cholet: And `data.rotate_left(1); *data.back_mut().unwrap() = value;` is no better.

phimuemue: > I executed the benchmark I wrote earlier (`v.iter().tail(10)` with `let v = (0..1024).collect_vec();`) and the `VecDeque` version is 4 times slower. Not sure why without investigation but apart from updating `len`, it also checks `is_empty` and `is_full` while the `Vec` version may just be more straightforward.

Good catch. I guess it depends on the iterator length and the number of elements you collect. (When collecting the 10-tail of 1024 elements, maybe the iterations dominate the runtime, so that a `Vec` is better there and the final `rotate_left` does not matter that much?)

Under your findings, and unless we have evidence that `VecDeque` is better in realistic circumstances, I'd go with the simpler `Vec`, even accepting the breaking change if we one day discover the contrary.

Philippe-Cholet: I think people would mostly want a small tail. If they want most of the iterator, `collect_vec+truncate` is simpler.

After running cargo-flamegraph on the benchmark, I don't really know more: it spends most of the time in `push_back` and a bit (7 times less) in `pop_front`, nothing else matters. So the iteration dominate runtime.

Maybe I would have better understanding with _godbolt_. I think my straightforward `Vec` version is better optimized than the `VecDeque` one.

---

While we don't do const-generics yet, the `Vec` version would be easily generalizable to both cases because update buffers (`[_; N]` and `Vec<_>`) is update a slice while the `VecDeque` version hide details.

src/lib.rs
```diff
@@ -3133,6 +3135,49 @@ pub trait Itertools: Iterator {
...
+            _ => {
+                let mut iter = self.fuse();
```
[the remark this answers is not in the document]
Philippe-Cholet: The idea of skipping a large portion of the iteration sure seems interesing (for both versions).
I don't recall seeing size hints used that way (as it's mostly for allocations and `ExactSizeIterator::len`). I'm curious if you have such examples in mind.

src/lib.rs
Philippe-Cholet: Your "skip" idea is _awesome_ in the case the size hint knows enough but it does not change a thing otherwise.
This second idea of yours reduce the impact of the final `rotate_left`, nice!

And if there was a `pop_front_push_back` in the future (as efficient as my cyclic update of the vector), we could revisit this, collect to VecDeque, then `iter.for_each(|val| { data.pop_front_push_back(val); })`.

src/lib.rs
```diff
@@ -3133,6 +3135,59 @@ pub trait Itertools: Iterator {
...
+                // Skip the starting part of iterator if possible.
+                let (low, _) = self.size_hint();
+                let mut iter = self.fuse().skip(low.saturating_sub(n));
+                let mut data: Vec<_> = iter.by_ref().take(n).collect();
+                // Update `data` cyclically.
+                let idx = iter.fold(0, |i, val| {
+                    data[i] = val;
+                    if i + 1 == n {
+                        0
+                    } else {
```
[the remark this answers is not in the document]
Philippe-Cholet: I saw the codecov report but did not realize this, I need sleep. I just added a similar doctest.

Philippe-Cholet:
I polished the commits and added @scottmcm as co-author.
I particularly like his idea of `skip`ping the starting part of the iterator.

Is there a better name than `tail`? (see [this](https://github.com/rust-itertools/itertools/pull/899#discussion_r1525200029))
If not then I think this can be merged.

**Note for myself:** File a rust issue for a possible method more optimized than `.pop_front()+.push_back(_)`, compared to my vector implementation (see [this](https://github.com/rust-itertools/itertools/pull/899#discussion_r1526801449)).

src/lib.rs
```diff
@@ -3133,6 +3135,65 @@ pub trait Itertools: Iterator {
         self.k_largest_by(k, k_smallest::key_to_cmp(key))
     }
 
+    /// Consumes the iterator and return an iterator of the last `n` elements.
+    ///
+    /// It allocates up to `n` elements.
+    /// The iterator, if directly collected to a `VecDeque`, is converted
+    /// without any extra copying or allocation cost.
+    /// If directly collected to a `Vec`, it may need some data movement
+    /// but no re-allocation.
```
the8472, line 3144: You might want to leave some leeway for the implementation here. For example allocating a `VecDeque` with twice the necessary capacity will allow you to do the rotating in batches or use `next_chunk` to overfill it and then trim it back.

Philippe-Cholet: Allocating twice the necessary capacity definitely improved my benchmark for a `VecDeque`, but it still remains slower the current `Vec` version.

the8472: I was more talking about the general case, not documentating too much about today's implementation to leave room for future improvements.

Philippe-Cholet: Fair enough. Just about "_allocates up to `n`_" I suppose.

src/lib.rs
```diff
@@ -3133,6 +3135,64 @@ pub trait Itertools: Iterator {
...
+        match n {
+            0 => {
+                self.last();
+                VecDeque::new()
+            }
+            1 => self.last().into_iter().collect(),
+            _ => {
+                // Skip the starting part of the iterator if possible.
+                let (low, _) = self.size_hint();
+                let mut iter = self.fuse().skip(low.saturating_sub(n));
```
phimuemue, line 3174: I hope I did not miss relevant discussion, but is it ok here if `size_hint` is incorrect? E.g. an iterator with 3 elements, but `size_hint` returns "at least 4": `tail(2)` would then return 1 element, right?

src/lib.rs
```diff
@@ -3133,6 +3135,64 @@ pub trait Itertools: Iterator {
...
+                // TODO: If VecDeque has a more efficient method than
+                // `.pop_front();.push_back(val)` in the future then maybe revisit this.
+                let mut data: Vec<_> = iter.by_ref().take(n).collect();
+                // Update `data` cyclically.
+                let idx = iter.fold(0, |i, val| {
+                    data[i] = val;
+                    if i + 1 == n {
```
phimuemue, line 3181: Might `debug_assert_eq!(data.len(), n)` help here?

Philippe-Cholet: The iterator is fused, so if `data.len() != n` after `collect` then the iterator is finished and the closure in fold is not called. Do we really need this? I don't mind adding it though.

src/lib.rs
Philippe-Cholet: I had similar surprise myself (I asked for similar usage [here](https://github.com/rust-itertools/itertools/pull/899#discussion_r1526853345)) but the principal usage of `size_hint` outside memory allocation is `ExactSizeIterator::len` which we consider to be correct, right?

The [docs](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.size_hint) mentions that the size hint should not be trusted if an incorrect implementation might lead to memory safety violation. Other than that, an incorrect implementation is "a violation of the trait’s protocol" just as any other incorrect specialization I would say.

src/lib.rs
Philippe-Cholet: We have a similar assertion in "similar" `k_smallest` but not for its recent variants.

PR #900, opened 2024-03-09
Fix `Itertools::k_smallest` on short unfused iterators

Philippe-Cholet:
I'm shocked! 😲 I found a bug in my favorite method!

In the case of an unfused iterator yielding at most k-1 elements, then None, then other elements, the heap is not k-long and `for_each` is called on the other elements leading `debug_assert_eq` to rightfully panic.

Our recent variants of `k_smallest` do not have this bug (the `.len() ==` check is present)!

Alternatively, iterators could be fused: `let mut iter = iter.fuse();`.

I probably should add a test about this.

**Story:** After recently reviewing #654 where I deeply looked at `k_smallest` and its variants and my really recent #899 where I similarly thought of checking the length of the collected elements before calling `for_each`, I eventually/_luckily_ saw this bug.

### DIFF
--- a/src/k_smallest.rs
+++ b/src/k_smallest.rs
@@ -2,7 +2,7 @@ use alloc::vec::Vec;
 use core::cmp::Ordering;
 
 /// Consumes a given iterator, returning the minimum elements in **ascending** order.
-pub(crate) fn k_smallest_general<I, F>(mut iter: I, k: usize, mut comparator: F) -> Vec<I::Item>
+pub(crate) fn k_smallest_general<I, F>(iter: I, k: usize, mut comparator: F) -> Vec<I::Item>
 where
     I: Iterator,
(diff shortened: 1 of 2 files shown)


Philippe-Cholet:
I'm exploring the possibility of buggy methods in the case of unfused iterators.

phimuemue:
I think we should fuse them: `collect_vec`+`sort`+`truncate` should be the same as `k_smallest` (see https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=8e6e791d95adc335900ea6b8a2bfd95d).

Philippe-Cholet:
Explicitely fuse the iterator is also clearer than checking `.len() == k`.

Done. And I did the same for the recent `k_smallest_general`.

phimuemue (approved):
Thank you.

PR #901, opened 2024-03-12
More generic maps (in `GroupingMap`)

### LINKED ISSUE #588: Is there any way to make the `into_grouping_map` family of functions support custom hashers
I want to use FxHash for some HashMaps for speed, but into_grouping_map seems to be restricted to the builtin hasher. It would be great if this could be even more generic

Philippe-Cholet:
At the moment, this is only a proof of concept, but it would close #588 (and eventually solve some issues, see https://github.com/rust-itertools/itertools/labels/generic-container).

- I wrote a straightforward _private_ trait generalizing both `BTreeMap` and `HashMap` (with any hasher).
  Maybe too much methods, or not enough. I will adjust it.
  I don't think we should generalize to more than those two maps but I guess it's possible.
- Changed `use_std` to `use_alloc` where possible (because `BTreeMap` only needs `use_alloc` while `HashMap` needs `use_std`) and relaxed some `Eq + Hash` bounds.
- I added a alternative `aggregate_in` to `aggregate` which _takes one more argument_:
  any empty map (or one we would `.clear()`).

The last point is where I struggled the most but that's pretty much the only real possibility because there is so many possibilities of initializations: `BTreeMap::new()`, `HashMap::new()`, `HashMap::default()` for some hashers, and `with_capacity` shenanigans. And I think it's the most general solution (I can even imagine one single `HashMap` being used multiple times to avoid reallocations).

This could extend to other areas than just `GroupingMap`, and a `Set` trait could be wrote to generalize `BTreeSet` and `HashSet` but it's for another PR.

For now, I would like some feedback.

CI currently fails because of temporary unused imports.

### DIFF
--- a/src/generic_containers.rs
+++ b/src/generic_containers.rs
@@ -4,6 +4,7 @@
 #![cfg(feature = "use_alloc")]
 
 use alloc::collections::BTreeMap;
+use alloc::vec::Vec;
 #[cfg(feature = "use_std")]
 use core::hash::{BuildHasher, Hash};
 #[cfg(feature = "use_std")]
@@ -58,5 +59,36 @@ where
         V: Default,
     {
         self.entry(key).or_default()
+    }
+}
+
+impl<K, V> Map for Vec<(K, V)>
+where
+    K: Eq,
+{
+    type Key = K;
+    type Value = V;
+    fn insert(&mut self, key: K, value: V) -> Option<V> {
+        match self.iter_mut().find(|(k, _)| k == &key) {
+            Some((_, v)) => Some(core::mem::replace(v, value)),
+            None => {
+                self.push((key, value));
+                None
+            }
+        }
+    }
+    fn remove(&mut self, key: &K) -> Option<V> {
+        let index = self.iter().position(|(k, _)| k == key)?;
+        Some(self.swap_remove(index).1)
+    }
+    fn entry_or_default(&mut self, key: K) -> &mut V
+    where
+        V: Default,
+    {
+        let index = self.iter().position(|(k, _)| k == &key).unwrap_or_else(|| {
+            self.push((key, V::default()));
+            self.len() - 1
+        });
+        &mut self[index].1
     }
 }
--- a/src/generic_containers.rs
+++ b/src/generic_containers.rs
@@ -1,0 +1,62 @@
+//! **Private** generalizations of containers:
+//! - `Map`: `BTreeMap` and `HashMap` (any hasher).
+
+#![cfg(feature = "use_alloc")]
+
+use alloc::collections::BTreeMap;
+#[cfg(feature = "use_std")]
+use core::hash::{BuildHasher, Hash};
+#[cfg(feature = "use_std")]
+use std::collections::HashMap;
+
+pub trait Map {
+    type Key;
+    type Value;
+    fn insert(&mut self, key: Self::Key, value: Self::Value) -> Option<Self::Value>;
+    fn remove(&mut self, key: &Self::Key) -> Option<Self::Value>;
+    fn entry_or_default(&mut self, key: Self::Key) -> &mut Self::Value
+    where
+        Self::Value: Default;
+}
+
+impl<K, V> Map for BTreeMap<K, V>
+where
+    K: Ord,
+{
+    type Key = K;
+    type Value = V;
+    fn insert(&mut self, key: K, value: V) -> Option<V> {
+        self.insert(key, value)
+    }
+    fn remove(&mut self, key: &K) -> Option<V> {
+        self.remove(key)
+    }
+    fn entry_or_default(&mut self, key: K) -> &mut V
+    where
+        V: Default,
+    {
+        self.entry(key).or_default()
+    }
+}
+
+#[cfg(feature = "use_std")]
+impl<K, V, S> Map for HashMap<K, V, S>
+where
+    K: Eq + Hash,
+    S: BuildHasher,
+{
+    type Key = K;
+    type Value = V;
+    fn insert(&mut self, key: K, value: V) -> Option<V> {
+        self.insert(key, value)
+    }
+    fn remove(&mut self, key: &K) -> Option<V> {
+        self.remove(key)
+    }
+    fn entry_or_default(&mut self, key: K) -> &mut V
+    where
+        V: Default,
+    {
+        self.entry(key).or_default()
+    }
+}
--- a/src/grouping_map.rs
+++ b/src/grouping_map.rs
@@ -447,7 +447,7 @@ where
     where
         V: Ord,
     {
-        self.minmax_by(|_, v1, v2| V::cmp(v1, v2))
+        self.minmax_in(HashMap::new())
     }
 
     /// Groups elements from the `GroupingMap` source by key and find the maximum and minimum of
@@ -473,32 +473,11 @@ where
     /// assert_eq!(lookup[&2], OneElement(5));
     /// assert_eq!(lookup.len(), 3);
     /// ```
-    pub fn minmax_by<F>(self, mut compare: F) -> HashMap<K, MinMaxResult<V>>
+    pub fn minmax_by<F>(self, compare: F) -> HashMap<K, MinMaxResult<V>>
     where
         F: FnMut(&K, &V, &V) -> Ordering,
     {
-        self.aggregate(|acc, key, val| {
-            Some(match acc {
-                Some(MinMaxResult::OneElement(e)) => {
-                    if compare(key, &val, &e) == Ordering::Less {
-                        MinMaxResult::MinMax(val, e)
-                    } else {
-                        MinMaxResult::MinMax(e, val)
-                    }
-                }
-                Some(MinMaxResult::MinMax(min, max)) => {
-                    if compare(key, &val, &min) == Ordering::Less {
-                        MinMaxResult::MinMax(val, max)
-                    } else if compare(key, &val, &max) != Ordering::Less {
-                        MinMaxResult::MinMax(min, val)
-                    } else {
-                        MinMaxResult::MinMax(min, max)
-                    }
-                }
-                None => MinMaxResult::OneElement(val),
-                Some(MinMaxResult::NoElements) => unreachable!(),
-            })
-        })
+        self.minmax_by_in(compare, HashMap::new())
     }
 
     /// Groups elements from the `GroupingMap` source by key and find the elements of each group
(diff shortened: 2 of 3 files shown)


Philippe-Cholet:
Currently, we can do `it.into_grouping_map().product();` but we can't have a generic map.

```rust
let my_map = HashMap::new();
let my_map = HashMap::with_capacity(100);
let my_map = FxHashMap::default(); // rustc_hash
let mut my_map = FxHashMap::default(); my_map.reserve(100);
let my_map = BTreeMap::new();
```

### But where to give the map? (without breaking change)
**1.** Initial idea: suffix all `GroupingMap` methods with `_in`. But it results in many weird names like `minmax_by_key_in`.

    it.into_grouping_map().product_in(my_map);
    //                            --- ------

For each `GroupingMap` method like `product`, old `it.into_grouping_map().product();` would delegate to new `it.into_grouping_map().product_in(HashMap::new());`.

**2.** Suffix `Itertools` methods. The trait would be a bit more crowded. But only two redirections.

    it.into_grouping_map_in(my_map).product();
    //                  --- ------
Old `it.into_grouping_map();` would delegate to new `it.into_grouping_map_in(HashMap::new());`.
Similar for `into_grouping_map_by` -> `into_grouping_map_by_in` but that's it.

That would be similar to other possible extensions like `Itertools::counts_in` and `Itertools::all_unique_in`.

**3.** A bit nested. But no `_in` suffix.

    it.into_grouping_map().with_map(my_map).product();
    //                    -----------------
For each `GroupingMap` method like `product`, old `it.into_grouping_map().product();` would delegate to new `it.into_grouping_map().with_map(HashMap::new()).product();`.

phimuemue:
Hi @Philippe-Cholet, two thoughts:

* Does `into_grouping_map` play nicely with type inference? (Currently, `GroupingMap` returns maps with different value types, e.g. `minmax` returns `MinMaxResult<V>`s, but `max` returns `V`s.) If so, I would go with it (offers a central place to supply a map and generalizes well to `counts` et al).
* I know that for many day-to-day cases, a plain old (possibly sorted) `Vec<(K, V)>` can be more efficient than `{BTree|Hash}Map`. Thus, I think we should ensure that `trait Map` could support this type.

Philippe-Cholet:
@phimuemue 
I see your performance point about `Vec<(K, V)>`. It can always be added later (in the next PR, or any release) as we would not break anything by implementing `Map` for a new type.
It feels a bit tricky though.

**_Well, silly me..._!!** I don't know how it could work in a central place (cases 2 and 3 in my last comment).
We would create a new struct `GroupingMapIn` that `GroupingMap` would alias, right?
`pub type GroupingMap<I> = GroupingMapIn<I, HashMap<_, _>>;`... Simply blocked! The first one would be `<I as Iterator>::Item.0` (which I can't write) and the second is still unknown since it's the method using it that will determine it...
Then only _remain my initial idea (case 1)_ `method_name_in<..., M: Map<Key = ..., Value = ...>>(..., map: M) -> M`.

Philippe-Cholet:
- We need empty maps, but which of the following?
    - `map.clear();`
    - `assert!(map.is_empty());`
    - `debug_assert!(map.is_empty());`
- Implement `Map` for `Vec<(K, V)>` is possible but do we keep it?!

src/generic_containers.rs
```diff
@@ -0,0 +1,114 @@
...
+    where
+        V: Default,
+    {
+        self.entry(key).or_default()
+    }
+}
+
+impl<K, V> Map for Vec<(K, V)>
+where
+    K: PartialEq,
```
Philippe-Cholet, line 81: I chose vectors to be unsorted to have a smaller bound (`PartialEq` instead of `PartialOrd` for sorted ones). So implement `Map` for `Vec<(K, V)>` is possible but _do we keep it?!_

(if kept: add tests, update docs, update doc of "generic_containers" module).

src/grouping_map.rs
```diff
@@ -609,6 +559,241 @@ where
...
+    I: Iterator<Item = (K, V)>,
+{
+    /// Apply [`aggregate`](Self::aggregate) with a provided empty map
+    /// (`BTreeMap` or `HashMap` with any hasher).
+    pub fn aggregate_in<FO, R, M>(self, mut operation: FO, mut map: M) -> M
+    where
+        FO: FnMut(Option<R>, &K, V) -> Option<R>,
+        M: Map<Key = K, Value = R>,
+    {
+        debug_assert!(map.is_empty());
```
Philippe-Cholet, line 577: We need empty maps, but which of the following?
- `map.clear();` More convenient but may be surprising?
- `assert!(map.is_empty());` Maybe it should be checked outside of debug mode.
- `debug_assert!(map.is_empty());`

(Same for `collect_in`).

Philippe-Cholet (commented):
I will add tests using `BTreeMap`.

EDIT: doctests added.

Philippe-Cholet:
@SkiFire13 I see you are the author of `GroupingMap`. What do you think of such extension?

src/generic_containers.rs
SkiFire13: I think the point of using a `Vec<(K, V)>` is that it is faster when you have so few items that testing all of them is faster than doing something smarter with sorted or hashed data structures, so for me it makes sense to keep it unsorted.

If there's something I would change is to require `Eq` instead of `PartialEq` to catch possible mistakes with e.g. floating point numbers and to be more in line with the other map types. A counterpoint to this is that `PartialEq` is more flexible and a `Vec<(K, V)>` wouldn't break as hard as a `HashMap`/`BTreeMap` is the `Eq` contract wasn't respected.

src/grouping_map.rs
SkiFire13: I don't think this assert is really needed. The logic doesn't depend on the fact the map is clear, so I would leave this choice to the user. Allowing non-empty maps would also allow using these methods as `extend`-like variants of the others.

src/generic_containers.rs
```diff
@@ -0,0 +1,114 @@
...
+#[cfg(feature = "use_std")]
+use core::hash::{BuildHasher, Hash};
+#[cfg(feature = "use_std")]
+use std::collections::HashMap;
+
+pub trait Map {
+    type Key;
+    type Value;
+    fn is_empty(&self) -> bool;
+    fn clear(&mut self);
```
SkiFire13, line 17: I would remove these since they are pretty much unused.

src/generic_containers.rs
Philippe-Cholet: I `grep`ped `Eq` to see our usage of it and it's only used at the same we use `Hash` (so `HashMap/HashSet`).
However, we could use `Eq` (I like the consistency with other maps) and eventually relax the bound to `PartialEq` if someone comes with a really good argument.

src/grouping_map.rs
```diff
@@ -609,6 +565,484 @@ where
...
+
+        self.iter.for_each(|(key, val)| {
+            let acc = map.remove(&key);
+            if let Some(op_res) = operation(acc, &key, val) {
+                map.insert(key, op_res);
+            }
```
SkiFire13, line 611: Since these two are the only uses of `remove` and `insert` you could replace them in the trait with the combined operation, which should also be implementable in a more efficient way for some types (e.g. `Vec<(K, V)>` but also `hashbrown`'s `HashMap` if that will ever be added)

src/generic_containers.rs
```diff
@@ -0,0 +1,114 @@
...
+#![cfg(feature = "use_alloc")]
+
+use alloc::collections::BTreeMap;
+use alloc::vec::Vec;
+#[cfg(feature = "use_std")]
+use core::hash::{BuildHasher, Hash};
+#[cfg(feature = "use_std")]
+use std::collections::HashMap;
+
+pub trait Map {
```
SkiFire13, line 13: If we want to commit to the "pass an existing map to extend" pattern then you could add an implementation of `Map` for `&mut M where M: Map` to make it even more flexible.

SkiFire13 (commented):
> **_Well, silly me..._!!** I don't know how it could work in a central place (cases 2 and 3 in my last comment).
> We would create a new struct `GroupingMapIn` that `GroupingMap` would alias, right?
> `pub type GroupingMap<I> = GroupingMapIn<I, HashMap<_, _>>;`... Simply blocked! The first one would be `<I as Iterator>::Item.0` (which I can't write) and the second is still unknown since it's the method using it that will determine it...

You could add two generics (one for the key and one for the value) to `GroupingMap` (which would feel a bit weird though, but I think this is fundamentally required if we want to move the "in" at the start of the chain) and then restrict the various methods to `M: Map<Value = ExpectedValue>`.

---

Overall, this looks good to me. The duplication of the methods is a bit unfortunate though.

Could this be extended to `into_group_map[_by]` too?

src/generic_containers.rs
Philippe-Cholet: I planned on removing the unused one, but will remove both indeed.

src/grouping_map.rs
Philippe-Cholet: Good point. But I might prefer to optimize later as `Map` is meant to be used for way more than `GroupingMap` and I would prefer a larger point of view first. It won't be long (honestly `GroupingMap` is the biggest chunk for it).

Currently, `Map` is implemented for `std::collections::HashMap` so `hashbrown::HashMap` (which is not an alias of it) is not supported. What would be nice is `HashMap` to be moved to `alloc` but we are not there yet.

src/generic_containers.rs
Philippe-Cholet: I find this interesting and I will keep it in mind but I want to use `Map` on other things than `GroupingMap` so I'd rather keep it as simple as possible for now.

Philippe-Cholet:
@SkiFire13 

> The duplication of the methods is a bit unfortunate though.

I totally agree, I'm gonna investigate your "add two generics" idea.

> Could this be extended to `into_group_map[_by]` too?

Yes, in the next PR. I intend to extend all `HashMap` usage in the crate (and `HashSet` later), most https://github.com/rust-itertools/itertools/labels/generic-container issues should be solved with this `Map`.

Because I want to make all this work first, I'm gonna delay two of your ideas:

- Optimize `remove+insert` in `GroupingMap::aggregate_in`.
- ``impl<M: Map> Map for &mut M``.

SkiFire13:
> Because I want to make all this work first, I'm gonna delay two of your ideas:
> 
>     * Optimize `remove+insert` in `GroupingMap::aggregate_in`.
> 
>     * `impl<M: Map> Map for &mut M`.

Understandable. As long as `Map` is private this can be changed later, so I'm fine with it.

PR #906, opened 2024-03-19
More generic maps in `GroupingMap` (v2)

### LINKED ISSUE #588: Is there any way to make the `into_grouping_map` family of functions support custom hashers

Philippe-Cholet:
This an alternative to #901 that I learnt from, it would close #901, close #588 and eventually solve most https://github.com/rust-itertools/itertools/labels/generic-container issues.

- I reuse `Map` polished earlier.
- Transform `GroupingMap[By]` into `GroupingGenericMap[By]`
  This is a huge but quite straightforward transformation:
  - `GroupingMap[By]` becomes `GroupingGenericMap[By]` with a generic type `M`
  - `K: Hash` becomes `M: Map<Key = K>`
  - `HashMap<K, ...>` becomes `M` with a `M: Map<Value = ...>` bound
  - Remove `Hash` and `HashMap`
  - `use_std` becomes `use_alloc`
- Reintroduce `GroupingMap[By]` with **one more generic parameter**
  This is a **breaking change!** However, this is mostly inferred.
  Note that it breaks (only) laziness tests because the `GroupingMap` object is not used (I could write any type but `u8` is one of the possible returned types). If it was used then it would have been inferred.
  I copied/pasted the code of old `into_grouping_map[_by]` methods, added the generic `R` and updated each function body with `_in(.., HashMap::new())`.
  _The old types are aliases of the new ones._

The main interest of this PR over #901 is that **there is not `_in` versions for each `GroupingMap` method.**

---

More than the additional parameter, the following code (weird but currently working) would not compile anymore.

```rust
let g = (0..10).into_grouping_map_by(|n| n % 2); // the map would now lives here so the values type is fixed earlier.
if true {
    println!("{:?}", g.collect::<Vec<_>>()); // HashMap<i32, Vec<i32>>
} else {
    println!("{:?}", g.sum());               // HashMap<i32, i32>
}
```
However, it's easily solvable here, `.into_grouping_map_by(|n| n % 2)` should be moved to both branches for the code to work as before.

_Now, in other cases, it might be not that easily solvable!_ **Is it a too big breaking change?**

### DIFF
--- a/src/generic_containers.rs
+++ b/src/generic_containers.rs
@@ -15,6 +15,15 @@ pub trait Map {
     type Value;
     fn insert(&mut self, key: Self::Key, value: Self::Value) -> Option<Self::Value>;
     fn remove(&mut self, key: &Self::Key) -> Option<Self::Value>;
+    fn aggregate<T, F>(&mut self, key: Self::Key, t: T, mut operation: F)
+    where
+        F: FnMut(Option<Self::Value>, &Self::Key, T) -> Option<Self::Value>,
+    {
+        let opt_value = self.remove(&key);
+        if let Some(value) = operation(opt_value, &key, t) {
+            self.insert(key, value);
+        }
+    }
     fn entry_or_default(&mut self, key: Self::Key) -> &mut Self::Value
     where
         Self::Value: Default;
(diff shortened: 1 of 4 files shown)


Philippe-Cholet:
To generalize the "quick tests", I would need to generalize `into_group_map` as well.
To avoid a very huge PR, I think that I should implement `Map` only for `HashMap` first then `BTreeMap` and only then `Vec` in following PRs.

@SkiFire13 I was able to make it work with a single additional generic parameter.

@phimuemue @jswrenn  Do you think such breaking change is acceptable? I worry it might be too nasty for very few people.
I therefore searched on [github](https://github.com/search?q=into_grouping_map+language%3ARust&type=code) and eventually found only one place this would break: [they](https://github.com/Cardinal-Cryptography/AlephBFT/blob/43ab2917f0200f43999d6b4df72b8df21fd9cd10/consensus/src/units/store.rs#L88-L96) would be able to easily fix this by moving `.into_grouping_map()` into their two use cases.

PR #909, opened 2024-03-29
Case 0 and 1 of `k_smallest` and variants

### LINKED ISSUE #904: `k_smallest[_by]/tail`: similarities and differences
I find my new `tail` (#899) quite similar to `k_smallest` and its recent variants: we fuse the iterator, collect a limited number of items, than update the resulting slice with `iter.for_each`.

In `tail`, I consume the iterator (doing so can have wanted side effects) even when `n = 0` while `k_smallest` and variants do not but merely drop it.
In `tail`, I have a special case for `n = 1` while `k_smallest` and variants do not (it could use a possibly faster `min[_by]`).

I think we should update `k_smallest` and `k_smallest_general` accordingly.

Do you see other differences we should fix?

Philippe-Cholet:
Fixes #904.
About the case `k=0`, I wonder if all our iterators behave well on iterators with side-effects. I think we should update the specialization tests.

### DIFF
--- a/src/k_smallest.rs
+++ b/src/k_smallest.rs
@@ -42,6 +42,7 @@ where
     }
 
     if k == 0 {
+        iter.last();
         return Vec::new();
     }
     let mut iter = iter.fuse();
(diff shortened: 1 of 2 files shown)


phimuemue (approved):
Seems consistent with tail

PR #910, opened 2024-04-03
Implement DoubleEndedIterator for ProcessResults

shepmaster:


Philippe-Cholet (commented):
This looks good to me, and I appreciate the update on specialization tests.

However, this leads to a change in meaning. That way, the first error encountered is not necessarily the first in the iterator order. I know that's the all point but I'm unsure about such change.
We can `it.rev().process_results(..)` if we only need it in reverse so this change only matters when we need both iteration sides. Do you have an example?

phimuemue (approved):
Thanks for this.

As for the `DoubleEndedIterator`: I think it's fine to stop with another error.

As for the use cases: `try_rfold` might be specialized for the underlying iterator.

Philippe-Cholet (approved):

PR #914, opened 2024-04-13
Implement `Combinations::nth`

kinto-b:
Hi there, this PR addresses #301 in the same way as #329. @Philippe-Cholet I was indeed wrong that this wouldn't give a substantial performance gain

```
cargo bench --bench specializations "combinations[1234]/nth"

combinations1/nth       time:   [4.8406 ms 4.8562 ms 4.8734 ms]
                        change: [-53.963% -53.706% -53.442%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 15 outliers among 100 measurements (15.00%)
  6 (6.00%) high mild
  9 (9.00%) high severe

combinations2/nth       time:   [4.6338 ms 4.6532 ms 4.6762 ms]
                        change: [-55.198% -54.769% -54.327%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 11 outliers among 100 measurements (11.00%)
  1 (1.00%) high mild
  10 (10.00%) high severe

combinations3/nth       time:   [4.7743 ms 4.8138 ms 4.8600 ms]
                        change: [-53.934% -53.322% -52.707%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 14 outliers among 100 measurements (14.00%)
  8 (8.00%) high mild
  6 (6.00%) high severe

combinations4/nth       time:   [5.1255 ms 5.1524 ms 5.1841 ms]
                        change: [-53.459% -52.613% -51.852%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 18 outliers among 100 measurements (18.00%)
  6 (6.00%) high mild
  12 (12.00%) high severe
```

As discussed in the comments of #301, an even more efficient solution which would be performant even in the large n/k regime, would be to use the combinatorial number system. For record keeping sake, here's the function I sketched out before I realised that the itertools uses a 'reversed' lexicographic ordering.

```rust
    fn nth(&mut self, n: usize) -> Option<Self::Item> {
        self.state += n;

        // https://en.wikipedia.org/wiki/Combinatorial_number_system#Finding_the_k-combination_for_a_given_number
        let mut remainder = self.state;
        let mut delta = 0;
        for i in (0..self.k()).rev() {
            let mut m = i;
            while let Some(d) = checked_binomial(m, i + 1) {
                if d > remainder {
                    self.indices[i] = m - 1;
                    remainder -= delta;
                    break;
                }

                m += 1;
                delta = d;
            }
        }

        // We may need to pad out the pool
        if let Some(x) = self.indices.last() {
            self.pool.prefill(x + 1);
            if *x >= self.n() {
                return None;
            }
        }

        // We should only return an empty list once
        if self.indices.is_empty() && self.state > 0 {
            return None;
        }

        Some(self.indices.iter().map(|i| self.pool[*i].clone()).collect())
    }
```

It's possible to use this same idea with the lexicographic ordering used by itertools, but we'd need to know number of elements in the underlying iterator. Suppose we do, and we label this `N`. Then the function above would just need to be modified so that

1. We initialise `remainder = checked_binomial(N, self.k()) - n - 1`
2. We 'invert' the indices before returning the combination so that `i = N-1-i` and then reverse the order to ensure that the indices remain strictly increasing.

### DIFF
--- a/src/combinations.rs
+++ b/src/combinations.rs
@@ -14,6 +14,7 @@ pub struct Combinations<I: Iterator> {
     indices: Vec<usize>,
     pool: LazyBuffer<I>,
     first: bool,
+    done: bool,
 }
 
 impl<I> Clone for Combinations<I>
@@ -21,7 +22,7 @@ where
     I: Clone + Iterator,
     I::Item: Clone,
 {
-    clone_fields!(indices, pool, first);
+    clone_fields!(indices, pool, first, done);
 }
 
 impl<I> fmt::Debug for Combinations<I>
@@ -29,7 +30,7 @@ where
     I: Iterator + fmt::Debug,
     I::Item: fmt::Debug,
 {
-    debug_fmt_fields!(Combinations, indices, pool, first);
+    debug_fmt_fields!(Combinations, indices, pool, first, done);
 }
 
 /// Create a new `Combinations` from a clonable iterator.
@@ -41,6 +42,7 @@ where
         indices: (0..k).collect(),
         pool: LazyBuffer::new(iter),
         first: true,
+        done: false,
     }
 }
 
@@ -70,6 +72,7 @@ impl<I: Iterator> Combinations<I> {
     /// elements.
     pub(crate) fn reset(&mut self, k: usize) {
         self.first = true;
+        self.done = false;
 
         if k < self.indices.len() {
             self.indices.truncate(k);
@@ -90,9 +93,55 @@ impl<I: Iterator> Combinations<I> {
             indices,
             pool,
             first,
+            done: _,
         } = self;
         let n = pool.count();
         (n, remaining_for(n, first, &indices).unwrap())
+    }
+
+    /// Initialises the iterator by filling a buffer with elements from the
+    /// iterator.
+    fn init(&mut self) {
+        self.pool.prefill(self.k());
+        if self.k() > self.n() {
+            self.done = true;
+        } else {
+            self.first = false;
+        }
+    }
+
+    /// Increments indices representing the combination to advance to the next
+    /// (in lexicographic order by increasing sequence) combination. For example
+    /// if we have n=3 & k=2 then [0, 1] -> [0, 2] -> [0, 3] -> [1, 2] -> ...
+    fn increment_indices(&mut self) {
+        if self.indices.is_empty() {
+            self.done = true;
+            return;
+        }
+
+        // Scan from the end, looking for an index to increment
+        let mut i: usize = self.indices.len() - 1;
+
+        // Check if we need to consume more from the iterator
+        if self.indices[i] == self.pool.len() - 1 {
+            self.pool.get_next(); // may change pool size
+        }
+
+        while self.indices[i] == i + self.pool.len() - self.indices.len() {
+            if i > 0 {
+                i -= 1;
+            } else {
+                // Reached the last combination
+                self.done = true;
+                return;
+            }
+        }
+
+        // Increment index, and reset the ones to its right
+        self.indices[i] += 1;
+        for j in i + 1..self.indices.len() {
+            self.indices[j] = self.indices[j - 1] + 1;
+        }
     }
 }
 
@@ -104,40 +153,34 @@ where
     type Item = Vec<I::Item>;
     fn next(&mut self) -> Option<Self::Item> {
         if self.first {
-            self.pool.prefill(self.k());
-            if self.k() > self.n() {
-                return None;
-            }
-            self.first = false;
-        } else if self.indices.is_empty() {
-            return None;
+            self.init()
         } else {
-            // Scan from the end, looking for an index to increment
-            let mut i: usize = self.indices.len() - 1;
+            self.increment_indices()
+        }
 
-            // Check if we need to consume more from the iterator
-            if self.indices[i] == self.pool.len() - 1 {
-                self.pool.get_next(); // may change pool size
-            }
+        if self.done {
+            return None;
+        }
 
-            while self.indices[i] == i + self.pool.len() - self.indices.len() {
-                if i > 0 {
-                    i -= 1;
-                } else {
-                    // Reached the last combination
-                    return None;
-                }
-            }
+        Some(self.indices.iter().map(|i| self.pool[*i].clone()).collect())
+    }
 
-            // Increment index, and reset the ones to its right
-            self.indices[i] += 1;
-            for j in i + 1..self.indices.len() {
-                self.indices[j] = self.indices[j - 1] + 1;
+    fn nth(&mut self, n: usize) -> Option<Self::Item> {
+        // Delegate initialisation work to next()
+        let first = self.next();
+
+        if n == 0 {
+            return first;
+        }
+
+        for _ in 0..(n - 1) {
+            self.increment_indices();
+            if self.done {
+                return None;
             }
         }
 
-        // Create result vector based on the indices
-        Some(self.indices.iter().map(|i| self.pool[*i].clone()).collect())
+        self.next()
     }
 
     fn size_hint(&self) -> (usize, Option<usize>) {
--- a/src/combinations.rs
+++ b/src/combinations.rs
@@ -96,8 +96,7 @@ impl<I: Iterator> Combinations<I> {
     }
 
     /// Initialises the iterator by filling a buffer with elements from the
-    /// iterator, return a boolean indicating whether or not we've run out of
-    /// combinations.
+    /// iterator. Returns true if there are no combinations, false otherwise.
     fn init(&mut self) -> bool {
         self.pool.prefill(self.k());
         let done = self.k() > self.n();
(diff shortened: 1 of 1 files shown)


Philippe-Cholet (commented):
Thanks for this work, I really appreciate it.

I'm gonna skip discussion about using the "combinatorial number system" for now and focus on the current PR.

src/combinations.rs
```diff
@@ -14,22 +14,23 @@ pub struct Combinations<I: Iterator> {
     indices: Vec<usize>,
     pool: LazyBuffer<I>,
     first: bool,
+    done: bool,
```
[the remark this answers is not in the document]
kinto-b: Yep, that's fair. I initially had them doing exactly that but ended up refactoring as I felt it made the code a little bit harder to read. I'll revert it when I get a chance

src/combinations.rs
```diff
@@ -104,40 +153,34 @@ where
...
-                    // Reached the last combination
-                    return None;
-                }
-            }
+        Some(self.indices.iter().map(|i| self.pool[*i].clone()).collect())
+    }
+
+    fn nth(&mut self, n: usize) -> Option<Self::Item> {
+        // Delegate initialisation work to next()
+        let first = self.next();
```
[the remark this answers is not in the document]
kinto-b: No worries, I'll make this change shortly, thanks for reviewing :)

kinto-b:
Have made those two patches, thanks for reviewing :)

```
cargo bench --bench specializations "combinations[1234]/nth"

combinations1/nth       time:   [3.9183 ms 3.9953 ms 4.0795 ms]
                        change: [-19.349% -17.726% -16.196%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 9 outliers among 100 measurements (9.00%)
  8 (8.00%) high mild
  1 (1.00%) high severe

combinations2/nth       time:   [3.5200 ms 3.5950 ms 3.6761 ms]
                        change: [-24.364% -22.741% -21.022%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 17 outliers among 100 measurements (17.00%)
  10 (10.00%) high mild
  7 (7.00%) high severe

combinations3/nth       time:   [3.5204 ms 3.5581 ms 3.5994 ms]
                        change: [-27.161% -26.085% -24.983%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  2 (2.00%) high mild
  2 (2.00%) high severe

combinations4/nth       time:   [3.7733 ms 3.8175 ms 3.8691 ms]
                        change: [-26.955% -25.909% -24.814%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 6 outliers among 100 measurements (6.00%)
  2 (2.00%) high mild
  4 (4.00%) high severe
```

src/combinations.rs
```diff
@@ -94,6 +94,56 @@ impl<I: Iterator> Combinations<I> {
...
+        if !done {
+            self.first = false;
+        }
+
+        done
+    }
+
+    /// Increments indices representing the combination to advance to the next
+    /// (in lexicographic order by increasing sequence) combination. For example
+    /// if we have n=3 & k=2 then [0, 1] -> [0, 2] -> [0, 3] -> [1, 2] -> ...
```
Philippe-Cholet, line 113: If n=3 then the indices are 0 1 2 and there is no 3. So n=4 here.

src/combinations.rs
```diff
@@ -94,6 +94,56 @@ impl<I: Iterator> Combinations<I> {
...
+    ///
+    /// Returns a boolean indicating whether or not we've run out of combinations.
```
Philippe-Cholet, line 115: Please rephrase `Returns true if we've run out of...` as it better indicates what `true/false` means.
EDIT: I was not fully awake when I said that but I still think it's very slightly better.

src/combinations.rs
```diff
@@ -103,41 +153,42 @@ where
...
+        for _ in 0..(n - 1) {
+            done = self.increment_indices();
+            if done {
+                return None;
             }
         }
 
-        // Create result vector based on the indices
-        Some(self.indices.iter().map(|i| self.pool[*i].clone()).collect())
+        self.next()
```
Philippe-Cholet, line 191: Should we not iterate `0..n` and return `Some(...collect())` rather `self.next()` which has `init` code we don't need at this point, right?

I assume it was not to duplicate the `indices.iter...collect` part.
We could (in another PR) have one helper function to do this `indices.iter...collect` which would be basically used in all iterators returning vector items.

Philippe-Cholet (changes requested):

src/combinations.rs
kinto-b: Ah yes, excuse me. Will patch

src/combinations.rs
kinto-b: Yeah no worries :)

src/combinations.rs
kinto-b: Yep that was my thinking and yeah I make that change to skip the extra boolean comparison if you like?

Philippe-Cholet: I don't see the "extra boolean comparison".
If I inlined this last `next`, I'd get a `let done = if first { init /*won't get here, might as well delete it*/} else { increment }; if done {return None} Some(collect)` so it basically completes the `0..n-1` to `0..n` and then return some.

src/combinations.rs
```diff
@@ -103,41 +153,42 @@ where
...
+            self.increment_indices()
+        };
+
+        if done {
+            return None;
+        }
+
+        for _ in 0..(n - 1) {
+            done = self.increment_indices();
+            if done {
```
Philippe-Cholet, line 186: Real nitpick but we don't need to update `done` but just use the boolean.
`if self.increment_indices() {` (and ~`mut`~ `done` above)

src/combinations.rs
kinto-b: Ah yeah I mean the way I currently have it has an extra comparison. Happy to change it if you like?

src/combinations.rs
kinto-b: nitpicks are good! I welcome them

src/combinations.rs
Philippe-Cholet: Yes I'd like this change.

Philippe-Cholet (approved):
I'd prefer if you squashed the commits but I'll merge this anyway as it's nicely done.
I'm curious of the benchmarks, if you _want to_ run it once before the changes and once after.

kinto-b:
You can squash them all together at merge time:

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/about-pull-request-merges#squash-and-merge-your-commits

Here are the benchmarks

```
combinations1/nth       time:   [3.9468 ms 3.9861 ms 4.0303 ms]
                        change: [-65.786% -65.264% -64.765%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 8 outliers among 100 measurements (8.00%)
  3 (3.00%) high mild
  5 (5.00%) high severe

combinations2/nth       time:   [3.8809 ms 3.9623 ms 4.0542 ms]
                        change: [-65.562% -64.704% -63.750%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 14 outliers among 100 measurements (14.00%)
  3 (3.00%) high mild
  11 (11.00%) high severe

combinations3/nth       time:   [4.0474 ms 4.1469 ms 4.2530 ms]
                        change: [-65.039% -64.163% -63.323%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  3 (3.00%) high mild

combinations4/nth       time:   [4.0968 ms 4.1343 ms 4.1770 ms]
                        change: [-65.459% -65.056% -64.615%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 8 outliers among 100 measurements (8.00%)
  2 (2.00%) high mild
  6 (6.00%) high severe
```

Philippe-Cholet:
I think the "Squash and merge" option is not selected (and I'm not the owner). I'll merge as is.

Thanks for the benchmarks. -65% is nice right?
The benchmark does repeatedly `.nth(0)`s ... `.nth(9)`s so it does not give much info on each `nth(n)` though.
`nth(n)` was `n+1` heap-allocations before and is now only one, thanks! I think it's roughly `n+1` faster.

About making `nth` even more performant with the "combinatorial number system", it would require to store `state` field (and therefore update elsewhere too) or use my `remaining_for` function each time `nth` is called.
I guess it would be fantastic for `.nth(/*big*/n)`. I could imagine 

```rust
#[inline(never)]
fn nth_perf(&mut self, n: usize) ... { ... }

fn nth(&mut self, n: usize) ... {
    if n == 0 { ... }
    if n >= ARBITRARY_VALUE { return self.nth_perf(n); }
    ...
}
```
but I wonder if it's worth the trouble. With `Vec` items, our combinations are more convenient than truly performant.
It's nice to have a specialized `nth` but for more critical performance, another entire implementation could be better.
Without people asking for it, I tend to think we should not.

> also thanks for reviewing these PRs. If there's anything else you think I might be able to work on, let me know :)

Well, I'll surely mention you soon.

kinto-b:
@Philippe-Cholet Yep I'll add that helper and have a browse of the other outstanding fold specializations :) Feel free to @ me in anything

PR #915, opened 2024-04-15
Do not implement `ZipEq::fold` (and change test)

kinto-b:
Relates to #755. 

I can't really think of a better way to implement this, but it still yields a regression :(

```
cargo bench --bench specializations "zip_eq/fold"

zip_eq/fold             time:   [509.89 ns 510.95 ns 512.02 ns]
                        change: [+4.8744% +5.1665% +5.4970%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

### DIFF
--- a/tests/quick.rs
+++ b/tests/quick.rs
@@ -571,6 +571,14 @@ quickcheck! {
         let b = &b[..len];
         itertools::equal(zip_eq(a, b), zip(a, b))
     }
+
+    #[should_panic]
+    fn zip_eq_panics(a: Vec<u8>, b: Vec<u8>) -> TestResult {
+        if a.len() == b.len() { return TestResult::discard(); }
+        zip_eq(a.iter(), b.iter()).for_each(|_| {});
+        TestResult::passed() // won't come here
+    }
+
     fn equal_positions(a: Vec<i32>) -> bool {
         let with_pos = a.iter().positions(|v| v % 2 == 0);
         let without = a.iter().enumerate().filter(|(_, v)| *v % 2 == 0).map(|(i, _)| i);
--- a/tests/zip.rs
+++ b/tests/zip.rs
@@ -1,4 +1,3 @@
-use itertools::free::zip_eq;
 use itertools::multizip;
 use itertools::EitherOrBoth::{Both, Left, Right};
 use itertools::Itertools;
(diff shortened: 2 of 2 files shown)


src/zip_eq_impl.rs
```diff
@@ -52,6 +52,25 @@ where
...
+    fn fold<B, F>(self, init: B, mut f: F) -> B
+    where
+        Self: Sized,
+        F: FnMut(B, Self::Item) -> B,
+    {
+        let Self { a, mut b } = self;
+
+        let acc = a.fold(init, |acc, x| match b.next() {
+            Some(y) => f(acc, (x, y)),
+            None => panic!("itertools: .zip_eq() reached end of one iterator before the other"),
```
Philippe-Cholet, line 64: You could `b.next().expect(..)`. I doubt it's faster though.

tests/zip.rs
```diff
@@ -73,3 +73,21 @@ fn zip_eq_panic2() {
 
     zip_eq(&a, &b).count();
 }
+
+#[should_panic]
```
Philippe-Cholet, line 77: Ah! I just checked (`ripgrep should_panic`) we don't really have `should_panic` tests which is valuable. But maybe a quickcheck test instead in `quick.rs`, something like
```rust
#[should_panic]
fn zip_eq_panics(a: Vec<u8>, b: Vec<u8>) -> TestResult {
    if a.len() == b.len() { return TestResult::discard(); }
    a.zip_eq(b).for_each(|_| {});
}
```
unless quickcheck tests do not like to panic. I don't know I didn't try.

Philippe-Cholet (commented):
To be honest, I did specialize the `fold`s that felt the most promising to me but I thought this would be a win.

tests/zip.rs
kinto-b: works perfectly well :) have replaced the custom tests with a couple quick tests

src/zip_eq_impl.rs
kinto-b: just tested it, no difference

Philippe-Cholet (commented):
After changes, if you could squash commits.

tests/quick.rs
```diff
@@ -571,6 +571,21 @@ quickcheck! {
...
+        zip_eq(a.iter(), b.iter()).count();
+        TestResult::passed() // won't come here
+    }
+
+    #[should_panic]
+    fn zip_eq_fold_panics(a: Vec<u8>, b: Vec<u8>) -> TestResult {
+        if a.len() == b.len() { return TestResult::discard(); }
+        zip_eq(a.iter(), b.iter()).fold((), |_, _| ());
+        TestResult::passed() // won't come here
+    }
```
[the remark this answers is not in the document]
kinto-b: Ah gotcha, will update

Philippe-Cholet (approved):
Thanks.

kinto-b:
@Philippe-Cholet pleasure, thank you :)

PR #917, opened 2024-04-17
`LazyBuffer::get_at` (+ simplify `Combinations::nth`)

kinto-b:
As requested in https://github.com/rust-itertools/itertools/pull/914#issuecomment-2056489069, I've added a helper for returning the 'current' combination. 

I also realised that there's no reason to delegate from `.nth` to `.next()` when `n==0` so I junked it.

### DIFF
--- a/src/combinations.rs
+++ b/src/combinations.rs
@@ -162,14 +162,10 @@ where
             return None;
         }
 
-        Some(self.indices.iter().map(|i| self.pool[*i].clone()).collect())
+        Some(self.pool.get_at(&self.indices))
     }
 
     fn nth(&mut self, n: usize) -> Option<Self::Item> {
-        if n == 0 {
-            return self.next();
-        }
-
         let done = if self.first {
             self.init()
         } else {
@@ -186,7 +182,7 @@ where
             }
         }
 
-        Some(self.indices.iter().map(|i| self.pool[*i].clone()).collect())
+        Some(self.pool.get_at(&self.indices))
     }
 
     fn size_hint(&self) -> (usize, Option<usize>) {
--- a/src/permutations.rs
+++ b/src/permutations.rs
@@ -99,7 +99,7 @@ where
                             return None;
                         }
                     }
-                    let item = indices[0..*k].iter().map(|&i| vals[i].clone()).collect();
+                    let item = vals.get_at(&indices[0..*k]);
                     *state = PermutationState::Loaded { indices, cycles };
                     Some(item)
                 }
@@ -110,7 +110,7 @@ where
                     return None;
                 }
                 let k = cycles.len();
-                Some(indices[0..k].iter().map(|&i| vals[i].clone()).collect())
+                Some(vals.get_at(&indices[0..k]))
             }
             PermutationState::End => None,
         }
--- a/src/combinations_with_replacement.rs
+++ b/src/combinations_with_replacement.rs
@@ -30,17 +30,6 @@ where
     debug_fmt_fields!(CombinationsWithReplacement, indices, pool, first);
 }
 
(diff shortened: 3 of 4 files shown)


src/combinations.rs
```diff
@@ -162,14 +172,10 @@ where
...
         }
 
-        Some(self.indices.iter().map(|i| self.pool[*i].clone()).collect())
+        Some(self.current_combination())
     }
 
     fn nth(&mut self, n: usize) -> Option<Self::Item> {
-        if n == 0 {
-            return self.next();
-        }
```
Philippe-Cholet, line 171: Good catch.

src/combinations.rs
```diff
@@ -145,6 +145,16 @@ impl<I: Iterator> Combinations<I> {
     }
 }
 
+impl<I> Combinations<I>
+where
+    I: Iterator,
+    I::Item: Clone,
+{
+    fn current_combination(&self) -> Vec<I::Item> {
+        self.indices.iter().map(|i| self.pool[*i].clone()).collect()
```
Philippe-Cholet, line 154: This is not what I meant but I just was not clear enough.
This code `indices.iter().map("index clone").collect()` do not appear only in `Combinations`, there is also `CombinationsWithReplacement`, `Permutations`, ~`MultiProduct`~ (oops, no I changed this one, good it simplifies things).
`LazyBuffer` could have a method `fn from_indices(&self, indices: &[usize]) -> Vec<I::Item>` that would be used in 3 iterators (and 3 files).

Philippe-Cholet (changes requested):

src/combinations.rs
kinto-b: Ah I see what you mean. Yes almost the exact same function is already implemented in `CombinationsWithReplacement`. I'll push a patch in a moment

Philippe-Cholet: With `git grep "clone()).collect()"`, you would probably see most places to use this `from_indices` method.

Philippe-Cholet: And you can remove this `current_combination` method, and probable similar method(s) in other iterators.

src/permutations.rs
```diff
@@ -81,12 +81,9 @@ where
...
             PermutationState::Buffered { ref k, min_n } => {
                 if vals.get_next() {
-                    let item = (0..*k - 1)
-                        .chain(once(*min_n))
-                        .map(|i| vals[i].clone())
-                        .collect();
+                    let indices: Vec<usize> = (0..k - 1).chain(once(*min_n)).collect();
                     *min_n += 1;
-                    Some(item)
+                    Some(vals.get_at(&indices))
```
Philippe-Cholet, line 86: We strive to make as few heap allocations as possible. This creates an unnecessary vector `indices` so I would change this back. `get_at` is useful but has its limitations.

I've been thinking that it could be `fn get_at(indices: impl Iterator<Item = usize>) ...` instead but I prefer to keep this as simple as possible (and it would requires to change `&indices` to `indices.iter().copied()` which would kinda beat the purpose of this helper, or `impl IntoIterator<Item = &usize>` but it starts to be weird).

Philippe-Cholet (changes requested):
Away from the keyboard, I thought that `from_` methods usually are constructors (returns `Self`) which is not the case here. So `get_at` is a better name than my suggested `from_indices`.

I'll merge this once the requested change is done, thanks.

src/permutations.rs
kinto-b: Yeah I was questioning this one myself. I'll push a patch shortly

Philippe-Cholet (approved):

PR #925, opened 2024-04-27
Add k_smallest_relaxed and variants

adamreichold:
This implements the algorithm described [in](https://quickwit.io/blog/top-k-complexity) which consumes twice the amount of memory as the existing `k_smallest` algorithm but achieves linear time in the number of elements in the input.

I expect this to fail the MSRV job as `select_nth_unstable` was stabilized in 1.49 and is therefore not available in 1.43.1. I decided to propose it anyway for the time when an MSRV increase is planned or if deemed sufficiently useful as reason for one.

### DIFF
--- a/src/k_smallest.rs
+++ b/src/k_smallest.rs
@@ -88,6 +88,45 @@ where
     storage
 }
 
+pub(crate) fn k_smallest_relaxed_general<I, F>(iter: I, k: usize, mut comparator: F) -> Vec<I::Item>
+where
+    I: Iterator,
+    F: FnMut(&I::Item, &I::Item) -> Ordering,
+{
+    if k == 0 {
+        iter.last();
+        return Vec::new();
+    }
+
+    let mut iter = iter.fuse();
+    let mut buf = iter.by_ref().take(2 * k).collect::<Vec<_>>();
+
+    if buf.len() < k {
+        buf.sort_unstable_by(&mut comparator);
+        return buf;
+    }
+
+    buf.select_nth_unstable_by(k - 1, &mut comparator);
+    buf.truncate(k);
+
+    iter.for_each(|val| {
+        if comparator(&val, &buf[k - 1]) != Ordering::Less {
+            return;
+        }
+
+        buf.push(val);
+
+        if buf.len() == 2 * k {
+            buf.select_nth_unstable_by(k - 1, &mut comparator);
+            buf.truncate(k);
+        }
+    });
+
+    buf.sort_unstable_by(&mut comparator);
+    buf.truncate(k);
+    buf
+}
+
 #[inline]
 pub(crate) fn key_to_cmp<T, K, F>(mut key: F) -> impl FnMut(&T, &T) -> Ordering
 where
--- a/src/k_smallest.rs
+++ b/src/k_smallest.rs
@@ -114,6 +114,7 @@ where
             return;
         }
 
+        assert_ne!(buf.len(), buf.capacity());
         buf.push(val);
 
         if buf.len() == 2 * k {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3153,6 +3153,105 @@ pub trait Itertools: Iterator {
         self.k_smallest_by(k, k_smallest::key_to_cmp(key))
     }
 
+    /// Sort the k smallest elements into a new iterator, in ascending order, relaxing the amount of memory required.
+    ///
+    /// **Note:** This consumes the entire iterator, and returns the result
+    /// as a new iterator that owns its elements.  If the input contains
+    /// less than k elements, the result is equivalent to `self.sorted()`.
+    ///
+    /// This is guaranteed to use `2 * k * sizeof(Self::Item) + O(1)` memory
+    /// and `O(n + k log k)` time, with `n` the number of elements in the input,
+    /// meaning it uses more memory than the minimum obtained by [`k_smallest`](Itertools::k_smallest)
+    /// but achieves linear time in the number of elements.
+    ///
+    /// The sorted iterator, if directly collected to a `Vec`, is converted
+    /// without any extra copying or allocation cost.
+    ///
+    /// **Note:** This is functionally-equivalent to `self.sorted().take(k)`
+    /// but much more efficient.
+    ///
+    /// ```
+    /// use itertools::Itertools;
+    ///
+    /// // A random permutation of 0..15
+    /// let numbers = vec![6, 9, 1, 14, 0, 4, 8, 7, 11, 2, 10, 3, 13, 12, 5];
+    ///
+    /// let five_smallest = numbers
+    ///     .into_iter()
+    ///     .k_smallest_relaxed(5);
+    ///
+    /// itertools::assert_equal(five_smallest, 0..5);
+    /// ```
+    #[cfg(feature = "use_alloc")]
+    fn k_smallest_relaxed(self, k: usize) -> VecIntoIter<Self::Item>
+    where
+        Self: Sized,
+        Self::Item: Ord,
+    {
+        self.k_smallest_relaxed_by(k, Ord::cmp)
+    }
+
+    /// Sort the k smallest elements into a new iterator using the provided comparison, relaxing the amount of memory required.
+    ///
+    /// The sorted iterator, if directly collected to a `Vec`, is converted
+    /// without any extra copying or allocation cost.
+    ///
+    /// This corresponds to `self.sorted_by(cmp).take(k)` in the same way that
+    /// [`k_smallest_relaxed`](Itertools::k_smallest_relaxed) corresponds to `self.sorted().take(k)`,
+    /// in both semantics and complexity.
+    ///
+    /// ```
+    /// use itertools::Itertools;
+    ///
+    /// // A random permutation of 0..15
+    /// let numbers = vec![6, 9, 1, 14, 0, 4, 8, 7, 11, 2, 10, 3, 13, 12, 5];
+    ///
+    /// let five_smallest = numbers
+    ///     .into_iter()
+    ///     .k_smallest_relaxed_by(5, |a, b| (a % 7).cmp(&(b % 7)).then(a.cmp(b)));
+    ///
+    /// itertools::assert_equal(five_smallest, vec![0, 7, 14, 1, 8]);
+    /// ```
+    #[cfg(feature = "use_alloc")]
+    fn k_smallest_relaxed_by<F>(self, k: usize, cmp: F) -> VecIntoIter<Self::Item>
+    where
+        Self: Sized,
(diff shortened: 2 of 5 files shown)


src/k_smallest.rs
```diff
@@ -88,6 +88,45 @@ where
...
+
+    buf.select_nth_unstable_by(k - 1, |a, b| comparator(a, b));
+    buf.truncate(k);
+
+    iter.for_each(|val| {
+        if comparator(&val, &buf[k - 1]) != Ordering::Less {
+            return;
+        }
+
+        buf.push(val);
```
Philippe-Cholet, line 117: Before `push`, it might be valuable to add something like `assert!(buf.len() < 2 * k /*or buf.capacity()*/);` to ensure that the compiler cut part of `push` code (we know `buf` does not need to grow, and less code makes things faster as the compiler can optimize things).
It seems we don't have a benchmark for `k_smallest` but such benchmark would make it clear if it's valuable or not.

src/k_smallest.rs
```diff
@@ -88,6 +88,45 @@ where
...
+
+    let mut iter = iter.fuse();
+    let mut buf = iter.by_ref().take(2 * k).collect::<Vec<_>>();
+
+    if buf.len() < k {
+        buf.sort_unstable_by(comparator);
+        return buf;
+    }
+
+    buf.select_nth_unstable_by(k - 1, |a, b| comparator(a, b));
```
Philippe-Cholet, line 109: `&mut comparator` should be enough instead of the closure. Same below.

src/lib.rs
```diff
@@ -3060,6 +3060,105 @@ pub trait Itertools: Iterator {
...
+    ///
+    /// itertools::assert_equal(five_smallest, 0..5);
+    /// ```
+    #[cfg(feature = "use_alloc")]
+    fn k_smallest_relaxed(self, k: usize) -> VecIntoIter<Self::Item>
+    where
+        Self: Sized,
+        Self::Item: Ord,
+    {
+        self.k_smallest_by(k, Ord::cmp)
```
Philippe-Cholet, line 3098: Name does not contain `relaxed` which seems to be a mistake. Same two functions below.

Philippe-Cholet (commented):
I did not check the docs yet. I would if we decide to eventually add this.
As you mentionned, it breaks the MSRV a little so it won't be added right now but the MSRV should become 1.51 soon enough.
Good first iteration. Thanks for adding tests.

I appreciate the idea here and _it sure makes sense to use `select_nth_unstable_by`_ which I was not familiar with.
Complexity sure seems interesting. I however wonder what is the performance difference with the un-relaxed versions in practice. A compared benchmark might be valuable, we don't have one yet so I think [this](https://bheisler.github.io/criterion.rs/book/user_guide/comparing_functions.html) might help if you want to add one.

I note that we would have 12 methods... `k_(smallest|largest)[_relaxed][_by[_key]]` which is a lot!

src/k_smallest.rs
adamreichold: In Tantivy - where the benchmarks from the linked article come from - this was indeed a measurable improvement, but the `assert` was insufficient and unsafe code was required to actually achieve the desired results, c.f.

https://github.com/quickwit-oss/tantivy/blob/99a59ad37e833e1129f7e1abd25738a0adfae73a/src/collector/top_score_collector.rs#L820-L830

Would you be open to such an implementation here?

adamreichold:
> I note that we would have 12 methods... k_(smallest|largest)[_relaxed][_by[_key]] which is a lot!

I agree and if pressed to make a choice, I would argue that the relaxed would be a reasonable general choice, i.e. could completely replace the heap-based versions.

For one, `k` usually is bounded by small fixed numbers and I do not really see a lot of situations where I have the memory to compute the top 100 but not the enough to invest into the top 200. And since the implementation can directly work with `select_nth_unstable_by`, the complications of having a custom heap implementation could be avoided.

However, making this choice does not appear obviously better than having twelve methods though.

Alternatively, the combinatorics could be reigned in by adding a "strategy-like" type parameter, e.g. `Strict` and `Relaxed` to indicate the desired memory bound.

Finally, I am not sure the `k_largest*` methods really carry their weight now that the `k_smallest_by` and `k_smallest_by_key` methods exist as in applications of these methods that I am aware of, adapting the way items are sorted in required in any case and hence choosing ascending or descending order does not really increase call site complexity.

Philippe-Cholet:
@phimuemue @jswrenn I don't know if you would agree on the current PR but have 12 similar methods seems a bit much to me. `k_smallest` variants are in itertools 0.13.0 which is _not released yet_, so we could still change some things without breaking and avoid the 12 methods. We could have

    iter.top_k(k).smallest() // redirected from the current k_smallest (that we could deprecate)
    iter.top_k(k).smallest_by(f)
    iter.top_k(k).smallest_by_key(key)
    iter.top_k(k).largest()
    iter.top_k(k).largest_by(f)
    iter.top_k(k).largest_by_key(key)
    And once MSRV >= 1.49, this PR would introduce:
    iter.top_k(k).relaxed().smallest()
    iter.top_k(k).relaxed().smallest_by(f)
    iter.top_k(k).relaxed().smallest_by_key(key)
    iter.top_k(k).relaxed().largest()
    iter.top_k(k).relaxed().largest_by(f)
    iter.top_k(k).relaxed().largest_by_key(key)

<details><summary>Implementation draft (click to see)</summary>

```rust
trait Itertools: Iterator {
    fn top_k(self, k: usize) -> Top<Self> { ... }
    #[deprecate...]
    fn k_smallest(self, k: usize) -> ... { self.top_k().smallest() }
}

pub struct GeneralTop<I, A> {
    iter: I,
    k: usize,
    marker: PhantomData<A>,
}
pub type Top<I> = GeneralTop<I, Unrelaxed>;
pub type TopRelaxed<I> = GeneralTop<I, Relaxed>;

impl<I: Iterator> Top<I> {
    pub fn relaxed(self) -> TopRelaxed<I> { ... }
}
impl<I: Iterator, A> GeneralTop<I, A> {
    pub fn smallest(self) -> ...
    pub fn smallest_by<F: FnMut...>(self, f: F) -> ...
    pub fn smallest_by_key<F: FnMut...>(self, key: F) -> ...
    pub fn largest(self) -> ...
    pub fn largest_by<F: FnMut...>(self, f: F) -> ...
    pub fn largest_by_key<F: FnMut...>(self, key: F) -> ...
}
```
The relaxed version would be added once the MSRV is big enough.
</details>

"Minimal allocation" vs "linear complexity" are both valuable tradeoffs here, but maybe it would be nice to not multiply methods.

src/k_smallest.rs
Philippe-Cholet: I went to source and `push` does `if self.len == self.buf.capacity() { reserve_for_push }` so maybe `assert_ne!(buf.len(), buf.capacity());` but `assert!(buf.len() < buf.capacity())` should really be enough.
About unsafe, you can read this https://github.com/rust-itertools/itertools/pull/560#issuecomment-2025715434 . So I would rather avoid unsafe code anywhere I can and I really think it should be possible here.

Philippe-Cholet: It can be improved later if an `assert` really is not enough.

adamreichold: I will give the assert a try when I add the benchmarks. But only later tonight when I hopefully have some spare time.

adamreichold: The `assert_ne` variant measurably improves the benchmark results. I therefore included it in the second commit.

adamreichold:
Looking at the results of the benchmark pushed as part of the second commit, I would summarize them as no difference for the ascending/best case

grafik

a small but consistent lead in the random/average case

grafik

and a large improvement for the descending/worst case

grafik

phimuemue:
Hi there, nice algorithm. Unsure about what to do best. And the algorithm begs the question of the scaling factor: Could a capacity of 3*k or 4*k even be more performant?

As for the 12 methods. I think that - if we decide to keep all of them - we should just offer 12 methods right on `Itertools`. I still feel the same for our grouping_map thing: It adds an indirection, and - from what I've seen so far - does not help in composability or reusability (see our trials to extend it to custom hashers). Sure, it occupies lots of space in docs.rs, but I do not see a real problem with that.

adamreichold:
> Sidenote: What's the unstable/stable behavior of the current implementation?

It is unstable, at least because it calls into the unstable selection/sorting routines from `std`.

adamreichold:
> And the algorithm begs the question of the scaling factor: Could a capacity of 3k or 4k even be more performant?

The [original issue](https://github.com/quickwit-oss/tantivy/issues/298#issue-322365190) and I think I agree with the assessment there: Further increasing the capacity is not too interesting as it will only change the constant factor attached to the `O(n)` part of the runtime, i.e. I don't think it is worth it complicating the API by giving this choice to the calling code.

Philippe-Cholet:
Ok about having 12 methods. The scaling factor 2 is fine.

TODO:
- Close this OR add this as an alternative OR replace our heap-based versions with it.
  Personally I would choose to keep this as an alternative as both tradeoffs are valuable enough?
- A bit unsure about the name: change "relaxed"?
  OR should we make the relaxed versions the default and the heap-based ones the ones with a longer name instead? It might be a better choice perf-wise but do so would be a breaking change IMHO.
- Review docs.
- Wait a bit for MSRV>=1.49 to merge this.

adamreichold:
> A bit unsure about the name: change "relaxed"?

Note that I am very much _not_ invested in that name and made it up on the spot when I had to name the methods somehow to avoid conflicting with the existing ones. I think the only requirement for the name is that they should reflect the space-time trade-off somehow.

src/lib.rs
```diff
@@ -3060,6 +3060,105 @@ pub trait Itertools: Iterator {
...
+    ///
+    /// ```
+    /// use itertools::Itertools;
+    ///
+    /// // A random permutation of 0..15
+    /// let numbers = vec![6, 9, 1, 14, 0, 4, 8, 7, 11, 2, 10, 3, 13, 12, 5];
+    ///
+    /// let five_smallest = numbers
+    ///     .into_iter()
+    ///     .k_smallest_by_key(5, |n| (n % 7, *n));
```
Philippe-Cholet, line 3148: Missing `_relaxed`.

Philippe-Cholet (commented):
I've been thinking about changing "relaxed" but the only alternative I found is "linear" which I kinda find worse (compare docs is necessary to understand anyway) so I'm fine with "relaxed".
I finally reviewed the docs (one small change to do).
Then I would be fine to merge this once our MSRV is big enough (probably later this year).

Philippe-Cholet (approved):
Thanks!
Now we wait.

Philippe-Cholet:
@adamreichold The long wait has ended, thanks again for this!

PR #927, opened 2024-05-01
Implement `FlattenOk::{fold, rfold}`

kinto-b:
Relates to #755 

```
$ cargo bench --bench specializations flatten_ok/fold -- --baseline flatten_ok

flatten_ok/fold         time:   [3.0078 µs 3.2944 µs 3.5502 µs]
                        change: [-47.164% -42.833% -38.023%] (p = 0.00 < 0.05)
                        Performance has improved.
```

Edit: `.rfold` is an almost identical implementation, so I'll put it through in a moment. Just running benchmarks...

Edit:

```
$ cargo bench --bench specializations flatten_ok/rfold -- --baseline flatten_ok

flatten_ok/rfold        time:   [2.7442 µs 2.8254 µs 2.9134 µs]
                        change: [-42.924% -39.835% -36.485%] (p = 0.00 < 0.05)
                        Performance has improved.
```

### DIFF
--- a/src/flatten_ok.rs
+++ b/src/flatten_ok.rs
@@ -72,6 +72,29 @@ where
         }
     }
 
+    fn fold<B, F>(self, init: B, mut f: F) -> B
+    where
+        Self: Sized,
+        F: FnMut(B, Self::Item) -> B,
+    {
+        // Front
+        let mut acc = match self.inner_front {
+            Some(x) => x.fold(init, |a, o| f(a, Ok(o))),
+            None => init,
+        };
+
+        acc = self.iter.fold(acc, |acc, x| match x {
+            Ok(it) => it.into_iter().fold(acc, |a, o| f(a, Ok(o))),
+            Err(e) => f(acc, Err(e)),
+        });
+
+        // Back
+        match self.inner_back {
+            Some(x) => x.fold(acc, |a, o| f(a, Ok(o))),
+            None => acc,
+        }
+    }
+
     fn size_hint(&self) -> (usize, Option<usize>) {
         let inner_hint = |inner: &Option<T::IntoIter>| {
             inner
--- a/src/flatten_ok.rs
+++ b/src/flatten_ok.rs
@@ -153,6 +153,29 @@ where
             }
         }
     }
+
+    fn rfold<B, F>(self, init: B, mut f: F) -> B
+    where
+        Self: Sized,
+        F: FnMut(B, Self::Item) -> B,
+    {
+        // Back
+        let mut acc = match self.inner_back {
+            Some(x) => x.rfold(init, |a, o| f(a, Ok(o))),
+            None => init,
+        };
+
+        acc = self.iter.rfold(acc, |acc, x| match x {
+            Ok(it) => it.into_iter().rfold(acc, |a, o| f(a, Ok(o))),
+            Err(e) => f(acc, Err(e)),
+        });
+
+        // Front
+        match self.inner_front {
+            Some(x) => x.rfold(acc, |a, o| f(a, Ok(o))),
+            None => acc,
+        }
+    }
 }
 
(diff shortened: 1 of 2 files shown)


Philippe-Cholet (changes requested):
How can it pass tests without using the fields `inner_front` and `inner_back`?! That's troubling!

Once `fold` is done, could you also specialize `rfold` in the same PR?

kinto-b:
@Philippe-Cholet I was just looking back over the code wondering that myself! Investigating now...

kinto-b:
Ah so the tests fail to capture what's wrong with this code because they use `Option<u8>` values for each Ok item in the iterator. We need non-singleton values. 

There is a comment there about it being too slow to use `Vec<u8>` but that's what's necessary to get proper coverage. Are you happy for me to make that change, i.e. `Option<u8>` to `Vec<u8>` in the below and alter the comment?

https://github.com/rust-itertools/itertools/blob/dd6a56932141c8b18b9d479eba30134c5afbfb88/tests/specializations.rs#L455-L456

Philippe-Cholet:
Ah I think I wrote this comment. Use `Vec<u8>` here is really slow, it's not really an option. So if we can't give `[u8; 3]` instead of `Option<u8>` then maybe we can write our own limited vector (what a pain) and implement the necessary traits for this to work. I'm gonna investigate this.

Philippe-Cholet:
So we basically need a light small vector, here is one with max 2 elements. The `flatten_ok` test pass under 1 second for me.
```rust
use quickcheck::Arbitrary;
use rand::Rng;
...
    // remove comment
    fn flatten_ok(v: Vec<Result<SmallIter2<u8>, char>>) -> () {
...

/// Like `VecIntoIter<T>` with maximum 2 elements.
#[derive(Debug, Clone, Default)]
enum SmallIter2<T> {
    #[default]
    Zero,
    One(T),
    Two(T, T),
}

impl<T: Arbitrary> Arbitrary for SmallIter2<T> {
    fn arbitrary<G: quickcheck::Gen>(g: &mut G) -> Self {
        match g.gen_range(0u8, 3) {
            0 => Self::Zero,
            1 => Self::One(T::arbitrary(g)),
            2 => Self::Two(T::arbitrary(g), T::arbitrary(g)),
            _ => unreachable!(),
        }
    }
}

impl<T> Iterator for SmallIter2<T> {
    type Item = T;

        match std::mem::take(self) {
            Self::Zero => None,
            Self::One(val) => Some(val),
            Self::Two(val, second) => {
                *self = Self::One(second);
                Some(val)
            }
        }
    }

        let len = match self {
            Self::Zero => 0,
            Self::One(_) => 1,
            Self::Two(_, _) => 2,
        };
        (len, Some(len))
    }
}

impl<T> DoubleEndedIterator for SmallIter2<T> {
        match std::mem::take(self) {
            Self::Zero => None,
            Self::One(val) => Some(val),
            Self::Two(first, val) => {
                *self = Self::One(first);
                Some(val)
            }
        }
    }
}
```

Philippe-Cholet:
I thought of `ArrayVec<T, N>` but I would have to wrap it anyway to implement `quickcheck::Arbitrary` (and therefore `Iterator`s too). I would have added it if we wanted more control than a fixed max-length of 2.

Philippe-Cholet:
@kinto-b Just add this in "tests/specializations.rs" where it's needed. We can move it later if we need to.

Sidenote: For CI to pass again (because of recent Clippy 1.78), you will need to _rebase_ your branch on master.

Philippe-Cholet (approved):
Looks good to me.
The specialization test did slow us down. An `Option` seemed good enough when I wrote it.
I guess the benchmark is up-to-date. -40% is nice enough.

Thanks again!

kinto-b:
Thanks!

PR #932, opened 2024-05-08
`assert_equal`: fix `clippy::default_numeric_fallback`

Philippe-Cholet:
Type of `i` defaults to `i32` when `usize` should be used.
It's only a problem if this function is called with two iterators longer than `i32::MAX`. It seems unlikely to me but who knows?
`i32` should not be the default integer type in the library. `usize` would have more sense so **_I add the lint as warning_** (for the library only, it's okay in tests and benchmarks).

See lint doc: https://rust-lang.github.io/rust-clippy/master/index.html#/default_numeric_fallback


jswrenn (approved):
Good catch.

PR #945, opened 2024-05-23
Prevent exponentially large type sizes in `tuple_combinations`

lcnr:
This slightly improves the performance of `cargo build --release` on stable, and prevents a hang[^1]  when using the next-generation trait solver `-Znext-solver`.

Long term such manually optimizations should be unnecessary, but for now it unfortunately requires some manual effort.

[^1]: it would finish at some point, but the sun will have probably consumed the earth before then

### DIFF
--- a/src/adaptors/mod.rs
+++ b/src/adaptors/mod.rs
@@ -773,16 +773,28 @@ macro_rules! impl_tuple_combination {
             where
                 F: FnMut(B, Self::Item) -> B,
             {
+                // We outline this closure to prevent it from unnecessarily
+                // capturing the type parameters `I`, `B`, and `F`. Not doing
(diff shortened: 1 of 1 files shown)


Philippe-Cholet (approved):
I wrote this `fold` specialization. It's sad we need such work but it's fine by me.

lcnr:
after fixing #125460 the following causes a compiler error before this PR:
```
use itertools::Itertools;

fn main() {
    (0..).tuple_combinations().fold(1, |_,  (t0, t1, t2, t3, t4, t5, t6, t7, t8, t9, t10, t11)| 1);
}
```

Philippe-Cholet:
Not sure what I could do here.
To be fair, call `fold` on an infinite iterator can't end well and probably shouldn't.

Issue #947, opened 2024-05-27
Implement `DoubleEndedIterator` for `FilterOk`


Xenira:
Should be doable with `rfind` unless there is a reason I overlooked.

Would be willing to implement this.

Philippe-Cholet:
`FlattenOk`, `MapSpecialCase` (for `MapOk`, `MapInto`) are double ended and `ProcessResults` since a few weeks.
`FilterOk` and `FilterMapOk` are not (yet) but it seems reasonable to me.

Xenira:
Like that? Have to admit, not entirely comprehending the test system yet.

Philippe-Cholet:
@Xenira I did not notice they were not double ended so thanks for this.

PR #948, opened 2024-05-27
feat(`FilterOk`): implement `DoubleEndedIterator`

Xenira:
Implemented `DoubleEndedIterator` for `FilterOk`

Refs: #947


Xenira:
Added the `rfold` impl

Philippe-Cholet (approved):
Good. Unless @phimuemue or @jswrenn think we should not make `FilterOk` double ended (and `FilterMapOk`), I'll merge this as is.

PR #954, opened 2024-05-31
Const generic permutations

FlareFlo:
In an application of mine i use a fixed size of permutations, which would cut down on some very wasteful temporary allocations. Therefore I am proposing a const-generic counterpart to the regular permutations.

This drafts purpose is to get some initial feedback on the idea.
Do not expect feature completeness or sanity, it is 3AM. Though I do intend to properly implement this if interest exists.

### DIFF
--- a/src/permutations_const.rs
+++ b/src/permutations_const.rs
@@ -1,0 +1,188 @@
+use alloc::boxed::Box;
+use alloc::vec::Vec;
+use std::fmt;
+use std::iter::once;
+use std::iter::FusedIterator;
+use std::convert::TryInto;
+
+use super::lazy_buffer::LazyBuffer;
+use crate::size_hint::{self, SizeHint};
+
+/// An iterator adaptor that iterates through all the `k`-permutations of the
+/// elements from an iterator.
+///
+/// See [`.permutations()`](crate::Itertools::permutations) for
+/// more information.
+#[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
+pub struct PermutationsConst<I: Iterator, const K: usize> {
+    vals: LazyBuffer<I>,
+    state: PermutationState<K>,
+}
+
+impl<I, const K: usize> Clone for PermutationsConst<I, K>
+where
+    I: Clone + Iterator,
+    I::Item: Clone,
+{
+    clone_fields!(vals, state);
+}
+
+#[derive(Clone, Debug)]
+enum PermutationState<const K: usize> {
+    /// No permutation generated yet.
+    Start,
+    /// Values from the iterator are not fully loaded yet so `n` is still unknown.
+    Buffered { min_n: usize },
+    /// All values from the iterator are known so `n` is known.
+    Loaded {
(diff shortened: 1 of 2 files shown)


src/permutations_const.rs
```diff
@@ -0,0 +1,188 @@
...
+    type Item = [I::Item; K];
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let Self { vals, state } = self;
+        match state {
+            PermutationState::Start => {
+                *state = PermutationState::End;
+                // TODO: Consider this case and handle it somehow, currently just using default
+                Some(std::array::from_fn(|_|I::Item::default()))
+            }
```
FlareFlo, line 74: Not sure how to proceed about this, from what i can tell the API guarantees the vector to always be K elements long, but the reference returns an empty vec?

src/permutations_const.rs
```diff
@@ -0,0 +1,188 @@
...
+                    *state = PermutationState::Loaded { indices, cycles };
+                    Some(item.try_into().ok()?) // TODO: Handle error case
+                }
+            }
+            PermutationState::Loaded { indices, cycles } => {
+                if advance(indices, cycles) {
+                    *state = PermutationState::End;
+                    return None;
+                }
+                let k = cycles.len();
```
FlareFlo, line 114: Indexing into LazyBuffer would require a const-generic function also, otherwise this effort would be pointless

Philippe-Cholet:
Unless the lexicographic order of the permutations is important for you, the [permutohedron](https://crates.io/crates/permutohedron) crate should plainly satify you (created by **bluss** who first wrote **itertools**).

Now, some context... Right now, our MSRV is 1.43.1 while const-generics require 1.51.0. We will move to it in a near future though. But `core::array::from_fn` requires 1.63.0 and `core::array::try_from_fn` is still nightly, same for [`Iterator::next_chunk`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.next_chunk). So we are missing some helpful bricks at the moment.

This will eventually be done along with combinations & Co. but not immediately. #560 would be a more intermediary step.

FlareFlo:
> Unless the lexicographic order of the permutations is important for you, the [permutohedron](https://crates.io/crates/permutohedron) crate should plainly satify you (created by **bluss** who first wrote **itertools**). It should even be faster than a const-generics version of our permutations, due to the algorithm.
> 
> Now, some context... Right now, our MSRV is 1.43.1 while const-generics require 1.51.0. We will move to it in a near future though. But `core::array::from_fn` requires 1.63.0 and `core::array::try_from_fn` is still nightly, same for [`Iterator::next_chunk`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.next_chunk). So we are missing some helpful bricks at the moment.
> 
> This will eventually be done along with combinations & Co. but not immediately. #560 would be a more intermediary step.

permutohedron covers my case, thanks!

Seeing that the MSRV is restricted like that, should I close this?

PR #955, opened 2024-06-09
improve documentation of tree_reduce

### LINKED ISSUE #946: How does tree_reduce achieve O(ln(n)) operations?
How does `Itertools::tree_reduce` achieve `O(ln(n))` operations? 

If we take the visualization from the documentation:

```
1 2 3 4 5 6 7
│ │ │ │ │ │ │
└─f └─f └─f │
  │   │   │ │
  └───f   └─f
      │     │
      └─────f
```

This looks like a full binary tree with `n` being the number of leaf nodes and the operations `f` being the internal nodes. A binary tree with `I` internal nodes has `L = I + 1` leaf nodes (if I'm not mistaken). That would mean the number of operations is always `n - 1`, which would be `O(n)` and the same as the normal `Iterator::reduce` operation.

What am I missing?

rbruenig:
PR for improving documentation of `tree_reduce`, it will close #946

This PR only changes documentation and does not touch any actual code.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2528,24 +2528,26 @@ pub trait Itertools: Iterator {
     ///
     /// If `f` is associative you should also decide carefully:
     ///
-    /// - if `f` is a trivial operation like `u32::wrapping_add`, prefer the normal
-    /// [`Iterator::reduce`] instead since it will most likely result in the generation of simpler
-    /// code because the compiler is able to optimize it
-    /// - otherwise if `f` is non-trivial like `format!`, you should use `tree_reduce` since it
-    /// reduces the number of operations from `O(n)` to `O(ln(n))`
+    /// For an iterator producing `n` elements, both [`Iterator::reduce`] and `tree_reduce` will
+    /// call `f` `n - 1` times. However, `tree_reduce` will call `f` on earlier intermediate
+    /// results, which is beneficial for `f` that allocate and produce longer results for longer
+    /// arguments. For example if `f` combines arguments using `format!`, then `tree_reduce` will
+    /// operate on average on shorter arguments resulting in less bytes being allocated overall.
     ///
-    /// Here "non-trivial" means:
-    ///
-    /// - any allocating operation
-    /// - any function that is a composition of many operations
+    /// If 'f' does not benefit from such a reordering, like `u32::wrapping_add`, prefer the
+    /// normal [`Iterator::reduce`] instead since it will most likely result in the generation of
+    /// simpler code because the compiler is able to optimize it.
     ///
     /// ```
     /// use itertools::Itertools;
     ///
+    /// let f = |a: String, b: String| {
+    ///     format!("f({a}, {b})")
+    /// };
+    ///
     /// // The same tree as above
-    /// let num_strings = (1..8).map(|x| x.to_string());
-    /// assert_eq!(num_strings.tree_reduce(|x, y| format!("f({}, {})", x, y)),
-    ///     Some(String::from("f(f(f(1, 2), f(3, 4)), f(f(5, 6), 7))")));
+    /// assert_eq!((1..8).map(|x| x.to_string()).tree_reduce(f),
+    ///            Some(String::from("f(f(f(1, 2), f(3, 4)), f(f(5, 6), 7))")));
     ///
     /// // Like fold1, an empty iterator produces None
     /// assert_eq!((0..0).tree_reduce(|x, y| x * y), None);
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -2528,34 +2528,63 @@ pub trait Itertools: Iterator {
...
+    ///         r
+    ///     })
+    ///     .unwrap();
+    ///
+    /// dbg!(total_capacity_reduce, total_capacity_tree_reduce,
+    ///      reduce_res.len(), tree_reduce_res.len());
+    /// // total_capacity_reduce = 65630
+    /// // total_capacity_tree_reduce = 7284
+    /// // reduce_res.len() = 679
+    /// // tree_reduce_res.len() = 679
```
Philippe-Cholet, line 2587: We usually don't use `dbg!` but assertions. These comments could silently be outdated later. But I guess there is no guarantees on allocated capacities for such example? I'm not sure if it's better to have assertions that might fail in the future because of changed allocations in future Rust.

src/lib.rs
```diff
@@ -2528,34 +2528,63 @@ pub trait Itertools: Iterator {
...
     /// If `f` is associative you should also decide carefully:
     ///
-    /// - if `f` is a trivial operation like `u32::wrapping_add`, prefer the normal
-    /// [`Iterator::reduce`] instead since it will most likely result in the generation of simpler
-    /// code because the compiler is able to optimize it
-    /// - otherwise if `f` is non-trivial like `format!`, you should use `tree_reduce` since it
-    /// reduces the number of operations from `O(n)` to `O(ln(n))`
+    /// For an iterator producing `n` elements, both [`Iterator::reduce`] and `tree_reduce` will
+    /// call `f` `n - 1` times. However, `tree_reduce` will call `f` on earlier intermediate
+    /// results, which is beneficial for `f` that allocate and produce longer results for longer
```
Philippe-Cholet, line 2533: I'm not really convinced by the formulation "_on earlier intermediate results_" but I don't have an alternative right now.

src/lib.rs
```diff
@@ -2528,34 +2528,63 @@ pub trait Itertools: Iterator {
     ///
     /// If `f` is associative you should also decide carefully:
     ///
-    /// - if `f` is a trivial operation like `u32::wrapping_add`, prefer the normal
-    /// [`Iterator::reduce`] instead since it will most likely result in the generation of simpler
-    /// code because the compiler is able to optimize it
-    /// - otherwise if `f` is non-trivial like `format!`, you should use `tree_reduce` since it
-    /// reduces the number of operations from `O(n)` to `O(ln(n))`
```
Philippe-Cholet, line 2535: Maybe we should keep something about `O(ln(n))` (height of the tree) vs `O(n)`.

src/lib.rs
jswrenn: Perhaps we can create a mock, hidden version of `String` , here? I don't want us to have fragile assertions.

src/lib.rs
rbruenig: I also thought about the wording here for a while and wasn't super happy with this. But its the best that I could some up with that transports the idea. Open for any suggestions.

src/lib.rs
rbruenig: I thought about that and decided against it, because it seemed confusing. Is the information that the tree height is `O(ln(n))` relevant to the user of this API and why is it relevant?

src/lib.rs
rbruenig: Its probably useful to have some example that shows the possible runtime improvements of `tree_reduce` over `reduce`, because the text itself is quite abstract. But maybe we can use another example that doesn't involve `String` and is deterministic?

src/lib.rs
rbruenig: That's a really good point. Its not about a performance benefit but about the result of the operation being different. I like the binary tree construction idea, I'll add something about that to the doc. In that context the `O(ln(n))` about the height is also clear.

src/lib.rs
rbruenig: I changed the example to using `.len()` instead of `.capacity()` which is deterministic and changed the `dbg!` to `assert!`. What do you think about that?

src/lib.rs
rbruenig: Added a paragraph about building a binary search tree that includes the remark about tree height.

src/lib.rs
Philippe-Cholet: Yeah, I thought `capacity` was somehow needed over `len` but it seems more appropriate.

rbruenig:
@Philippe-Cholet Is there anything that still should be improved or is this ready to be merged 🙂 ?

src/lib.rs
```diff
@@ -2528,34 +2528,64 @@ pub trait Itertools: Iterator {
...
     ///
-    /// Here "non-trivial" means:
+    /// Moreover, the output of `tree_reduce` is preferable to that of [`Iterator::reduce`] in
+    /// certain cases. For example, building a binary search tree using `tree_reduce` will result in
+    /// a balanced tree with height O(ln(n)), while [`Iterator::reduce`] will output a tree with
+    /// height O(n), essentially a linked list.
     ///
-    /// - any allocating operation
-    /// - any function that is a composition of many operations
+    /// If 'f' does not benefit from such a reordering, like `u32::wrapping_add`, prefer the
```
Philippe-Cholet, line 2542: nitpick: `f` instead of 'f'.

src/lib.rs
```diff
@@ -2528,34 +2528,64 @@ pub trait Itertools: Iterator {
...
+    /// call `f` `n - 1` times. However, `tree_reduce` will call `f` on earlier intermediate
+    /// results, which is beneficial for `f` that allocate and produce longer results for longer
+    /// arguments. For example if `f` combines arguments using `format!`, then `tree_reduce` will
+    /// operate on average on shorter arguments resulting in less bytes being allocated overall.
     ///
-    /// Here "non-trivial" means:
+    /// Moreover, the output of `tree_reduce` is preferable to that of [`Iterator::reduce`] in
+    /// certain cases. For example, building a binary search tree using `tree_reduce` will result in
+    /// a balanced tree with height O(ln(n)), while [`Iterator::reduce`] will output a tree with
+    /// height O(n), essentially a linked list.
```
Philippe-Cholet, line 2540: Another nitpick: `O(n)` and `O(ln(n))` within backsticks.

Philippe-Cholet (commented):
I checked out your branch and opened the doc to read it. The text is fine, only got 2 visual nitpicks. I'll merge once fixed, thanks for the remainder.

src/lib.rs
rbruenig: @Philippe-Cholet fixed

src/lib.rs
rbruenig: fixed

Philippe-Cholet (approved):

Philippe-Cholet:
Thanks for the needed clarification!

PR #961, opened 2024-06-29
ci: Run most tests with miri

jswrenn:
This PR additionally reduces the size of some heavy tests to reduce their running time. I don't think these reductions will hinder their bug-detecting power too much.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,6 +54,21 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
       - run: cargo test --all-features
 
+  miri:
+    runs-on: ubuntu-latest
+    env:
+      CARGO_TERM_COLOR: always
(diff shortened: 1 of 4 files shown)


Philippe-Cholet (commented):
I note that you use `nextest` which does not run doctests while we could want to run miri on some of them?

jswrenn:
Looks like this issue is tracked by https://github.com/nextest-rs/nextest/issues/16. I'm using nextest because it, unlike miri's default test runner, runs tests in parallel. The speedup is significant.

Until nextest gets doctest support, we'll just [run the doc tests separately](https://github.com/rust-itertools/itertools/compare/7b58706c4e8bcd246548603038d982cd651622ba..7a510b2c6a399aa710a7511d932b1b8e1c886054).

Philippe-Cholet (approved):

PR #971, opened 2024-07-12
Allow `Q: ?Sized` in `Itertools::contains`

Alexendoo:
To enable `.contains("foo")` on an `Iterator<Item = String>` and similar


phimuemue (approved):
Thanks.

PR #972, opened 2024-07-15
Give `take_while_inclusive` a more intuitive doc alias

Ciel-MC:
I could not for the life of me find this function lmao, it's called take until most places so I think an alias for searchability would be nice

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1594,6 +1594,7 @@ pub trait Itertools: Iterator {
     ///     .collect();
     /// let expected: Vec<_> = vec![1, 2, 3].into_iter().map(NoCloneImpl).collect();
     /// assert_eq!(filtered, expected);
+    #[doc(alias = "take_until")]
     fn take_while_inclusive<F>(self, accept: F) -> TakeWhileInclusive<Self, F>
     where
         Self: Sized,


Philippe-Cholet (approved):
It seems reasonable to me.

However, there is [`Iterator::take_while`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.take_while) so it does not come from nowhere. 😀

Philippe-Cholet:
@Ciel-MC Maybe `#[doc(alias = "TakeUntil")]` on `TakeWhileInclusive`?

jswrenn:
Seems reasonable to me!

Ciel-MC:
> However, there is [`Iterator::take_while`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.take_while) so it does not come from nowhere. 😀

I do understand that it didn't come from nowhere, but it is a bit like calling 2 "1+1", like yes it came from adding 1 to 1, but it has a pretty common name already.

> @Ciel-MC Maybe `#[doc(alias = "TakeUntil")]` on `TakeWhileInclusive`?

I mean, I guess that is just as searchable? Though I'm not sure why you'd want it on the struct instead of the method(excuse my inexperience with api design)

Philippe-Cholet:
It was a question. After searching doc aliases on rust-lang/rust, it sure does not alias structs much if at all, my bad.

Issue #983, opened 2024-08-07
Return first Ok or last Err


axelkar:
* Input iterator item: `Result<T, E>`
* Collected output: `Result<Option<T>, E>`

This "collector" should return the first `Ok` value it finds as `Ok(Some(value))`.
If no `Ok` values are found, return the last `Err`.
If no values are in the iterator, return `Ok(None)`

I've written this a couple of times in my own code and figured it'd be useful for Itertools to include.

```rs
fn find_first_ok<T, E, I>(iter: I) -> Result<Option<T>, E>
where
    I: IntoIterator<Item = Result<T, E>>,
{
    let mut last_err = None;

    for item in iter {
        match item {
            Ok(value) => return Ok(Some(value)),
            Err(err) => last_err = Some(err),
        }
    }

    match last_err {
        Some(err) => Err(err),
        None => Ok(None),
    }
}

#[test]
fn test_no_items() {
    assert_eq!(find_first_ok::<(), (), _>([]), Ok(None));
}
#[test]
fn test_only_err() {
    assert_eq!(find_first_ok::<(), u8, _>([
        Err(1),
        Err(2),
        Err(3),
    ]), Err(3));
}
#[test]
fn test_ok() {
    assert_eq!(find_first_ok::<(), u8, _>([
        Err(1),
        Ok(()),
        Err(2),
    ]), Ok(Some(())));
}
```

phimuemue:
If we include this, we should think about a generalized version, right? To me, it seems that this basically groups the iterator's items into two categories. If we find an item of the first category, return it right away. If there's no such item, return the last item of the second category. (Please correct me if I got that wrong.)

If that's true, I guess a natural API would be something like this:

```
 fn get_first_satisfying_predicate_or_last_not_satisfying_predicate(self, mut pred: impl FnMut(&Item)->bool) -> Option<Item> {
  let mut last_not_satisfying_predicate = None;

    for item in self {
        if pred(&item) {
         return item; 
        } else {
         last_not_satisfying_predicate = Some(item);
        }
    }
    return last_not_satisfying_predicate;
 }
```

That would allow to use it for iterators with arbitrary `Item`s.

axelkar:
> ... which makes me think: Isn't this [`find_or_last`](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.find_or_last)?

Yes:
```rs
fn find_first_ok<T, E, I>(iter: I) -> Result<Option<T>, E>
where
[...]

PR #984, opened 2024-08-07
Add note about Result to `find_or_last` and `find_or_first`

axelkar:
See GH-983

@phimuemue 

Is the wording okay? What about the explicit type definition?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1999,6 +1999,15 @@ pub trait Itertools: Iterator {
     /// assert_eq!(numbers.iter().find_or_last(|&&x| x > 5), Some(&4));
     /// assert_eq!(numbers.iter().find_or_last(|&&x| x > 2), Some(&3));
     /// assert_eq!(std::iter::empty::<i32>().find_or_last(|&x| x > 5), None);
+    ///
+    /// // An iterator of Results can return the first Ok or the last Err:
+    /// let input = vec![Err(()), Ok(11), Err(()), Ok(22)];
+    /// assert_eq!(input.into_iter().find_or_last(Result::is_ok), Some(Ok(11)));
+    ///
+    /// let input: Vec<Result<(), i32>> = vec![Err(11), Err(22)];
+    /// assert_eq!(input.into_iter().find_or_last(Result::is_ok), Some(Err(22)));
+    ///
+    /// assert_eq!(std::iter::empty::<Result<(), i32>>().find_or_last(Result::is_ok), None);
     /// ```
     fn find_or_last<P>(mut self, mut predicate: P) -> Option<Self::Item>
     where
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -2000,6 +2000,21 @@ pub trait Itertools: Iterator {
     /// assert_eq!(numbers.iter().find_or_last(|&&x| x > 2), Some(&3));
     /// assert_eq!(std::iter::empty::<i32>().find_or_last(|&x| x > 5), None);
     /// ```
+    ///
+    /// An iterator of [`Result`]s can find the first `Ok` or the last `Err`:
```
phimuemue, line 2004: I think we can make it less prominent. Maybe just add the following after the previous `assert_eq`s:

```
// find first Ok if present, or return the last Err:
assert_eq!(vec![Err(1), Ok(2), Err(3), Ok(4)].find_or_last(Result::is_ok), Ok(2));
assert_eq!(vec![Err(1), Err(2)].find_or_last(Result::is_ok), Err(2));
...
```

phimuemue:
> What about the explicit type definition?

We can use them where needed, otherwise elide them.

phimuemue:
@axelkar There are some CI errors. Ours should have been fixed by https://github.com/rust-itertools/itertools/commit/4f49f2636aa95283cc4c921e6a3e04efe317fda5. Could you rebase and fix yours if needed?

src/lib.rs
```diff
@@ -1999,6 +1999,15 @@ pub trait Itertools: Iterator {
     /// assert_eq!(numbers.iter().find_or_last(|&&x| x > 5), Some(&4));
     /// assert_eq!(numbers.iter().find_or_last(|&&x| x > 2), Some(&3));
     /// assert_eq!(std::iter::empty::<i32>().find_or_last(|&x| x > 5), None);
+    ///
+    /// // An iterator of Results can return the first Ok or the last Err:
+    /// let input = vec![Err(()), Ok(11), Err(()), Ok(22)];
+    /// assert_eq!(.into_iter().find_or_last(Result::is_ok), Some(Ok(11)));
```
phimuemue, line 2005: Should this be `assert_eq!(input.into_iter().find_or_last(Result::is_ok), Some(Ok(11)));`?

phimuemue:
Thanks a bunch.

src/lib.rs
```diff
@@ -2027,6 +2036,15 @@ pub trait Itertools: Iterator {
     /// assert_eq!(numbers.iter().find_or_first(|&&x| x > 5), Some(&1));
     /// assert_eq!(numbers.iter().find_or_first(|&&x| x > 2), Some(&3));
     /// assert_eq!(std::iter::empty::<i32>().find_or_first(|&x| x > 5), None);
+    ///
+    /// // An iterator of Results can return the first Ok or the last Err:
```
phimuemue, line 2040: This should be "or the first Err". Sorry for the back-and-forth, I did not realize it the first time.

phimuemue (changes requested):

src/lib.rs
axelkar: It's really something I overlooked. Thanks for reviewing!

phimuemue:
Thanks.

phimuemue (approved):

PR #985, opened 2024-08-12
Change an instance of `default_features` in `Cargo.toml` to `default-features`

LikeLakers2:
As per cargo itself when running a `cargo build`:

> `default_features` is deprecated in favor of `default-features` and will not work in the 2024 edition (in the `quickcheck` dependency)

Since `default_features` and `default-features` work the same, this should not change any functionality beyond removing the warning when building itertools. To be sure, however, I ran `cargo test` after the change, and all tests passed.


jswrenn (approved):
Thanks!

PR #988, opened 2024-08-22
ConsTuples is a MapSpecialCase

phimuemue:


phimuemue:
I'm unsure about the failing CI: Is the problem that `ConsTuples` required 2 generic arguments, whereas now it only needs 1? If so, I'd be tempted to accept this.

jswrenn (approved):
We already have a breaking change queued up anyways for our MSRV bump, so the breaking change in this PR is a non-issue.

PR #991, opened 2024-09-10
array_combinations using array::map

ronnodas:
This is an implementation of `array_combinations` that imitates `combinations` except using `[_; K]` (with `K` a const generic) for the indices and the item type instead of `Vec`. Uses `array::from_fn` and `array::map` although the former could be easily avoided.


src/combinations.rs
```diff
@@ -210,8 +209,17 @@ where
 {
 }
 
+pub(crate) fn n_and_count<I: Iterator>(
```
jswrenn, line 212: Can you provide documentation here?

jswrenn (commented):
Could you add specialization tests?

See https://github.com/rust-itertools/itertools/blob/91f96182488788e01a40742c5269a0a3d02da49f/tests/specializations.rs#L136

src/combinations.rs
ronnodas: Added documentation, although as far as I can tell the existence of that function (and not just `count` by itself) is to enable `powerset::count` so maybe there is a better refactoring here.

Also added the specialization tests, but I'm not really sure what that's doing so adapted the one for `tuple_combinations`.

ronnodas:
Note that `tuple_combinations` does not need to allocate but does need to clone the input iterator (and not just its elements) because it is using a different algorithm, in a way recursing on the length of tuple. I do not see a straightforward way to do that with const generics.

phimuemue:
Hi there, thanks for this.

As far as I can judge, this is pretty much a copy-paste of combinations.rs, right?

As such, I strongly suggest to try and generalize our existing combinations (that works on `Vec`) so that it can work on arrays, too. The algorithms are non-trivial and we should strive to have one source of truth.

In addition, I hope that we are able to not only generalize combinations to arrays, but also other algorithms.

@Philippe-Cholet You refactored lots of these algorithms. What's your take on this?

ronnodas:
Hi, thanks for the suggestion! Yes, it's basically copy-paste from combinations. 

I will refactor to share as much of the code as makes sense.

I also hope that there will be more array versions soon but I think most of the other clear candidates would be easy to implement once `array::try_from_fn` is available.

src/combinations.rs
```diff
@@ -6,45 +8,112 @@ use alloc::vec::Vec;
...
-pub struct Combinations<I: Iterator> {
-    indices: Vec<usize>,
+pub struct CombinationsGeneric<I: Iterator, Idx> {
+    indices: Idx,
     pool: LazyBuffer<I>,
     first: bool,
 }
 
-impl<I> Clone for Combinations<I>
+pub trait PoolIndex<T>: IndexMut<usize, Output = usize> {
```
phimuemue, line 42: We can play the supertrait game, but instead of requiring `IndexMut`, couldn't we go with `BorrowMut<[usize>>` and get both `as_slice` and (once we have the slice) `IndexMut` for free?

ronnodas: Sure, I don't see the need to be overly general here

src/combinations.rs
```diff
@@ -135,18 +180,18 @@ impl<I: Iterator> Combinations<I> {
...
+            indices[j] = indices[j - 1] + 1;
         }
-
         // If we've made it this far, we haven't run out of combos
         false
     }
 
     /// Returns the n-th item or the number of successful steps.
-    pub(crate) fn try_nth(&mut self, n: usize) -> Result<<Self as Iterator>::Item, usize>
+    pub(crate) fn try_nth(&mut self, n: usize) -> Result<Idx::Item, usize>
```
phimuemue, line 192: I'm not sure if `Idx::Item` is really better than `<Self as Iterator>::Item` here. Sure, it's less work to type it, but I think `try_nth` is a method on the iterator thing, so its result does not need to include the (inner) `Idx` type.

src/combinations.rs
```diff
@@ -203,15 +249,39 @@ where
...
     I::Item: Clone,
+    Idx: PoolIndex<I::Item>,
 {
 }
 
+impl<I: Iterator> Combinations<I> {
+    /// Resets this `Combinations` back to an initial state for combinations of length
+    /// `k` over the same pool data source. If `k` is larger than the current length
+    /// of the data pool an attempt is made to prefill the pool so that it holds `k`
+    /// elements.
```
phimuemue, line 264: Thanks for correcting the comment.

src/combinations.rs
```diff
@@ -203,15 +249,39 @@ where
...
+            }
+            self.indices.extend(self.indices.len()..k);
+            self.pool.prefill(k);
+        }
+    }
+}
+
 /// For a given size `n`, return the count of remaining combinations or None if it would overflow.
-fn remaining_for(n: usize, first: bool, indices: &[usize]) -> Option<usize> {
+pub(crate) fn remaining_for(n: usize, first: bool, indices: &[usize]) -> Option<usize> {
```
phimuemue, line 284: Is `pub(crate)` needed?

src/lib.rs
```diff
@@ -1671,6 +1671,52 @@ pub trait Itertools: Iterator {
         adaptors::tuple_combinations(self)
     }
 
+    /// Return an iterator adaptor that iterates over the combinations of the
+    /// elements from an iterator.
+    ///
+    /// Iterator element can be any array of type `Self::Item`.
```
phimuemue, line 1677: Please use "Iterator element type is `[Self::Item; K]`. The iterator produces a new array per iteration, and clones the iterator elements" (just like for the existing `combinations`).

phimuemue (commented):
Thank you, this is very nice. Could you please address my nitpicks?

@jswrenn After the nits have been fixed (or refused with reason), please have one final look, and if you're satisfied, I think we can merge this. 

Note: Honestly, I'm unsure about the day-to-day speedups achievable by this, but I think it serves as a starting point for our generics initiative.

src/combinations.rs
ronnodas: I didn't actually touch this comment, the function has just moved to a separate `impl` since it is only used by powerset and so only makes sense to have for variable length combinations.

ronnodas:
Thanks for the extremely helpful guidance!

ronnodas:
Is there something to do on my end about the semver check? I'm not sure why it's complaining about `ConsTuples`. `Combinations` did change from a `pub struct` to a `pub type` made of a more generic struct but I think with the same public API so I don't know if this counts as a breaking change.

phimuemue (approved):

phimuemue (approved):

phimuemue:
@jswrenn Do you agree we can include this?

src/combinations.rs
```diff
@@ -6,45 +8,99 @@ use alloc::vec::Vec;
...
-pub struct Combinations<I: Iterator> {
-    indices: Vec<usize>,
+pub struct CombinationsGeneric<I: Iterator, Idx> {
+    indices: Idx,
     pool: LazyBuffer<I>,
     first: bool,
 }
 
-impl<I> Clone for Combinations<I>
+pub trait PoolIndex<T>: BorrowMut<[usize]> {
```
jswrenn, line 42: Can you add doc comments to this trait?

jswrenn (changes requested):
This looks good to me! Just one remaining nit about doc comments.

PR #1000, opened 2024-12-05
Fix into_group_map_by documentation errors

docwilco:
This method returns a HashMap, not an Iterator.
Bonus whitespace fix in example.

See also the `into_group_map` documentation.


phimuemue:
Thank you.

phimuemue (approved):

PR #1004, opened 2024-12-14
Proposal: `ExactSizeItertools`

JonathanWoollett-Light:
Introduces a new trait based on `std::iter::ExactSizeIterator` that can enable new functionality.

I was making an application that uses something like `zip_squash` and thought it could be implemented here.


src/lib.rs
```diff
@@ -4541,6 +4545,82 @@ pub trait Itertools: Iterator {
...
+    /// ```ignore
+    /// use itertools::ExactSizeItertools;
+    /// let it = vec![0,1].iter().zip_squash(vec![1,2,3].iter());
+    /// itertools::assert_equal(it, vec![(0,1),(1,3)]);
+    /// ```
+    fn zip_squash<J>(self, other: J) -> ZipSquash<Self, J::IntoIter>
+    where
+        J: IntoIterator,
+        <J as IntoIterator>::IntoIter: ExactSizeIterator,
+        Self: Sized;
```
jswrenn, line 4569: Couldn't this method just be on `Itertools` with a `Self: ExactSizeIterator` bound? Why introduce a new trait?

JonathanWoollett-Light: Oh yeah I can do that. Updated to do that.

Issue #783, opened 2023-10-16
Support for custom hasher for std groupping


a14e:
# Overview
I want to suggest an improvement for the library -- the ability to use custom hashes (specifically interested in ahash).

# Why?
The standard library uses a sip hash, which is protected against various types of attacks on hash tables, but it is known to be quite slow. Alternative hashes can be 10+ times faster (for example, there is a benchmark example in [ahash](https://github.com/tkaitchuck/aHash))

# What difficulties might arise?
It's unclear how to implement this
1. Option 1. Keep the current implementation, but add a third argument with a default value, for example, the methods would look like this:
   ```rust
   #[cfg(feature = "use_std")]
       fn into_group_map<K, V, R = RandomState>(self) -> HashMap<K, Vec<V>, R>
       where
           Self: Iterator<Item = (K, V)> + Sized,
           K: Hash + Eq,
           R: BuildHasher + Default
       {
           group_map::into_group_map(self)
       }
   ```
   but the current Rust doesn’t know how to do this and writes something like
   ```rust
   error: defaults for type parameters are only allowed in `struct`, `enum`, `type`, or `trait` definitions
       --> src/lib.rs:3076:29
        |
   3076 |     fn into_group_map<K, V, R = RandomState>(self) -> HashMap<K, Vec<V>, R>
        |                             ^^^^^^^^^^^^^^^
        |
        = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
        = note: for more information, see issue #36887 <https://github.com/rust-lang/rust/issues/36887>
        = note: `#[deny(invalid_type_param_default)]` on by default
   ```
2. Option 2, we can try replacing all imports with a flag from std::collections::HashMap to ahash::HashMap, but this breaks backward compatibility

Maybe there are some other options?

a14e:
This issue seems to be a duplicate of [#322](https://github.com/rust-itertools/itertools/issues/322).

a14e:
Other possible options:
1. Generalize the current implementation and create duplicates of methods with custom hashers (for example, `HashMap` has a constructor `new` and `with_hasher`).
2. Create some kind of middleware, and call, for instance, `vec![1, 2, 3].iter().grouping::<ahash::HashMap>`, and implement all grouping options on top of this middleware. What do you think?

Philippe-Cholet:
As a maintainer or not, I certainly would like to customize the hasher too.
[...]

Issue #462, opened 2020-07-16
Specify hasher for unique?


camsteffen:
Would there be interest in adding a variant of `unique` which allows you to specify the hasher? How about the name `unique_default`? The name isn't great but it's the best I can think of. It is inspired by `HashMap` which allows you to specify the hasher using `HashMap::default()`.

**Example:**

```rust
 iter.unique_default::<BuildHasherDefault<AHasher>>();
```

I have implemented this on my machine but I want to get feedback before opening a PR.

phimuemue:
I could imagine that some users want to check uniqueness with more efficient containers, but shouldn't we at least think about generalizing a bit more so that users can essentially specify a container that can insert an element, returning if the element already has been in there or not (such as `BTreeSet`, or even sorted `Vec`s)?

camsteffen:
I don't know how it could be that generic. There's always `.filter(move |item| set.insert(item))`.

phimuemue:
It would be a bit of work, but I imagine one could e.g. define a trait that encapsulates inserting an element into the container and - in the same method -  returning if the element was already present. This trait could then be implemented for `HashMap`, `BTreeMap`, `Vec` and so on.

camsteffen:
What's wrong with just using HashMap as is?

phimuemue:
Nothing per se - it just struck me that if a user wants to specify a custom hasher, they might as well want something else. But without trying I confess that generalizing this could well result in too much work to be worth it.

PR #1007, opened 2024-12-23
Implement `with_hasher` adaptors

### LINKED ISSUE #998: Provide `with_hasher` alternative adaptors
All adaptors bounded by `Hash` would potentially benefit from a `with_hasher` variant that allowed supplying a faster `Hasher` (e.g., `ahash` or `fxhash`):
- [duplicates](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.duplicates)
  - [ ] `duplicates_with_hasher`
- [duplicates_by](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.duplicates_by)
  - [ ] `duplicates_by_with_hasher`
- [unique](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.unique)
  - [ ] `unique_with_hasher`
- [unique_by](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.unique_by)
  - [ ] `unique_by_with_hasher`
- [all_unique](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.all_unique)
  - [ ] `all_unique_with_hasher`
- [into_group_map](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.into_group_map)
  - [ ] `into_group_map_with_hasher`
- [into_group_map_by](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.into_group_map_by)
  - [ ] `into_group_map_by_with_hasher`
- [into_grouping_map](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.into_grouping_map)
  - [ ] `into_grouping_map_with_hasher`
- [into_grouping_map_by](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.into_grouping_map_by)
  - [ ] `into_grouping_map_by_with_hasher`
- [counts](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.counts)
  - [ ] `counts_with_hasher`
- [counts_by](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.counts_by)
  - [ ] `counts_by_with_hasher`

Implementing each of these is a three step process:
1. Add trailing`S = RandomState` generic parameter to the adaptor type, `where S: BuildHasher`
2. Replace relevant calls to `HashMap::new` and `HashMap::default` with [HashMap::with_hasher](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.with_hasher)
3. Modify the adaptor type's internal constructor to consume an extra `hash_builder: S` parameter.
4. Add a `with_hasher` adaptor to `Itertools` with a `hash_builder: S` parameter that's passed to the internal constructor.
    - The documentation for this should defer to the non-`with_hasher` variant.
    - The documentation should include the same warning as [HashMap::with_hasher](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.with_hasher)
6. Redefine the non-`with_hasher` `Itertools` method to defer to the `with_hasher` variant.

424ever:
Closes #998

Implementations for the following methods:

- duplicates_with_hasher
- duplicates_by_with_hasher
- unique_with_hasher
- unique_by_with_hasher
- all_unique_with_hasher
- into_group_map_with_hasher
- into_group_map_by_with_hasher
- into_grouping_map_with_hasher
- into_grouping_map_by_with_hasher
- counts_with_hasher
- counts_by_with_hasher

### DIFF
--- a/src/duplicates_impl.rs
+++ b/src/duplicates_impl.rs
@@ -1,31 +1,39 @@
+use core::hash::BuildHasher;
+use std::collections::hash_map::RandomState;
 use std::hash::Hash;
 
 mod private {
+    use core::hash::BuildHasher;
+    use std::collections::hash_map::RandomState;
     use std::collections::HashMap;
     use std::fmt;
     use std::hash::Hash;
 
     #[derive(Clone)]
     #[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
-    pub struct DuplicatesBy<I: Iterator, Key, F> {
+    pub struct DuplicatesBy<I: Iterator, Key, F, S = RandomState>
+    where
+        S: BuildHasher,
+    {
         pub(crate) iter: I,
-        pub(crate) meta: Meta<Key, F>,
+        pub(crate) meta: Meta<Key, F, S>,
     }
 
-    impl<I, V, F> fmt::Debug for DuplicatesBy<I, V, F>
+    impl<I, V, F, S> fmt::Debug for DuplicatesBy<I, V, F, S>
     where
         I: Iterator + fmt::Debug,
         V: fmt::Debug + Hash + Eq,
+        S: BuildHasher,
     {
         debug_fmt_fields!(DuplicatesBy, iter, meta.used);
     }
 
-    impl<I: Iterator, Key: Eq + Hash, F> DuplicatesBy<I, Key, F> {
-        pub(crate) fn new(iter: I, key_method: F) -> Self {
+    impl<I: Iterator, Key: Eq + Hash, F, S: BuildHasher> DuplicatesBy<I, Key, F, S> {
+        pub(crate) fn new(iter: I, key_method: F, hash_builder: S) -> Self {
             Self {
                 iter,
                 meta: Meta {
-                    used: HashMap::new(),
+                    used: HashMap::with_hasher(hash_builder),
                     pending: 0,
                     key_method,
                 },
@@ -34,15 +42,16 @@ mod private {
     }
 
     #[derive(Clone)]
-    pub struct Meta<Key, F> {
-        used: HashMap<Key, bool>,
+    pub struct Meta<Key, F, S> {
+        used: HashMap<Key, bool, S>,
         pending: usize,
         key_method: F,
     }
 
-    impl<Key, F> Meta<Key, F>
+    impl<Key, F, S> Meta<Key, F, S>
     where
         Key: Eq + Hash,
+        S: BuildHasher,
     {
         /// Takes an item and returns it back to the caller if it's the second time we see it.
         /// Otherwise the item is consumed and None is returned
@@ -68,11 +77,12 @@ mod private {
         }
     }
 
-    impl<I, Key, F> Iterator for DuplicatesBy<I, Key, F>
+    impl<I, Key, F, S> Iterator for DuplicatesBy<I, Key, F, S>
     where
         I: Iterator,
         Key: Eq + Hash,
         F: KeyMethod<Key, I::Item>,
+        S: BuildHasher,
     {
         type Item = I::Item;
 
@@ -102,11 +112,12 @@ mod private {
         }
     }
 
-    impl<I, Key, F> DoubleEndedIterator for DuplicatesBy<I, Key, F>
+    impl<I, Key, F, S> DoubleEndedIterator for DuplicatesBy<I, Key, F, S>
     where
         I: DoubleEndedIterator,
         Key: Eq + Hash,
         F: KeyMethod<Key, I::Item>,
+        S: BuildHasher,
     {
         fn next_back(&mut self) -> Option<Self::Item> {
             let Self { iter, meta } = self;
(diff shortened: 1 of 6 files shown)


src/duplicates_impl.rs
```diff
@@ -1,31 +1,37 @@
...
 mod private {
+    use core::hash::BuildHasher;
     use std::collections::HashMap;
     use std::fmt;
     use std::hash::Hash;
 
     #[derive(Clone)]
     #[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
-    pub struct DuplicatesBy<I: Iterator, Key, F> {
+    pub struct DuplicatesBy<I: Iterator, Key, F, S>
```
jswrenn, line 12:
```suggestion
    pub struct DuplicatesBy<I: Iterator, Key, F, S = RandomState >
```

jswrenn (changes requested):
One small nit, but otherwise this LGTM.

src/lib.rs
```diff
@@ -1460,7 +1518,33 @@ pub trait Itertools: Iterator {
...
+    /// Return an iterator which yields the same elements as the one returned by
+    /// [.unique()](crate::Itertools::unique), but uses the specified hash builder to hash the
+    /// elements for comparison.
+    ///
+    /// Warning: `hash_builder` is normally randomly generated, and is designed to allow it's
+    /// users to be resistant to attacks that cause many collisions and very poor performance.
+    /// Setting it manually using this function can expose a DoS attack vector.
+    ///
+    /// ```
+    /// use std::hash::RandomState;
```
424ever, line 1533: What do you think about using a 3rd party hasher here in the docs? This would actually help ensure the types work with other hashers.

For example, if the (defaulted) generic argument is forgotten in one of the `impl` blocks:
``` rust
impl<I, Key, F> Iterator for DuplicatesBy<I, Key, F> { }
// which actually means
impl<I, Key, F> Iterator for DuplicatesBy<I, Key, F, std::hash::RandomState> { }
```

the types only are iterators with the default hasher.

But I didn't think it was worth pulling in a dependency just for this.

phimuemue: Is it feasible to test all of them with a simple Hasher that always hashes to 0? Could be written by ourself, and is maybe feasible to run in each test run.

(I’m on mobile and my phone is too slow to display the diffs, so please excuse if this is already addressed.)

424ever: I've added a hasher which just delegates to the default hasher but has a different type

phimuemue:
Two nits:

* I'm a bit unhappy about duplicated tests. We can leave them as is and add a TODO that we'd ideally want to write all test cases once and apply them to both normal and `_with_hasher` functions automagically.
* (Possibly over-nitpicky:) Should we rely on the fact that `RandomState` is the default `BuildHasher` or should we extract it so that `std` can do whatever it wants and we just adapt it? Could be done as follows:
  
  ```
  trait THashMap {
      type BuildHasher;
  }
  impl<K, V, S> THashMap for std::collections::HashMap<K, V, S> {
      type BuildHasher = S;
  }
  type StdDefaultBuildHasher = <std::collections::HashMap<(), ()> as THashMap>::BuildHasher;
  
  ```

@jswrenn Feel free to proceed as you deem appropriate.

jswrenn:
> * Should we rely on the fact that `RandomState` is the default `BuildHasher` or should we extract it so that `std` can do whatever it wants and we just adapt it?

It would be a breaking change for Rust to change its default hasher, so we're fine to rely on `RandomState` directly.

tests/test_std.rs
```diff
@@ -20,8 +20,31 @@ use rand::{
...
+    }
+}
+
+impl BuildHasher for TestHasher {
+    type Hasher = <RandomState as BuildHasher>::Hasher;
+
+    fn build_hasher(&self) -> Self::Hasher {
+        self.0.build_hasher()
+    }
+}
```
jswrenn, line 46: Does this buy us much? If it behaves identically to `RandomState`, then what's the advantage of newtyping it? I'm not worried that we've defined our `with_hasher` methods to *literally* only accept `RandomState`.

424ever: Because RandomState is the default for the type parameters, it's possible to implement the types only for S = RandomState. This is just the easiest way I thought of to catch that.

jswrenn: Gotcha. We can test that more directly by doing:
```rust
let _: HashMap<_, _, TestHasher> = empty::<u8>().into_group_map_with_hasher(TestHasher::new());
```
No need to then repeat the assertions on the output value.

xtqqczze:
@424ever Are you planning to continue with this PR?

424ever:
> @424ever Are you planning to continue with this PR?

Sorry for the late reply. Yes, I'll rebase this and clean it up a bit

424ever:
Do I need to do something about the failing semver checks? Those aren't in any files this PR touches.

xtqqczze:
@phimuemue @jswrenn Can we get this in for 0.15?

xtqqczze:
> Do I need to do something about the failing semver checks?

The semver check failure is informational, and expected because there are changes to the public API.

jswrenn:
> @phimuemue @jswrenn Can we get this in for 0.15?

Yeah, sure! :)

424ever:
Hi!

Unfortunately my graphics card died today so I probably can't respond to requests changes for at least a few days.

xtqqczze:
@phimuemue @jswrenn Do you have any outstanding review comments?

phimuemue (approved):
@jswrenn **tldr;** I suggest we take this as is. Could you have a look and merge it?

@424ever Thanks for this. Looks clean and relatively straightforward. Please don't push additional changes on top of eba26787e7c55bec8718df891e05ee12ef6db6d2 - I'd like to avoid re-reviewing.

---------

Once this is merged, we may come back and fine-tune:

* Documentation could be more uniform.
* "Delegating" functions could be more uniform.
* `GroupingMap` could go without `S`, and have a `_with_hasher` variant for each function.
* Tests check the interface, but not that the passed hasher is actually used inside.
* Something that automagically detects presence of `_with_hasher`-variant and compares it against its ordinary counterpart would be nice, but is hard to come up with.

jswrenn (approved):

xtqqczze:
Thanks, looking forward to the `0.15` release!

xtqqczze:
Closes https://github.com/rust-itertools/itertools/issues/462?

xtqqczze:
Closes https://github.com/rust-itertools/itertools/issues/783?

xtqqczze:
Closes https://github.com/rust-itertools/itertools/issues/588?

PR #1008, opened 2024-12-23
fix clippy lints

jswrenn:


phimuemue (approved):
I saw these lints and I approve the PR because you’re reasonable.

Still, could you share a link how these elisions work, in particular if multiple lifetimes are present?

PR #1009, opened 2024-12-24
Add "[must-use]" attribute into collect vec

Pzixel:
This might lead to the problems as following:

```
    pub fn unique_id(&self) -> impl Hash + PartialEq + Eq {
        self.items.iter().map(|x| x.id_string.clone()).collect::<Vec<_>>();
    }
```

This code has a typo and rust warns against it:

> unused return value of `collect` that must be used
> if you really need to exhaust the iterator, consider `.for_each(drop)` instead
> `#[warn(unused_must_use)]` on by defaultrustc[Click for full compiler diagnostic](rust-analyzer-diagnostics-view:/diagnostic%20message%20[0]?0#file:///Users/pzixel/Documents/Repos/backend-garden/src/models/mod.rs)
> mod.rs(964, 9): use `let _ = ...` to ignore the resulting value: `let _ = `

If you replace it with following:

```
    pub fn unique_id(&self) -> impl Hash + PartialEq + Eq {
        self.items.iter().map(|x| x.id_string.clone()).collect_vec();
    }
```

No warning will be generated, as no `must_use` attribute exist on the function.

This PR adds it, while small, it can lead to huge problems, since `()` implements a lot of traits and `impl Trait` are used very often with iterators, so you can erroneously use one instead of another.

P.S. I also noticed that `try_collect` only works with `#[cfg(feature = "use_alloc")]`, but `try_collect` doesn't. I'm not sure if this is intended.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2299,6 +2299,7 @@ pub trait Itertools: Iterator {
 
     /// `.collect_vec()` is simply a type specialization of [`Iterator::collect`],
     /// for convenience.
+    #[must_use = "if you really need to exhaust the iterator, consider `.for_each(drop)` instead"]
     #[cfg(feature = "use_alloc")]
     fn collect_vec(self) -> Vec<Self::Item>
     where
@@ -2330,6 +2331,7 @@ pub trait Itertools: Iterator {
     ///
     /// # let _ = do_stuff;
     /// ```
+    #[must_use = "if you really need to exhaust the iterator, consider `.for_each(drop)` instead"]
     fn try_collect<T, U, E>(self) -> Result<U, E>
     where
         Self: Sized + Iterator<Item = Result<T, E>>,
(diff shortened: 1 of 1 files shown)


Pzixel:
Ping?

phimuemue:
Hi there, thanks for this.

Rust itself does it for [`collect`](https://doc.rust-lang.org/src/core/iter/traits/iterator.rs.html#1967-1969), so I'm fine with this. But it does not do it for [`try_collect`](https://doc.rust-lang.org/src/core/iter/traits/iterator.rs.html#2045-2049), and I'd like to know why.

If you want, cut this PR down to only `collect`, so that we can merge to be in line with Rust. Once we find an answer to the `try_collect` thing, we can revisit.

Pzixel:
@phimuemue Fixed. I suppose the problem is that with `try_collect` you don't know what is the output type, so for some 3rd party type it might have sense to collect it without using, maybe. So I removed the `try_collect_vec` attribute for now.

phimuemue:
Thanks.

phimuemue (approved):

Issue #1001, opened 2024-12-08
Add `tuple_permutations`


JonathanBrouwer:
There is the `combinations()` and `tuple_combinations()` functions, I'm however missing the `tuple_permutations()` function.

Is there a reason this couldn't exist?
If not, I might even be open to implementing it depending on how much macro hackery is needed

ronnodas:
It may be easier to add `array_permutations()` if you want to avoid macro hackery.

JonathanBrouwer:
That's true and in general probably nicer to use than tuples, thanks for the pointer!

ronnodas:
It seems possible to share most of the code from `permutations()`, similar to the `array_combinations()` implementation. In that case we probably want to start with:
```rust
impl<T> PoolIndex<T> for Box<[usize]> {
    type Item = Vec<T>;

   ...
}
```
and then make `PermutationState` generic. (Aside: I'm not sure why `PermutationState::Loaded.indices` is a `Box<[usize]>` instead of a `Vec<usize>`.)

phimuemue:
> Aside: I'm not sure why `PermutationState::Loaded.indices` is a `Box<[usize]>` instead of a `Vec<usize>`.

Canonically we use `Box<[usize]>` if the size stays unchanged.

PR #1013, opened 2025-01-08
add `array_permutations()`

ronnodas:
Completes a variant of #1001.

### DIFF
--- a/src/permutations.rs
+++ b/src/permutations.rs
@@ -10,7 +10,7 @@ use crate::size_hint::{self, SizeHint};
 
 /// Iterator for `Vec` valued permutations returned by
 /// [`.permutations()`](crate::Itertools::permutations)
-pub type ArrayPermutations<I, const K: usize> = PermutationsGeneric<I, [usize; K]>;
+pub type ArrayPermutations<I, const K: usize> = PermutationsGeneric<I, [<I as Iterator>::Item; K]>;
 /// Iterator for const generic permutations returned by
 /// [`.array_permutations()`](crate::Itertools::array_permutations)
 pub type Permutations<I> = PermutationsGeneric<I, Vec<<I as Iterator>::Item>>;
--- a/src/permutations.rs
+++ b/src/permutations.rs
@@ -1,5 +1,6 @@
 use alloc::boxed::Box;
 use alloc::vec::Vec;
+use core::marker::PhantomData;
 use std::fmt;
 use std::iter::once;
 use std::iter::FusedIterator;
@@ -7,23 +8,32 @@ use std::iter::FusedIterator;
 use super::lazy_buffer::LazyBuffer;
 use crate::size_hint::{self, SizeHint};
 
+/// Iterator for `Vec` valued permutations returned by
+/// [`.permutations()`](crate::Itertools::permutations)
+pub type ArrayPermutations<I, const K: usize> = PermutationsGeneric<I, [usize; K]>;
+/// Iterator for const generic permutations returned by
+/// [`.array_permutations()`](crate::Itertools::array_permutations)
+pub type Permutations<I> = PermutationsGeneric<I, Vec<<I as Iterator>::Item>>;
+
 /// An iterator adaptor that iterates through all the `k`-permutations of the
 /// elements from an iterator.
 ///
-/// See [`.permutations()`](crate::Itertools::permutations) for
+/// See [`.permutations()`](crate::Itertools::permutations) and
+/// [`.array_permutations()`](crate::Itertools::array_permutations) for
 /// more information.
 #[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
-pub struct Permutations<I: Iterator> {
+pub struct PermutationsGeneric<I: Iterator, Item> {
     vals: LazyBuffer<I>,
     state: PermutationState,
+    _item: PhantomData<Item>,
 }
 
-impl<I> Clone for Permutations<I>
+impl<I, Item> Clone for PermutationsGeneric<I, Item>
 where
     I: Clone + Iterator,
     I::Item: Clone,
 {
-    clone_fields!(vals, state);
+    clone_fields!(vals, state, _item);
 }
 
 #[derive(Clone, Debug)]
@@ -41,7 +51,7 @@ enum PermutationState {
     End,
 }
 
-impl<I> fmt::Debug for Permutations<I>
+impl<I, Item> fmt::Debug for PermutationsGeneric<I, Item>
 where
     I: Iterator + fmt::Debug,
     I::Item: fmt::Debug,
@@ -49,26 +59,36 @@ where
     debug_fmt_fields!(Permutations, vals, state);
 }
 
-pub fn permutations<I: Iterator>(iter: I, k: usize) -> Permutations<I> {
-    Permutations {
+pub fn array_permutations<I: Iterator, const K: usize>(iter: I) -> ArrayPermutations<I, K> {
+    PermutationsGeneric {
         vals: LazyBuffer::new(iter),
-        state: PermutationState::Start { k },
+        state: PermutationState::Start { k: K },
+        _item: PhantomData,
     }
 }
 
-impl<I> Iterator for Permutations<I>
+pub fn permutations<I: Iterator>(iter: I, k: usize) -> Permutations<I> {
+    PermutationsGeneric {
+        vals: LazyBuffer::new(iter),
+        state: PermutationState::Start { k },
+        _item: PhantomData,
+    }
+}
+
+impl<I, Item> Iterator for PermutationsGeneric<I, Item>
 where
     I: Iterator,
     I::Item: Clone,
+    Item: PermItem<I::Item>,
 {
-    type Item = Vec<I::Item>;
+    type Item = Item;
 
     fn next(&mut self) -> Option<Self::Item> {
-        let Self { vals, state } = self;
+        let Self { vals, state, _item } = self;
         match state {
             PermutationState::Start { k: 0 } => {
                 *state = PermutationState::End;
-                Some(Vec::new())
+                Some(Item::extract_start(vals, 0, 0))
             }
             &mut PermutationState::Start { k } => {
                 vals.prefill(k);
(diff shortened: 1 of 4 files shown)


src/permutations.rs
```diff
@@ -77,14 +97,11 @@ where
...
-                Some(vals[0..k].to_vec())
+                Some(Item::extract_start(vals, k, k - 1))
             }
             PermutationState::Buffered { ref k, min_n } => {
                 if vals.get_next() {
-                    let item = (0..*k - 1)
-                        .chain(once(*min_n))
-                        .map(|i| vals[i].clone())
-                        .collect();
+                    let item = Item::extract_start(vals, *k, *min_n);
```
phimuemue, line 104: Having a separate variant `Buffered` prevents us from using `PoolIndex`. Maybe always constructing `indices` might allow `PoolIndex` here, simplifying implementation and improving maintenance.

src/permutations.rs
```diff
@@ -41,34 +51,44 @@ enum PermutationState {
...
+    type Item = Item;
 
     fn next(&mut self) -> Option<Self::Item> {
-        let Self { vals, state } = self;
+        let Self { vals, state, _item } = self;
         match state {
             PermutationState::Start { k: 0 } => {
                 *state = PermutationState::End;
-                Some(Vec::new())
+                Some(Item::extract_start(vals, 0, 0))
```
phimuemue, line 91: I see calling `extract_start` with 0 here has the merit to save one function, but it requires us to account for the special case `len==0` in `extract_start`, right? Would unconditionally working on `indices` simplify and get rid of special cases?

src/permutations.rs
```diff
@@ -184,3 +202,39 @@ impl PermutationState {
...
+}
+
+impl<T: Clone, const K: usize> PermItem<T> for [T; K] {
+    fn extract_start<I: Iterator<Item = T>>(buf: &LazyBuffer<I>, len: usize, last: usize) -> Self {
+        assert_eq!(len, K);
+        buf.get_array_from_fn(|i| if i + 1 < len { i } else { last })
+    }
+
+    fn extract_from_prefix<I: Iterator<Item = T>>(buf: &LazyBuffer<I>, indices: &[usize]) -> Self {
+        buf.get_array_from_fn(|i| indices[i])
```
phimuemue, line 221: Isn't this essentially `buf.get_array(indices)`? (Would support my theory that `PermItem` should actually be part of `PoolIndex`.)

src/permutations.rs
```diff
@@ -184,3 +202,39 @@ impl PermutationState {
         }
     }
 }
+
+/// A type that can be picked out from a pool or buffer of items from an inner iterator
+/// and in a generic way given their indices.
+pub trait PermItem<T> {
```
phimuemue, line 208: Is this new trait really necessary? Couldn't we re-use/generalize/extend `PoolIndex`?

src/permutations.rs
```diff
@@ -184,3 +202,39 @@ impl PermutationState {
...
+impl<T: Clone> PermItem<T> for Vec<T> {
+    fn extract_start<I: Iterator<Item = T>>(buf: &LazyBuffer<I>, len: usize, last: usize) -> Self {
+        if len == 0 {
+            Vec::new()
+        } else {
+            (0..len - 1)
+                .chain(once(last))
+                .map(|i| buf[i].clone())
+                .collect()
+        }
```
phimuemue, line 234: Special cases that might go away if we always worked on `indices`.

phimuemue (changes requested):
Hi there, thanks for you effort.

I fear `Permutations` has some quirks that I do not want to entrench in our code. I think the main problem boils down to the fact that it uses many different states instead of always working on `indices`. (I see that it may be nice to sometimes avoid constructing it, but it appears to complicate the surrounding code.)

We should first try to clean up `Permutations` to the usual pattern where we "operate on `indices`, `fn next` maps `indices` onto elements", so that we can - afterwards - go with the existing `PoolIndex` (instead of a new `PermIter`).

Unless this clean-up is infeasible, I want to avoid new abstractions. Thus, I'd like to postpone this PR until we have evidence that `Permutations` is as simple as possible.

ronnodas:
My original intention *was* to reuse `PoolIndex` (as I wrote in https://github.com/rust-itertools/itertools/issues/1001#issuecomment-2573884034) but that seems to actually be *more* complicated without tweaking the algorithm used to generate the indices. The crucial difference is that `PermutationState::Loaded.indices` does not have length `K`/`k` but `n`, the length of the original iterator. So you can't just make this field into an `IndexPool`. The actual array of indices that are looked up in the buffer is the length `k` prefix of this slice, as visible in the call `extract_from_prefix(vals, &indices[0..k])`. 

I can think about modifying the algorithm to only keep the `k` relevant indices but if someone else wants to make that modification to `Permutations` I'm happy to redo the PR based on that change.

src/permutations.rs
ronnodas: Note that when this function is called, there does not exist an instance of `[usize; K]` to call `buf.get_array()` on. We could do something like `buf.get_array(indices.try_into().unwrap())` instead if that seems better.

src/permutations.rs
ronnodas: That is what I started with but it seems that the operations required for `ArrayPermutations` are different from the ones for `{Array}Combinations`. Even in the existing `Permutations` implementation you can see that there are as many instances of directly indexing the buffer as calling `buf.get_at()`. Since the intermediate slice/array of indices would be created only to be immediately used to index the buffer, it seems like an unnecessary level of indirection to extend the PoolIndex trait.

src/permutations.rs
```diff
@@ -1,29 +1,39 @@
...
 
 use super::lazy_buffer::LazyBuffer;
 use crate::size_hint::{self, SizeHint};
 
+/// Iterator for `Vec` valued permutations returned by
+/// [`.permutations()`](crate::Itertools::permutations)
+pub type ArrayPermutations<I, const K: usize> = PermutationsGeneric<I, [usize; K]>;
+/// Iterator for const generic permutations returned by
+/// [`.array_permutations()`](crate::Itertools::array_permutations)
+pub type Permutations<I> = PermutationsGeneric<I, Vec<<I as Iterator>::Item>>;
```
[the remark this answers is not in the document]
ronnodas: This was an error, `array_permutations()` was only working on `Iterator<Item = usize>` before. Fixed now.

src/permutations.rs
```diff
@@ -1,29 +1,39 @@
...
-/// See [`.permutations()`](crate::Itertools::permutations) for
+/// See [`.permutations()`](crate::Itertools::permutations) and
+/// [`.array_permutations()`](crate::Itertools::array_permutations) for
 /// more information.
 #[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
-pub struct Permutations<I: Iterator> {
+pub struct PermutationsGeneric<I: Iterator, Item> {
     vals: LazyBuffer<I>,
     state: PermutationState,
+    _item: PhantomData<Item>,
```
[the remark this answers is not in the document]
phimuemue: Yes, using a `PoolIndex` for cycles itself is wrong, but we already exploit the fact that the length is `k` (https://docs.rs/itertools/latest/src/itertools/permutations.rs.html#112). So it should possibly be `PoolIdx<usize>::Item`.

src/permutations.rs
phimuemue: Correct, this attempt won't work. However, we could introduce `PoolIndex::from_fn` to do whatever we want and return either a `Vec` or an array.

phimuemue:
Thanks for your reply. I experimented a bit (https://github.com/phimuemue/rust-itertools/tree/array_permutations_via_poolindex) and I think we should extend (and probably rename) `PoolIndex`.

ronnodas:
Does it make sense to flip `PoolIndex` and instead of:
```rust
trait PoolIndex<T> {
    type Item;
    
    fn extract_item(&self, pool) -> Self::Item
}
```
have something like:
```rust
trait PoolOutput<T, Idx> {
    fn extract(pool, idx: &Self::Idx) -> Self;
}
```

Then you can have
```rust
impl PoolOutput<T, [usize]> for Vec<T> { ... }
impl PoolOutput<T, [usize; K]> for [T; K] { ... }
```
and even
```
struct IndexFn<F>;

impl<F: FnMut(usize) -> usize> PoolOutput<T, IndexFn<F>> for [T; K] {...}

impl<F: FnMut(usize) -> usize> PoolOutput<T, (IndexFn<F>, usize)> for Vec<T> {...}
```
instead of a `from_fn` method.

phimuemue:
> Thanks!
> 
> Does it make sense to flip `PoolIndex` and instead of:
> 
> ```rust
> trait PoolIndex<T> {
>     type Item;
>     
>     fn extract_item(&self, pool) -> Self::Item
> }
> ```
> 
> have something like:
> 
> ```rust
> trait PoolOutput<T, Idx> {
>     fn extract(pool, idx: &Self::Idx) -> Self;
> }
> ```
> 
> Then you can have
> 
> ```rust
> impl PoolOutput<T, [usize]> for Vec<T> { ... }
> impl PoolOutput<T, [usize; K]> for [T; K] { ... }
> ```
> 
> and even
> 
> ```rust
> struct IndexFn<F>;
> 
> impl<F: FnMut(usize) -> usize> PoolOutput<T, IndexFn<F>> for [T; K] {...}
> 
> impl<F: FnMut(usize) -> usize> PoolOutput<T, (IndexFn<F>, usize)> for Vec<T> {...}
> ```
> 
> instead of a `from_fn` method.
> 
> A big disadvantage would be that `CombinationsGeneric` would now need both the `Idx` and `Item` as parameters, since it has a field of type `Idx` and needs to have `Item` as an associated type to implement `Iterator`.

I can't reliably predict how simple the outcome would be, but my gut feeling is that `PoolIndex` is simpler than `PoolIndex<T, ...>` and `PoolIndex::from_fn` is simpler than `impl PoolOutput<T, IndexFn>`.

ronnodas:
I've tried to make some simplifications at https://github.com/ronnodas/itertools/tree/array-permutations-attempt-2. I do agree that `PoolIndex` is not the best name any more (especially with it being the type of `PermutationState::Loaded.cycles`), but don't have a good alternative.

phimuemue:
> I do agree that `PoolIndex` is not the best name any more (especially with it being the type of `PermutationState::Loaded.cycles`), but don't have a good alternative.

Maybe `ArrayOrVecHelper`. I'm unsure if you'd really want `impl ArrayOrVectorHelper for [usize; K]` or rather `struct ArrayHelper; impl ArrayOrVectorHelper for ArrayHelper`, i.e. introduce dedicated structs for arrays and `Vec`s, respectively.

**Note if you'd like to convert this into a PR** (please excuse this note you did not ask for, but I want to clarify things before huge PRs): https://github.com/rust-itertools/itertools/commit/a916d375dcebcfe5fc2c9cdbe5e82150d35ef2b8 is hard to review, because it lumps different things into one commit (moving `PoolIndex` to another file, introducing `fn start` and `fn item_from_fn`, changing `Buffered::k` to `Buffered::indices`, ...). https://github.com/rust-itertools/itertools/pull/790 is a good example how such large PRs could be structured.

ronnodas:
@phimuemue Thanks! Broke up into smaller commits and filed as https://github.com/rust-itertools/itertools/pull/1014.

PR #1015, opened 2025-01-21
Introduce Itertools::owned() / Owned.

hadronized:
This is very similar to std’s `Iterator::clone()` / `Cloned`, but works on more types, such as `CStr` / `CString`.

### DIFF
--- a/benches/specializations.rs
+++ b/benches/specializations.rs
@@ -1,5 +1,7 @@
 #![allow(unstable_name_collisions)]
 
+use std::ffi::CString;
+
 use criterion::black_box;
 use criterion::BenchmarkId;
 use itertools::Itertools;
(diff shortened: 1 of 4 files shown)


phimuemue:
Is this essentially `iterator.map(ToOwned.to_owned)`?

If so:
* We are reluctant to include even more "special-case `map` instances" (relevant discussion e.g. https://github.com/rust-itertools/itertools/pull/714#issuecomment-1963105704). I support these arguments and would prefer to not pull it into `Itertools`.
* If someone else wants to merge, it should probably be implemented in terms of `MapSpecialCase`.

hadronized:
Understood, I’ll wait for others to comment to decide what to do but I’m fine to close if you think it’s not a good fit.

phimuemue:
Thanks @hadronized, but I close this.

So far it's been established that we do not want to include even more of these special-case-map-methods.

hadronized:
That completely makes sense. Thanks for the feedback!

PR #1019, opened 2025-02-12
Add is_first/is_last helpers to Position

xmo-odoo:
They're just convenience method, but avoid having to dereference individual variants, and may mitigate missing the `Only` case.

### DIFF
--- a/src/with_position.rs
+++ b/src/with_position.rs
@@ -58,6 +58,21 @@ pub enum Position {
     Only,
 }
 
+impl Position {
+    /// Whether the current element is the first of the iterator,
+    /// regardless of length (includes both [`Position::First`]
+    /// and [`Position::Only`]).
+    pub fn is_first(&self) -> bool {
+        matches!(self, Self::First | Self::Only)
+    }
+    /// Whether the current element is the last of the iterator,
+    /// regardless of length (includes both [`Position::Last`]
+    /// and [`Position::Only`]).
+    pub fn is_last(&self) -> bool {
+        matches!(self, Self::Only | Self::Last)
+    }
+}
+
 impl<I: Iterator> Iterator for WithPosition<I> {
     type Item = (Position, I::Item);
 


phimuemue:
> To me that is the entire point, it's a predicate telling you whether an element is the first / last of the sequence regardless of the sequence's dimensions, so you don't need to know / care about the `Only` case.
> 
> I was reminder of this inconvenience upon stumbling (back) on https://stackoverflow.com/a/76223155/8182118

I appreciate the reasoning, but it goes against Rust's convention:
* `enum Result{Ok, Err}` has `is_ok`, `is_err`
* `enum Option{Some, None}` has `is_some`, `is_none`
* `enum ControlFlow{Continue, Break}` has `is_continue`, `is_break`
* `enum Poll{Ready, Pending}` has `is_ready`, `is_pending`
* `enum Cow{Borrowed, Owned}` has (experimental) `is_borrowed`, `is_owned`

I.e. `is_`-functions refer to *one specific* variant of the enum. Thus, I'd expect `is_first` to match `First` - but since this is probably not intentional I suggest renaming it to `is_first_or_only` so that the programmer gets a compiler error when trying `is_first` and is pushed towards thinking about the `Only` case.

(Notably, `enum Ordering{Less, Equal, Greater}` does not have `is_less` et al, but has has `is_lt` *and* `is_le`.)

jswrenn:
I mean that `iter.foo().bar().map_unwrap().baz()` isn't much of a win over `iter.foo().bar().map(Option::unwrap).baz()`, because it only saves a few keystrokes, but `iter.foo().bar().baz()` is a bigger win over `bar(iter.foo()).baz()`, because it preserves a left-to-right reading order (i.e., chainability). 

I'm more amenable to shorthands when they preserve chainability than when they merely safe keystrokes.

phimuemue:
Closing because of #1042.

Issue #1012, opened 2025-01-04
Add `array_chunks` and `array_windows`?


ronnodas:
Now that `next_array` exists, it's not so hard to write the methods `array_chunks` (functionally equivalent to `from_fn(|| self.next_array())`) and `array_windows`, similar to the unstable [`slice::array_chunks`](https://github.com/rust-lang/rust/issues/74985) and [`slice::array_windows`](https://github.com/rust-lang/rust/issues/75027) respectively. However it's similarly unclear what `array_chunks<0>` and `array_windows<0>` should do. Does it make sense to emit a post-monomorphization (but still compile-time) error, via say `const { assert!(N > 0) }`? Or is it better to produce run-time errors?

Is the name `arrays` preferable to `array_chunks`, to be consistent with `Itertools::tuples`?

ronnodas:
Another question, do we also want a `remainder()` or `into_inner()` method on `ArrayChunks` for the case where the length of the iterator isn't a perfect multiple of `N`? The `into_inner()` version is trivial, but to support accessing the remainder after exhaustion (via `by_ref()`), we may require more unsafe code (perhaps in methods on `ArrayBuilder`) and this possibly also makes the `impl Clone for ArrayChunks` more complicated (since `MaybeUninit<T>: Clone` only if `T: Copy`).

jswrenn:
A PME is always preferable to a runtime error, but, if possible, I'd prefer `array_chunks<0>` to produce an inexhaustible iterator of empty chunks.

We definitely want a `remainder()` method, but we have a high bar for merging unsafe code. If the needed unsafe code turns out to be trivial, we'd be more likely to merge it.

jswrenn:
Yeah, good point. PME, please, then.

PR #1023, opened 2025-03-02
Add array_chunks

ronnodas:
Added an implementation of `arrays` that just repeatedly calls `next_array()`. Currently the `N = 0` produces a post-monomorphization error, see the discussion in #1012.

The `remainder()` feature is *not* implemented, based on the discussion below, since this seems to require either:
 * gating this behind `use_alloc`, or
 * making the return type not `Clone`, or
 * more unsafe code implementing effectively more of `arrayvec`.

### DIFF
--- a/src/arrays.rs
+++ b/src/arrays.rs
@@ -2,6 +2,23 @@ use alloc::vec::Vec;
 
 use crate::next_array::ArrayBuilder;
 
+macro_rules! const_assert_positive {
+    ($N: ty) => {
+        trait StaticAssert<const N: usize> {
+            const ASSERT: bool;
+        }
+
+        impl<const N: usize> StaticAssert<N> for () {
+            const ASSERT: bool = {
+                assert!(N > 0);
+                true
+            };
+        }
+
+        assert!(<() as StaticAssert<N>>::ASSERT);
+    };
+}
+
 /// An iterator that groups the items in arrays of const generic size `N`.
 ///
 /// See [`.next_array()`](crate::Itertools::next_array) for details.
@@ -13,7 +30,7 @@ pub struct Arrays<I: Iterator, const N: usize> {
 
 impl<I: Iterator, const N: usize> Arrays<I, N> {
     pub(crate) fn new(iter: I) -> Self {
-        assert_positive::<N>();
+        const_assert_positive!(N);
 
         // TODO should we use iter.fuse() instead? Otherwise remainder may behave strangely
         Self {
@@ -84,22 +101,6 @@ impl<I: Iterator, const N: usize> Iterator for Arrays<I, N> {
 }
 
 impl<I: ExactSizeIterator, const N: usize> ExactSizeIterator for Arrays<I, N> {}
-
-/// Effectively assert!(N > 0) post-monomorphization
-fn assert_positive<const N: usize>() {
-    trait StaticAssert<const N: usize> {
-        const ASSERT: bool;
-    }
-
-    impl<const N: usize> StaticAssert<N> for () {
-        const ASSERT: bool = {
-            assert!(N > 0);
-            true
-        };
-    }
-
-    assert!(<() as StaticAssert<N>>::ASSERT);
-}
 
 #[cfg(test)]
 mod tests {
--- a/src/arrays.rs
+++ b/src/arrays.rs
@@ -13,9 +13,8 @@ pub struct Arrays<I: Iterator, const N: usize> {
 
 impl<I: Iterator, const N: usize> Arrays<I, N> {
     pub(crate) fn new(iter: I) -> Self {
-        const {
-            assert!(N > 0);
-        }
+        assert_positive::<N>();
+
         // TODO should we use iter.fuse() instead? Otherwise remainder may behave strangely
         Self {
             iter,
@@ -85,6 +84,22 @@ impl<I: Iterator, const N: usize> Iterator for Arrays<I, N> {
 }
 
 impl<I: ExactSizeIterator, const N: usize> ExactSizeIterator for Arrays<I, N> {}
+
+/// Effectively assert!(N > 0) post-monomorphization
+fn assert_positive<const N: usize>() {
+    trait StaticAssert<const N: usize> {
+        const ASSERT: bool;
+    }
+
+    impl<const N: usize> StaticAssert<N> for () {
+        const ASSERT: bool = {
+            assert!(N > 0);
+            true
+        };
+    }
+
+    assert!(<() as StaticAssert<N>>::ASSERT);
+}
 
 #[cfg(test)]
 mod tests {
--- a/src/arrays.rs
+++ b/src/arrays.rs
@@ -6,12 +6,12 @@ use crate::next_array::ArrayBuilder;
 ///
 /// See [`.next_array()`](crate::Itertools::next_array) for details.
 #[derive(Debug, Clone)]
-pub struct ArrayChunks<I: Iterator, const N: usize> {
+pub struct Arrays<I: Iterator, const N: usize> {
     iter: I,
     partial: Vec<I::Item>,
 }
 
-impl<I: Iterator, const N: usize> ArrayChunks<I, N> {
+impl<I: Iterator, const N: usize> Arrays<I, N> {
     pub(crate) fn new(iter: I) -> Self {
         const {
             assert!(N > 0);
@@ -36,13 +36,13 @@ impl<I: Iterator, const N: usize> ArrayChunks<I, N> {
     /// ```
     /// use itertools::Itertools;
     ///
-    /// let mut it = (1..9).array_chunks();
+    /// let mut it = (1..9).arrays();
     /// assert_eq!(Some([1, 2, 3]), it.next());
     /// assert_eq!(Some([4, 5, 6]), it.next());
     /// assert_eq!(None, it.next());
     /// itertools::assert_equal(it.remainder(), [7,8]);
     ///
-    /// let mut it = (1..9).array_chunks();
+    /// let mut it = (1..9).arrays();
     /// assert_eq!(Some([1, 2, 3]), it.next());
     /// itertools::assert_equal(it.remainder(), 4..9);
     /// ```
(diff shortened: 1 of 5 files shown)


ronnodas:
Re the MSRV failure, I'm not sure what the idiomatic way is to trigger a PME on 1.63 based on `N > 0`.

jswrenn:
> Re the MSRV failure, I'm not sure what the idiomatic way is to trigger a PME on 1.63 based on `N > 0`.

You can do it like this: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=233fb2f6ca3df52d15e3df96b8c4cb61

src/lib.rs
```diff
@@ -741,6 +745,55 @@ pub trait Itertools: Iterator {
...
+    /// ```
+    ///
+    /// ```compile_fail
+    /// use itertools::Itertools;
+    ///
+    /// let mut it = (1..5).arrays::<0>();
+    /// assert_eq!(Some([]), it.next());
+    /// ```
+    #[cfg(feature = "use_alloc")]
+    fn arrays<const N: usize>(self) -> Arrays<Self, N>
```
jswrenn, line 790: IMO `array_chunks` is the better name, since it differentiates it from `array_windows`.

src/arrays.rs
```diff
@@ -0,0 +1,133 @@
...
+use crate::next_array::ArrayBuilder;
+
+/// An iterator that groups the items in arrays of const generic size `N`.
+///
+/// See [`.next_array()`](crate::Itertools::next_array) for details.
+#[derive(Debug, Clone)]
+pub struct Arrays<I: Iterator, const N: usize> {
+    iter: I,
+    partial: Vec<I::Item>,
+}
```
jswrenn, line 12: It strikes me as a code smell that `N` doesn't appear in the definition here. Shouldn't `partial` be an `ArrayBuilder<T, N>`?

jswrenn (changes requested):

src/lib.rs
ronnodas: To add context, I was trying to avoid triggering the `unstable-name-collisions` lint, because of https://github.com/rust-lang/rust/issues/100450. Should I change it back to array_chunks?

src/arrays.rs
ronnodas: Making it `ArrayBuilder<T, N>` a priori makes sense, but I had trouble with the bounds for the `impl Clone` in that case. Note that `MaybeUninit<T>: Clone` [requires `T: Copy`](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#impl-Clone-for-MaybeUninit%3CT%3E). Suggestions?

src/arrays.rs
```diff
@@ -0,0 +1,133 @@
...
+
+    impl<const N: usize> StaticAssert<N> for () {
+        const ASSERT: bool = {
+            assert!(N > 0);
+            true
+        };
+    }
+
+    assert!(<() as StaticAssert<N>>::ASSERT);
+}
```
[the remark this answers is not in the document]
ronnodas: Is the change in [db21b2f](https://github.com/rust-itertools/itertools/pull/1023/commits/db21b2f9dfbaf4b7f24c8c05e77d8bb2fc2782e7) what you meant?

bend-n:
im curious why you didnt want to just copy the previous stdlib implementation? you can simply implement this on top of a manual `as_chunks`?
returning an owned array and using a vec for remainder is kind of funky.

bend-n:
oh, my bad. i forgot that iterator::array_chunks was still unstable and assumed this was with regards to slice array chunks for some reason.

ronnodas:
> you could still copy the implementation of iterator::array_chunks, though?

If you go through the source, `Iterator::array_chunks` is implemented in terms of (also unstable) `Iterator::next_chunk`, which is equivalent to `Itertools::next_array`. It made more sense to use the thing that's already in the crate rather than copy over something with duplicate functionality.

src/arrays.rs
```diff
@@ -0,0 +1,134 @@
...
+impl<I: Iterator, const N: usize> Iterator for Arrays<I, N> {
+    type Item = [I::Item; N];
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if !self.partial.is_empty() {
+            return None;
+        }
+        let mut builder = ArrayBuilder::new();
+        for _ in 0..N {
+            if let Some(item) = self.iter.next() {
```
bend-n, line 79: can this not be a ?

ronnodas: Only at `0`, otherwise `self.partial` needs to be updated so an early return is not ok.

src/arrays.rs
phimuemue: Yes, should be ArrayBuilder, because this brings it closer to no_alloc.

Maybe this array builder could also be re-used in `next`.

Can you remember/share the exact problem?

phimuemue:
If it turns out to be too complicated, maybe go without `partial` in our first iteration?

src/arrays.rs
ronnodas: @phimuemue If it holds an `ArrayBuilder<T>` then it won't be `Clone` without `T : Copy`. But we should have `Arrays<I, N>: Clone` when `I: Clone` and `I::Item : Clone`.

phimuemue: Ok. Then, let's implement it without partial/remaining for now.

(Because if we extend our ArrayBuilder further and further, we land at (a less tested version of) ArrayVec.)

ronnodas:
@phimuemue I've removed `partial`/`remainder()` and thereby the gating behind `use_alloc`.

src/next_array.rs
```diff
@@ -86,6 +88,23 @@ impl<T, const N: usize> ArrayBuilder<T, N> {
             None
         }
     }
+
+    #[cfg(feature = "use_alloc")]
```
phimuemue: Still needed?

ronnodas: Thanks, that file does not need to be touched at all.

src/arrays.rs
```diff
@@ -0,0 +1,85 @@
...
+
+impl<I: Iterator, const N: usize> Iterator for Arrays<I, N> {
+    type Item = [I::Item; N];
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.iter.next_array()
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        if N == 0 {
```
phimuemue: Can't we static assert that N is positive?

ronnodas: I just removed the check. Do you think it's better to add an `assert!(N > 0)` or `const_assert_positive!(N)`?

Also, for an unfused iterator, I'm assuming that the lower bound in `size_hint()` is about the number of items yielded before the first `None`, not overall.

src/arrays.rs
```diff
@@ -0,0 +1,82 @@
...
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let (lo, hi) = self.iter.size_hint();
+        // `N` guaranteed to be positive in `new()`
+        (lo / N, hi.map(|hi| hi / N))
+    }
+}
+
+impl<I: ExactSizeIterator, const N: usize> ExactSizeIterator for Arrays<I, N> {}
+
+#[cfg(test)]
```
phimuemue: I think we have quick check tests for these already. (Look for exact_size or exact_size_for_this.)

tests/quick.rs
```diff
@@ -1373,6 +1373,16 @@ quickcheck! {
     }
 }
 
+quickcheck!{
+    fn arrays_exact_size_1(a: Vec<u8>) -> bool {
+        exact_size(a.iter().arrays::<1>())
+    }
+
+    fn arrays_exact_size_4(a: Vec<u8>) -> bool {
+        exact_size(a.iter().arrays::<1>())
```
phimuemue: Should be 4 instead of 1, shouldn't it?

src/arrays.rs
phimuemue: Const assert to have it during compile time.

src/lib.rs
```diff
@@ -790,6 +792,45 @@ pub trait Itertools: Iterator {
...
+    /// for [a, b] in (1..5).arrays() {
+    ///     v.push([a, b]);
+    /// }
+    /// assert_eq!(v, vec![[1, 2], [3, 4]]);
+    ///
+    /// // this requires a type hint
+    /// let it = (1..7).arrays::<3>();
+    /// itertools::assert_equal(it, vec![[1, 2, 3], [4, 5, 6]]);
+    ///
+    /// // you can also specify the complete type
```
phimuemue, line 813: Let's not encourage people to spell iterator types

src/lib.rs
```diff
@@ -790,6 +792,45 @@ pub trait Itertools: Iterator {
         groupbylazy::new_chunks(self, size)
     }
 
+    /// Return an iterator that groups the items in arrays of const generic size `N`.
+    ///
+    /// `N == 0` is a compile-time (but post-monomorphization) error.
```
phimuemue, line 797: Is it really important for users that it is post-monomorphization?

src/arrays.rs
```diff
@@ -0,0 +1,51 @@
...
+
+impl<I: Iterator, const N: usize> Iterator for Arrays<I, N> {
+    type Item = [I::Item; N];
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.iter.next_array()
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let (lo, hi) = self.iter.size_hint();
```
phimuemue, line 47: Please add the const assert

phimuemue:
@jswrenn regarding naming: what's our policy? IMHO, the best name is array_chunks, and I would use it and deprecate our function once it's also available in std.

src/lib.rs
ronnodas: Note that such an example is included in the docs for `tuple_windows()`, `tuples()`, `tuple_combinations()` and `array_combinations()`. Should I still remove this or keep it for consistency?

src/lib.rs
ronnodas: It won't be caught by `cargo check`, for example, which seems relevant to users.

src/lib.rs
phimuemue: As you wish.

Issue #781, opened 2023-10-12
Use `fold` instead of `for_each`?


Philippe-Cholet:
https://github.com/rust-itertools/itertools/pull/780#discussion_r1355061606 by **jswrenn**

> In the Rust standard library, it's quite common for `fold` to be implemented in terms of `try_fold`. In your original code you construct a `TakeWhile` and call `fold` on it. But `TakeWhile`'s fold implementation delegates to `try_fold`!
> 
> The general rule to follow is something like this:
> 
> * If you need to completely consume an iterator, **use `for_each` or `fold`.**
>       
>   * **Use `for_each` if you don't need to maintain an accumulator.**
>   * **Use `fold` if you do need to maintain an accumulator.**
> 
> * If you need to partly consume an iterator, use `try_for_each` or `try_fold`.
>       
>   * Same rules about accumulators apply.

**Bold** is mine as I wonder if we are using `for_each` (convenient) at multiple places instead of `fold`.

Current `for_each` uses:
- lib.rs: `Itertools::{counts, partition_map, join}`
- extrema_set.rs: `min_set_impl`
- grouping_map.rs: `GroupingMap::{aggregate, collect}`
- group_map.rs: `into_group_map`
- k_smallest.rs: `k_smallest`

(There is also `Itertools::foreach` but this one is okay.)

Why do I care? While benchmarking a new `MapForGrouping::fold`, I found out that `some_slice_iter::for_each` does not delegate to `fold` (default behavior) and is not optimized \* (just `while let Some(x) = self.next() { f(x); }`) while `some_slice_iter::fold` is optimized (which is useful to see benchmark improvements of specializing `fold` methods of our adaptors by delegating to the inner iterator).

\*: The `for_each` has a comment: "_We override the default implementation, which uses `try_fold`, because this simple implementation generates less LLVM IR and is faster to compile._"

Philippe-Cholet:
Well, apart from `Itertools::join` that is known to be slow anyway, I benchmarked each "`for_each`  → `fold`" change and got mostly severe regressions. The only "win" was -5% for `into_group_map`.

Before I close this, do you know why?

Philippe-Cholet:
Thanks for the detailled precision.

PR #1027, opened 2025-04-29
into_group_map: make code more idiomatic

Kek5chen:
This shouldn't change any runtime. I just noticed this while using the function, as i replaced exactly this (new) code, with the itertools version and wanted to contribute it back.

### DIFF
--- a/src/group_map.rs
+++ b/src/group_map.rs
@@ -13,10 +13,10 @@ where
     I: Iterator<Item = (K, V)>,
     K: Hash + Eq,
 {
-    let mut lookup = HashMap::new();
+    let mut lookup = HashMap::<K, Vec<V>>::new();
 
     iter.for_each(|(key, val)| {
-        lookup.entry(key).or_insert_with(Vec::new).push(val);
+        lookup.entry(key).or_default().push(val);
     });
 
     lookup


phimuemue:
Hi there, thanks for this.

I'd like to take the `or_default` part, as it's more concise.

But I want to double check `for_each` -> `fold` part. As discussed in https://github.com/rust-itertools/itertools/issues/781#issuecomment-1760222069, `for_each` may be faster than `fold`. But this very discussion mentions that `into_group_map` may be better off using `fold`. Still, if we migrate to `fold`, I want to be sure we don't deteriorate performance.

Kek5chen:
Is there anything you expect of me for this to be merged? In case you do not want to use fold after deciding on the mentioned issue, i can quickly roll the change to fold back so this can be merged.

Kek5chen:
What i'm asking is, would you like me to benchmark this for some different cases?

jswrenn:
I'd say keep the `for_each` as-is, but I like the `or_default` change!

It's quite hard to mircobenchmark these sorts of things, since they often depend on how much gas the optimizer has in particular contexts. I don't think it's worth the effort. I'd rather just use the thing that *probably* optimizes well and isn't really any harder to read.

Kek5chen:
Yes. So, I dove a bit deeper into for_each vs fold. I wrote an example in [compiler explorer](https://godbolt.org/z/7osGnE44Y)

It seems like the for_each keeps more of the iteration state hot in registers while the fold spills much more into the stack and reloads half a dozen registers every trip through the loop, so it executes ~35% more instructions (rough estimate, i didn't pull out a calculator, and this also doesn't factor in cpu internal instruction cycle times) and performs a lot more memory traffic.

That's just about how much I could get from this. I'm not an assembly god. Though it seems like just running with the for_each but using default is the go-to until maybe compiler optimizations some day say otherwise.

Of course this also doesn't include different use cases with different concrete types. This is just a naive, simple example.

The inner part of the loop is largely the same.

phimuemue (approved):

PR #1031, opened 2025-05-24
track_caller for zip_eq

dspyz-matician:
This is expected to panic on mismatch


phimuemue:
Hi there, thanks for this.

@jswrenn I'm torn on this one: That `track_caller` is nice, but it does afaik not reliably catch all notable places: Here, we would "fix" `next`, but all functions calling `next` would still be unsolved. Thus, I'm unsure if we should start adopting this feature in `itertools`, especially if there are many places where its use is debatable.

PR #1032, opened 2025-06-05
Canonicalize `all_equal_value`'s error

phimuemue:
Similar to `exactly_one_error`'s error type, `all_equal_value`'s error type now implements `std::error::Error`. (I saw the discrepancy because [`exactly_one_error` works with `anyhow`, whereas `all_equal_value` does not](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=1a7b1ab8181cc5345ecc2461709b8cf6).)

Since I had to introduce a new type anyways, I converted from `(Item, Item)` to `[Item; 2]`. I suggest we generally lean towards arrays instead of tuples if the components have the same type.

### DIFF
--- a/src/all_equal_value_err.rs
+++ b/src/all_equal_value_err.rs
@@ -1,0 +1,3 @@
+/// Value returned for the error case of `Itertools::all_equal_value()`.
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub struct AllEqualValueError<Item>(pub Option<[Item; 2]>);
--- a/src/all_equal_value_err.rs
+++ b/src/all_equal_value_err.rs
@@ -1,3 +1,31 @@
-/// Value returned for the error case of `Itertools::all_equal_value()`.
+#[cfg(doc)]
+use crate::Itertools;
+#[cfg(feature = "use_std")]
+use std::error::Error;
+use std::fmt::{Debug, Display, Formatter, Result as FmtResult};
+
+/// Value returned for the error case of [`Itertools::all_equal_value`].
 #[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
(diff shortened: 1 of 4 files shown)


phimuemue:
Checks fail because `Display` implementation is not covered, similar to the existing `ExactlyOneError`.

src/all_equal_value_err.rs
```diff
@@ -0,0 +1,29 @@
+#[cfg(feature = "use_std")]
+use std::error::Error;
+use std::fmt::{Debug, Display, Formatter, Result as FmtResult};
+
+/// Value returned for the error case of `Itertools::all_equal_value()`.
```
jswrenn, line 5: Can you make this an intradoc link:
```suggestion
/// Value returned for the error case of [`Itertools::all_equal_value]`.
```
you might need to add a `#[cfg(doc)] use crate::Itertools;` to the top of the file.

jswrenn (changes requested):

src/all_equal_value_err.rs
phimuemue: Done.

phimuemue:
Rebased against our main branch.

jswrenn:
@phimuemue Can you rebase + add an entry to the changelog? Since we're cutting a breaking release anyways, I'd like to include this.

jswrenn (approved):

phimuemue:
> @phimuemue Can you rebase + add an entry to the changelog? Since we're cutting a breaking release anyways, I'd like to include this.

I tried via the web UI, so I'm not sure if it worked. If not, I'll try to get back to this this evening.

jswrenn:
Thanks! Feel free to hit the merge button once the CHANGELOG has an entry for this. :-)

Are you at RustWeek by any chance?

PR #1033, opened 2025-06-13
Add `array_combinations_with_replacement`

### LINKED ISSUE #1025: Feature request: `array_combinations_with_replacement`
For [combinations](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.combinations) there is [array_combinations](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.array_combinations).
For [combinations_with_replacement](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.combinations_with_replacement) there is no pendant, e.g `array_combinations_with_replacement`.

mendelsshop:
This PR adds a new method for iterators called `array_combinations_with_replacement` which is similar to `combinations_with_replacement`, except for that it uses const generics and arrays. 

This is similar to `array_combinations` and `combinations`.
I also generalized the underlying `CombinationsWithReplacement` type similar to how it is done with `Combinations`.

This closes #1025.

### DIFF
--- a/src/combinations.rs
+++ b/src/combinations.rs
@@ -1,4 +1,3 @@
-#[cfg(feature = "use_alloc")]
 use alloc::boxed::Box;
 use core::array;
 use core::borrow::BorrowMut;
@@ -54,7 +53,6 @@ pub trait PoolIndex<T>: BorrowMut<[usize]> {
         self.borrow().len()
     }
 }
-#[cfg(feature = "use_alloc")]
 impl<T> PoolIndex<T> for Box<[usize]> {
     type Item = Vec<T>;
 
--- a/src/combinations_with_replacement.rs
+++ b/src/combinations_with_replacement.rs
@@ -1,4 +1,3 @@
-#[cfg(feature = "use_alloc")]
 use alloc::boxed::Box;
 use std::fmt;
 use std::iter::FusedIterator;
@@ -22,7 +21,6 @@ where
     first: bool,
 }
 
-#[cfg(feature = "use_alloc")]
 /// Iterator for `Box<[I]>` valued combinations_with_replacement returned by [`.combinations_with_replacement()`](crate::Itertools::combinations_with_replacement)
 pub type CombinationsWithReplacement<I> = CombinationsWithReplacementGeneric<I, Box<[usize]>>;
 /// Iterator for const generic combinations_with_replacement returned by [`.array_combinations_with_replacement()`](crate::Itertools::array_combinations_with_replacement)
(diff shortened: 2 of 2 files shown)


src/combinations_with_replacement.rs
```diff
@@ -73,39 +72,52 @@ where
...
+    Idx: PoolIndex<I::Item>,
 {
-    type Item = Vec<I::Item>;
+    type Item = Idx::Item;
 
     fn next(&mut self) -> Option<Self::Item> {
         if self.first {
             // In empty edge cases, stop iterating immediately
-            if !(self.indices.is_empty() || self.pool.get_next()) {
+            if !(core::borrow::Borrow::<[usize]>::borrow(&self.indices).is_empty()
```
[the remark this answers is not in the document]
mendelsshop: Weird, that work, when I originally tried doing that I got an error.

src/combinations_with_replacement.rs
```diff
@@ -24,6 +24,10 @@ where
 
 /// Iterator for `Box<[I]>` valued combinations_with_replacement returned by [`.combinations_with_replacement()`](crate::Itertools::combinations_with_replacement)
 pub type CombinationsWithReplacement<I> = CombinationsWithReplacementGeneric<I, Box<[usize]>>;
+/// Iterator for const generic combinations_with_replacement returned by [`.array_combinations_with_replacement()`](crate::Itertools::array_combinations_with_replacement)
```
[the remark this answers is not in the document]
mendelsshop: I don't know I just copied and modified the doc comment from two lines above.

src/combinations.rs
```diff
@@ -52,7 +52,16 @@ pub trait PoolIndex<T>: BorrowMut<[usize]> {
         self.borrow().len()
     }
 }
+impl<T> PoolIndex<T> for Box<[usize]> {
+    type Item = Vec<T>;
```
phimuemue, line 59: We can think about returning `Box<[T]>` later, but since we'd probably construct it via a `Vec`, `Vec` is fine.

src/combinations_with_replacement.rs
```diff
@@ -1,3 +1,4 @@
+#[cfg(feature = "use_alloc")]
```
phimuemue, line 1: I think `cfg(feature="use_alloc")` in `combinations_with_replacement.rs` is redundant because the module itself is only conditionally (see https://github.com/rust-itertools/itertools/blob/master/src/lib.rs#L101-L102). Please double-check and remove them if that's the case.

phimuemue (changes requested):
Thank you.

Can you limit the `cfg`s to where they are necessary please?

And by the way: Thanks for splitting up the PR into nicely diffable chunks, and for just amending instead of force-pushing and overwriting old commits.

src/combinations_with_replacement.rs
mendelsshop: Yeah I see, I was getting confused because of a missing `Box` import, when writing it.

phimuemue (approved):
Thank you for this.

PR #1037, opened 2025-07-01
feat: ExactSizeIterator impl for Chunks

Huterenok:
`ExactSizeIterator` impl `Chunks` and `IntoChunks::len` method. Also added `#[must_use]` for `Chunk`

### DIFF
--- a/src/groupbylazy.rs
+++ b/src/groupbylazy.rs
@@ -607,8 +607,8 @@ where
 
 impl<'a, I> ExactSizeIterator for Chunks<'a, I>
 where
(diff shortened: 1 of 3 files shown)


Huterenok:
CI failes with 3 issues:
* `semver-checks` complains about moved `CombinationsWithReplacement` but these changes don't relate to my PR
* `semver-checks` fails on minor check about new `#[must_use]` on `Chunk`. Should i remove this change from PR?
* `clippy` complains about not used `size_hint::div_ceil_scalar` but it's definitely used in `groupbylazy.rs` and locally it doesn't complain. What's the deal?

PR #1038, opened 2025-07-01
Implement `Debug` for remaining public types

psimonyi:
Implement `Debug` for the `ChunkBy`-related types.  Also derive it for two `Coalesce`-related types; I'm not sure they are actually exposed but the compiler thinks so, and it's easy to do.

There are no more public types without a `Debug` impl (at least a conditional one) so enable warning for missing `Debug` implementations.

### DIFF
--- a/src/adaptors/coalesce.rs
+++ b/src/adaptors/coalesce.rs
@@ -110,8 +110,10 @@ where
 {
 }
 
+#[derive(Debug)]
 pub struct NoCount;
 
+#[derive(Debug)]
 pub struct WithCount;
 
 pub trait CountItem<T> {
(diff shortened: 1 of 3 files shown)


psimonyi:
Neither of the CI check failures seem to be related to my changes:
- "CI / check" runs `RUSTFLAGS="--deny warnings" cargo clippy`, which is complaining about the format string in `write!` invocations in lib.rs. (note: Locally, I have to explicitly enable that warning with `cargo clippy -- --warn clippy::uninlined_format_args`.)
- "CI / semver-checks" is complaining about `itertools::structs::CombinationsWithReplacement` whose visibility is definitely not affected by my changes implementing `Debug`.

Issue #605, opened 2022-03-03
Make grouping maps container agnostic


farnoy:
Could we make the grouping operations generic so they're able to produce collections other than `std::collections::HashMap`?

I'm proposing turning

```rs
    fn into_group_map<K, V>(self) -> HashMap<K, Vec<V>>
```

into

```rs
    fn into_group_map<K, V, C>(self) -> C where C: FromIterator<(K, Vec<V>)>
```

And similarly with `into_grouping_map()` and that family of types.

My use case is to directly produce HashMaps from the hashbrown crate and not std, but this should enable different applications as well.

Since the HashMap itself is a key component of how the grouping is achieved, it's probably insufficient to use `FromIterator` (we don't want to allocate an intermediate collection). I want to start a conversation and see what others think.

msrd0:
I needed this so I build a version that has a trait implemented for both std's and hashbrown's map: https://github.com/rust-itertools/itertools/compare/master...msrd0:itertools:faa514f

As a proof of concept that works, but to be useful in the wild that trait probably needs to live outside of itertools in a small crate used and implemented by all the alternative map implementations in the wild.

Issue #322, opened 2018-12-10
Support different hash algorithms in `into_group_map`


hcpl:
### Motivation

Currently [`Itertools::into_group_map`](https://docs.rs/itertools/0.8.0/itertools/trait.Itertools.html#method.into_group_map) returns `HashMap<K, V>` which equals to `HashMap<K, V, RandomState>` with `RandomState` being https://doc.rust-lang.org/std/collections/hash_map/struct.RandomState.html.

Supporting `HashMap<K, V, S>` where `S: BuildHasher + Default` will allow `into_group_map` to interoperate with hash maps that use other hash algorithms, e.g. [`FnvHashMap`](https://docs.rs/fnv/1.0.6/fnv/type.FnvHashMap.html) or [`FxHashMap`](https://docs.rs/fxhash/0.2.1/fxhash/type.FxHashMap.html).

### Implementation issues

The problem is that naively adding an `S: BuildHasher + Default` type parameter like this:

<details>
<summary>src/group_map.rs</summary>

```diff
 #![cfg(feature = "use_std")]
 
 use std::collections::HashMap;
-use std::hash::Hash;
+use std::hash::{BuildHasher, Hash};
 use std::iter::Iterator;
 
 /// Return a `HashMap` of keys mapped to a list of their corresponding values.
 ///
 /// See [`.into_group_map()`](../trait.Itertools.html#method.into_group_map)
 /// for more information.
-pub fn into_group_map<I, K, V>(iter: I) -> HashMap<K, Vec<V>>
+pub fn into_group_map<I, K, V, S>(iter: I) -> HashMap<K, Vec<V>, S>
     where I: Iterator<Item=(K, V)>,
           K: Hash + Eq,
+          S: BuildHasher + Default,
 {
-    let mut lookup = HashMap::new();
+    let mut lookup = HashMap::default();
 
     for (key, val) in iter {
         lookup.entry(key).or_insert(Vec::new()).push(val);
     }
 
     lookup
}
```
</details>

<details>
<summary>src/lib.rs</summary>

```diff
@@ -60,7 +60,7 @@ use std::iter::{IntoIterator};
 use std::cmp::Ordering;
 use std::fmt;
 #[cfg(feature = "use_std")]
-use std::hash::Hash;
+use std::hash::{BuildHasher, Hash};
 #[cfg(feature = "use_std")]
 use std::fmt::Write;
 #[cfg(feature = "use_std")]
[...]

PR #1051, opened 2025-08-13
Support `hashbrown` hash maps and sets with or without `use_std`

aqrln:
Add an optional `hashbrown` feature that switches the underlying hash map implementation from `std::collections` to `hashbrown` and makes items that depend on hash maps available without `use_std` (although it still requires and implicitly enables `use_alloc`).

This addresses the motivating use case of [#605][] (which is the same use case that I am interested in) but does not implement a generic approach as the issue suggests. [#901][] would be a better solution long term and would make these changes obsolete, but the PR appears to be stalled and I wanted a simple solution that allows using APIs like `unique`/`unique_by` with `no_std` now.

It also partially addresses [#322][] because enabling this feature does switch to *a* faster hash algorithm, it just doesn't allow to specify which one exactly (currently the default hasher in `hashbrown` is `foldhash`, so this is what's going to be used until/unless changed upstream in the future).

[#605]: https://github.com/rust-itertools/itertools/issues/605
[#901]: https://github.com/rust-itertools/itertools/pull/901
[#322]: https://github.com/rust-itertools/itertools/issues/322

--- 

TODO:
- [ ] fix tests compilation with `--all-features`/`-F hashbrown`
- [ ] run tests on CI with both hash map implementations
- [ ] adjust contributor doc in the section about `use_std`

I'm opening a draft PR early because I'd like to get feedback on whether it's something you want to have.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,10 @@ test = false
 
 [dependencies]
 either = { version = "1.0", default-features = false }
+hashbrown = { version = "0.15", optional = true, default-features = false, features = [
+    "alloc",
+    "default-hasher",
+] }
 
 [dev-dependencies]
 rand = "0.7"
@@ -36,6 +40,7 @@ quickcheck = { version = "0.9", default-features = false }
 default = ["use_std"]
 use_std = ["use_alloc", "either/use_std"]
 use_alloc = []
+hashbrown = ["dep:hashbrown", "use_alloc"]
 
 [profile]
 bench = { debug = true }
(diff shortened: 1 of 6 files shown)


cuviper:
This could be ok for internal use, but switching the public API based on a feature makes it a conditionally-breaking change. For example, one dependent might use this expecting the `std` type, while another dependent enables this `hashbrown` feature that changes it for everyone.

It would be safer to make this change unconditional (with a semver bump).

aqrln:
That makes sense, although it might arguably make the common case more surprising. It would be much more valuable to find a way to be generic over containers anyhow. Happy to finish this work if you think it's useful to switch to `hashbrown` completely, otherwise it's fine to close this PR. I ended up eventually not needing this myself for the time being.

PR #1079, opened 2025-12-05
Fix `pad_using` bug producing items when it shouldn't

### LINKED ISSUE #1065: Question about `pad_using`
Reading the docs for `pad_using`:

> Return an iterator adaptor that pads the sequence to a minimum length of min by filling missing elements using a closure f.

And the examples:

```rust
let it = (0..5).pad_using(10, |i| 2*i);
itertools::assert_equal(it, vec![0, 1, 2, 3, 4, 10, 12, 14, 16, 18]);
```

It seems like if the padding is greater than the iterator's length, it's padded up to a minimum of those items with the given function. That makes sense.

I tried out `pad_using`, and padded to an iterator of [1]:

```rust
use itertools::Itertools;

#[test]
fn pad_tail_bug() {
    let mut iter = vec![].into_iter().pad_using(1, |_| 1);

    itertools::assert_equal(iter.clone(), vec![1]);

    assert_eq!(iter.next(), Some(1));
    assert_eq!(iter.next_back(), None);
}
```

This is the error message I get running this test:

```rust
running 1 test

thread 'pad_tail_bug' panicked at tests/pad_using_bug.rs:10:5:
assertion `left == right` failed
  left: Some(1)
 right: None
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
test pad_tail_bug ... FAILED

failures:

failures:
    pad_tail_bug

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

error: test failed, to rerun pass `--test pad_using_bug`
```

I would assume that calling `next` and then `next_back` should return `Some(1)` for an iterator with only one element and then the next item should be None, no?

Takashiidobe:
I added a test to the double ended iterator specialization that catches this bug in `pad_using`, where if exhaust the iterator from either calling `next` or `next_back` and then call the other pair, it would still produce elements.

If you just add that test, you'll see `pad_using` fails. Next, I fixed `pad_using` to keep track of how many items from the back it's taken, so we can use that to correctly exhaust the iterator when needed.

~~Question: since this adds fields + debug formatted fields to a public struct, this is a breaking change, right?~~ Apparently not.

Fixes: #1065

### DIFF
--- a/src/pad_tail.rs
+++ b/src/pad_tail.rs
@@ -1,4 +1,3 @@
-use crate::size_hint;
 use std::iter::{Fuse, FusedIterator};
 
 /// An iterator adaptor that pads a sequence to a minimum length by filling
@@ -13,6 +12,8 @@ pub struct PadUsing<I, F> {
     iter: Fuse<I>,
     min: usize,
     pos: usize,
+    back: usize,
+    total_len: usize,
     filler: F,
 }
 
@@ -20,7 +21,7 @@ impl<I, F> std::fmt::Debug for PadUsing<I, F>
 where
     I: std::fmt::Debug,
 {
-    debug_fmt_fields!(PadUsing, iter, min, pos);
+    debug_fmt_fields!(PadUsing, iter, min, pos, back, total_len);
 }
 
 /// Create a new `PadUsing` iterator.
@@ -33,6 +34,8 @@ where
         iter: iter.fuse(),
         min,
         pos: 0,
+        back: 0,
+        total_len: min,
         filler,
     }
 }
@@ -48,7 +51,7 @@ where
     fn next(&mut self) -> Option<Self::Item> {
         match self.iter.next() {
             None => {
-                if self.pos < self.min {
+                if self.pos + self.back < self.total_len {
                     let e = Some((self.filler)(self.pos));
                     self.pos += 1;
                     e
@@ -64,8 +67,18 @@ where
     }
 
     fn size_hint(&self) -> (usize, Option<usize>) {
-        let tail = self.min.saturating_sub(self.pos);
-        size_hint::max(self.iter.size_hint(), (tail, Some(tail)))
+        let (iter_lower, iter_upper) = self.iter.size_hint();
+        let consumed = self.pos.saturating_add(self.back);
+
+        let total_lower = iter_lower.saturating_add(self.pos).max(self.min);
+        let lower_bound = total_lower.saturating_sub(consumed);
+
+        let upper_bound = iter_upper.map(|iter_upper| {
+            let total_upper = iter_upper.saturating_add(self.pos).max(self.min);
+            total_upper.saturating_sub(consumed)
+        });
+
+        (lower_bound, upper_bound)
     }
 
     fn fold<B, G>(self, mut init: B, mut f: G) -> B
@@ -87,14 +100,25 @@ where
     F: FnMut(usize) -> I::Item,
 {
     fn next_back(&mut self) -> Option<Self::Item> {
-        if self.min == 0 {
-            self.iter.next_back()
-        } else if self.iter.len() >= self.min {
-            self.min -= 1;
-            self.iter.next_back()
+        let current_iter_len = self.iter.len();
+        let original_iter_len = current_iter_len.saturating_add(self.pos);
+        if self.total_len < original_iter_len {
+            self.total_len = original_iter_len;
+        }
+
+        if self.pos + self.back >= self.total_len {
+            return None;
+        }
+
+        let padding_count = self.total_len.saturating_sub(current_iter_len + self.pos);
+
+        if self.back < padding_count {
+            let idx = self.total_len - self.back - 1;
+            self.back += 1;
+            Some((self.filler)(idx))
         } else {
-            self.min -= 1;
-            Some((self.filler)(self.min))
+            self.back += 1;
+            self.iter.next_back()
         }
     }
 
@@ -102,10 +126,26 @@ where
     where
         G: FnMut(B, Self::Item) -> B,
     {
-        init = (self.iter.len()..self.min)
-            .map(self.filler)
-            .rfold(init, &mut f);
-        self.iter.rfold(init, f)
+        let PadUsing {
+            iter,
+            min: _,
+            pos,
+            back,
+            mut total_len,
+            filler,
+        } = self;
+        let iter_len = iter.len();
+        let original_iter_len = iter_len.saturating_add(pos);
+        if total_len < original_iter_len {
+            total_len = original_iter_len;
+        }
+
+        let start_idx = iter_len + pos;
+        let end_idx = total_len - back;
+
+        init = (start_idx..end_idx).rev().map(filler).fold(init, &mut f);
+
+        iter.rfold(init, f)
     }
 }
 
--- a/src/pad_tail.rs
+++ b/src/pad_tail.rs
@@ -51,7 +51,6 @@ where
 {
     type Item = I::Item;
 
-    #[inline]
     fn next(&mut self) -> Option<Self::Item> {
         let total_consumed = self.elements_from_next + self.elements_from_next_back;
 
@@ -82,21 +81,6 @@ where
 
         (lower_bound, upper_bound)
     }
-
-    fn fold<B, G>(self, mut init: B, mut f: G) -> B
-    where
-        G: FnMut(B, Self::Item) -> B,
-    {
-        let mut start = self.elements_from_next;
-        init = self.iter.fold(init, |acc, item| {
-            start += 1;
-            f(acc, item)
-        });
-
-        let end = self.elements_required - self.elements_from_next_back;
-
-        (start..end).map(self.filler).fold(init, f)
-    }
 }
 
 impl<I, F> DoubleEndedIterator for PadUsing<I, F>
@@ -111,29 +95,14 @@ where
             return self.iter.next_back();
         }
 
-        let elements_remaining = self.elements_required - total_consumed;
+        let index_from_back = self.elements_required - self.elements_from_next_back - 1;
         self.elements_from_next_back += 1;
 
-        if self.iter.len() < elements_remaining {
-            Some((self.filler)(
-                self.elements_required - self.elements_from_next_back,
-            ))
+        if index_from_back >= self.iter.len() {
+            Some((self.filler)(index_from_back))
         } else {
             self.iter.next_back()
         }
-    }
-
-    fn rfold<B, G>(self, mut init: B, mut f: G) -> B
-    where
-        G: FnMut(B, Self::Item) -> B,
-    {
-        let start = self.iter.len() + self.elements_from_next;
-        let remaining = self.elements_required.max(start);
-        let end = remaining - self.elements_from_next_back;
-
-        init = (start..end).rev().map(self.filler).fold(init, &mut f);
-
-        self.iter.rfold(init, f)
     }
 }
 
--- a/src/pad_tail.rs
+++ b/src/pad_tail.rs
@@ -87,12 +87,15 @@ where
     where
         G: FnMut(B, Self::Item) -> B,
     {
-        let mut pos = self.elements_from_next;
+        let mut start = self.elements_from_next;
         init = self.iter.fold(init, |acc, item| {
-            pos += 1;
+            start += 1;
             f(acc, item)
         });
-        (pos..self.elements_required).map(self.filler).fold(init, f)
+
+        let end = self.elements_required - self.elements_from_next_back;
+
+        (start..end).map(self.filler).fold(init, f)
     }
 }
 
--- a/src/pad_tail.rs
+++ b/src/pad_tail.rs
@@ -53,40 +53,32 @@ where
 
     #[inline]
     fn next(&mut self) -> Option<Self::Item> {
-        match self.iter.next() {
-            None => {
-                if self.elements_from_next + self.elements_from_next_back < self.elements_required {
-                    let e = Some((self.filler)(self.elements_from_next));
-                    self.elements_from_next += 1;
-                    e
-                } else {
-                    None
-                }
-            }
-            e => {
-                self.elements_from_next += 1;
-                e
-            }
+        let total_consumed = self.elements_from_next + self.elements_from_next_back;
+
+        if total_consumed >= self.elements_required {
+            self.iter.next()
+        } else if let Some(e) = self.iter.next() {
+            self.elements_from_next += 1;
+            Some(e)
+        } else {
+            let e = (self.filler)(self.elements_from_next);
+            self.elements_from_next += 1;
+            Some(e)
         }
     }
 
     fn size_hint(&self) -> (usize, Option<usize>) {
-        let (iter_lower, iter_upper) = self.iter.size_hint();
-        let consumed = self
-            .elements_from_next
-            .saturating_add(self.elements_from_next_back);
+        let total_consumed = self.elements_from_next + self.elements_from_next_back;
 
-        let total_lower = iter_lower
-            .saturating_add(self.elements_from_next)
-            .max(self.elements_required);
-        let lower_bound = total_lower.saturating_sub(consumed);
+        if total_consumed >= self.elements_required {
+            return self.iter.size_hint();
+        }
 
-        let upper_bound = iter_upper.map(|iter_upper| {
-            let total_upper = iter_upper
-                .saturating_add(self.elements_from_next)
-                .max(self.elements_required);
-            total_upper.saturating_sub(consumed)
-        });
+        let elements_remaining = self.elements_required - total_consumed;
+        let (low, high) = self.iter.size_hint();
+
+        let lower_bound = low.max(elements_remaining);
+        let upper_bound = high.map(|h| h.max(elements_remaining));
 
         (lower_bound, upper_bound)
     }
@@ -112,11 +104,11 @@ where
     fn next_back(&mut self) -> Option<Self::Item> {
         let total_consumed = self.elements_from_next + self.elements_from_next_back;
 
-        if self.iter.len() == 0 && total_consumed >= self.elements_required {
-            return None;
+        if total_consumed >= self.elements_required {
+            return self.iter.next_back();
         }
 
-        let elements_remaining = self.elements_required.saturating_sub(total_consumed);
+        let elements_remaining = self.elements_required - total_consumed;
         self.elements_from_next_back += 1;
 
         if self.iter.len() < elements_remaining {
@@ -124,9 +116,7 @@ where
                 self.elements_required - self.elements_from_next_back,
             ))
         } else {
-            let e = self.iter.next_back();
-            assert!(e.is_some());
-            e
+            self.iter.next_back()
         }
     }
 
@@ -134,25 +124,13 @@ where
     where
         G: FnMut(B, Self::Item) -> B,
     {
-        let PadUsing {
-            iter,
-            elements_from_next,
-            elements_from_next_back,
-            mut elements_required,
-            filler,
-        } = self;
-        let iter_len = iter.len();
-        let original_iter_len = iter_len.saturating_add(elements_from_next);
-        if elements_required < original_iter_len {
-            elements_required = original_iter_len;
-        }
+        let start = self.iter.len() + self.elements_from_next;
+        let remaining = self.elements_required.max(start);
+        let end = remaining - self.elements_from_next_back;
 
-        let start_idx = iter_len + elements_from_next;
-        let end_idx = elements_required - elements_from_next_back;
+        init = (start..end).rev().map(self.filler).fold(init, &mut f);
 
-        init = (start_idx..end_idx).rev().map(filler).fold(init, &mut f);
-
-        iter.rfold(init, f)
+        self.iter.rfold(init, f)
     }
 }
 
--- a/src/pad_tail.rs
+++ b/src/pad_tail.rs
@@ -10,10 +10,9 @@ use std::iter::{Fuse, FusedIterator};
 #[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
 pub struct PadUsing<I, F> {
     iter: Fuse<I>,
-    min: usize,
-    pos: usize,
-    back: usize,
-    total_len: usize,
+    elements_from_next: usize,
+    elements_from_next_back: usize,
+    elements_required: usize,
     filler: F,
 }
 
@@ -21,21 +20,26 @@ impl<I, F> std::fmt::Debug for PadUsing<I, F>
 where
     I: std::fmt::Debug,
 {
-    debug_fmt_fields!(PadUsing, iter, min, pos, back, total_len);
+    debug_fmt_fields!(
+        PadUsing,
+        iter,
+        elements_from_next,
+        elements_from_next_back,
+        elements_required
+    );
 }
 
 /// Create a new `PadUsing` iterator.
-pub fn pad_using<I, F>(iter: I, min: usize, filler: F) -> PadUsing<I, F>
+pub fn pad_using<I, F>(iter: I, elements_required: usize, filler: F) -> PadUsing<I, F>
 where
     I: Iterator,
     F: FnMut(usize) -> I::Item,
 {
     PadUsing {
         iter: iter.fuse(),
-        min,
-        pos: 0,
-        back: 0,
-        total_len: min,
+        elements_from_next: 0,
+        elements_from_next_back: 0,
+        elements_required,
         filler,
     }
 }
@@ -51,16 +55,16 @@ where
     fn next(&mut self) -> Option<Self::Item> {
         match self.iter.next() {
             None => {
-                if self.pos + self.back < self.total_len {
-                    let e = Some((self.filler)(self.pos));
-                    self.pos += 1;
+                if self.elements_from_next + self.elements_from_next_back < self.elements_required {
+                    let e = Some((self.filler)(self.elements_from_next));
+                    self.elements_from_next += 1;
                     e
                 } else {
                     None
                 }
             }
             e => {
-                self.pos += 1;
+                self.elements_from_next += 1;
                 e
             }
         }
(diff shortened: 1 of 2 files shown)


jswrenn:
Good catch! This isn't a breaking change since the added fields are private. In theory, we might have users depending on the current unintended behavior. Fortunately, we already having a breaking change on `master`, so our next release will be SemVer breaking, anyways.

src/pad_tail.rs
```diff
@@ -13,14 +12,16 @@ pub struct PadUsing<I, F> {
     iter: Fuse<I>,
     min: usize,
     pos: usize,
+    back: usize,
+    total_len: usize,
```
phimuemue: Wouldn't this suffice:

```
struct PadUsing {
 iter,
 filler,
 elements_from_next: usize, // no. of elements generated by calls to `next`. That's what's currently pos
 elements_from_next_back: usize, // no. of elements generated by calls to `next_back`
 elements_required: usize, // no. of elements required by pad_using caller
}
```

In my naive view, this should be sufficient to determine at each step what to generate next. (And it avoids the `min`/`total_len` dichotomy).

src/pad_tail.rs
```diff
@@ -48,7 +51,7 @@ where
     fn next(&mut self) -> Option<Self::Item> {
         match self.iter.next() {
             None => {
-                if self.pos < self.min {
+                if self.pos + self.back < self.total_len {
```
phimuemue: I like this line, especially with my proposed refactor: `if self.elements_from_next + self.elements_from_next_back < self.elements_required`, then we generate another one.

src/pad_tail.rs
```diff
@@ -87,25 +100,52 @@ where
...
+        if self.pos + self.back >= self.total_len {
+            return None;
+        }
+
+        let padding_count = self.total_len.saturating_sub(current_iter_len + self.pos);
+
+        if self.back < padding_count {
+            let idx = self.total_len - self.back - 1;
+            self.back += 1;
+            Some((self.filler)(idx))
```
phimuemue: I'd expect `next_back` to look sth like this:

```
if elements_from_next + elements_from_next_back < elements_required { // Note: Same condition as in `fn next`
 let elements_remaining = elements_required - (elements_from_next + elements_from_next_back);
 elements_from_next_back += 1;
 if iter.len() < elements_remaining {
  Some((self.filler)(elements_required - elements_from_next_back)) // TODO Is the index computed correctly?
 } else {
  let e = iter.next_back();
  assert!(e.is_some); // If this triggers, incoming ExactSizeIterator is implemented incorrectly. I'd 
  e
 }
} else {
 iter.next_back()
}
```

This seems simpler to me.

phimuemue (commented):
Hi there, thanks for this. Can we simplify your proposal?

In particular, `saturating_[add|sub]` in `next_back` makes me nervous because it introduces implicit special cases that are hard to reason about.

Do my suggestions make sense?

src/pad_tail.rs
Takashiidobe: Makes sense to me, applied

src/pad_tail.rs
Takashiidobe: applied this change

src/pad_tail.rs
Takashiidobe: This is close but it still fails tests cause it doesn't compute the index properly.

I fixed it up and tested it like this
```rust
if self.elements_from_next + self.elements_from_next_back < self.elements_required {
    // Note: Same condition as in `fn next`
    let elements_remaining =
        self.elements_required - (self.elements_from_next + self.elements_from_next_back);
    self.elements_from_next_back += 1;
    if self.iter.len() < elements_remaining {
        Some((self.filler)(
            self.elements_required - self.elements_from_next_back,
        ))
    } else {
        let e = self.iter.next_back();
        assert!(e.is_some()); // If this triggers, incoming ExactSizeIterator is implemented incorrectly. I'd
        e
    }
} else {
    None
}
```

I rewrote your code like this, but it still does have the saturating sub. Tests will fail without it so I'm not so sure what to do there.

```rust
let total_consumed = self.elements_from_next + self.elements_from_next_back;

if self.iter.len() == 0 && total_consumed >= self.elements_required {
    return None;
}

let elements_remaining = self.elements_required.saturating_sub(total_consumed);
self.elements_from_next_back += 1;

if self.iter.len() < elements_remaining {
    Some((self.filler)(
        self.elements_required - self.elements_from_next_back,
    ))
} else {
    let e = self.iter.next_back();
    assert!(e.is_some());
    e
}
```

Takashiidobe: One thing I'd also like to mention: there are sources of `saturating_*` in the codebase outside of `pad_tail`. 

I think it would also be an improvement to have tests that can trigger  cases that check for saturating ops saturating when they shouldn't if it's a concern for code reviews, since it would help the dev cycle. 

Trying to write this `pad_using` properly, there's no way to do this without really good tests, this is really difficult (at least for me).

```sh
$ rg "saturating_"
tests/test_std.rs
501:        let mut v: Vec<_> = (n..n.saturating_add(m as _)).collect();
537:        let mut v: Vec<_> = (n..n.saturating_add(m as _)).collect();
1241:            let len = binomial(usize::saturating_sub(n + k, 1), k);

tests/quick.rs
63:            org_lower.saturating_sub(self.underestimate),
1985:        let result = &v[v.len().saturating_sub(n)..];

src/pad_tail.rs
77:            .saturating_add(self.elements_from_next_back);
80:            .saturating_add(self.elements_from_next)
82:        let lower_bound = total_lower.saturating_sub(consumed);
86:                .saturating_add(self.elements_from_next)
88:            total_upper.saturating_sub(consumed)
119:        let elements_remaining = self.elements_required.saturating_sub(total_consumed);
145:        let original_iter_len = iter_len.saturating_add(elements_from_next);

src/size_hint.rs
11:    let min = a.0.saturating_add(b.0);
24:    low = low.saturating_add(x);
33:    low = low.saturating_sub(x);
34:    hi = hi.map(|elt| elt.saturating_sub(x));
41:    let low = a.0.saturating_mul(b.0);
54:    low = low.saturating_mul(x);

src/iter_index.rs
59:            (self.end() + 1).saturating_sub(*self.start())

src/peek_nth.rs
72:        let unbuffered_items = (n + 1).saturating_sub(self.buf.len());
113:        let unbuffered_items = (n + 1).saturating_sub(self.buf.len());

src/sources.rs
25:///     let next = x1.saturating_add(*x2);

src/lib.rs
3673:                let mut iter = self.fuse().skip(low.saturating_sub(n));

src/combinations_with_replacement.rs
177:            k.saturating_sub(1)
```

phimuemue: > One thing I'd also like to mention: there are sources of `saturating_*` in the codebase outside of `pad_tail`.

True, but we should strive to avoid them if not strictly necessary.

phimuemue: > Trying to write this `pad_using` properly [...], this is really difficult (at least for me).

For me, too. But we must not patch the code here and there until all of our tests succeed.

Instead, we should look for a simple explanation how `pad_using` works. My explanation looks like this:
- `iter.pad_using(n, ...)` will generate *at least* `n` elements.
- Iff `iter` itself generates `n` or more elements, `filler` will not supply elements.
  => If our iterator already generated more than `n` elements, we know that we must not call `filler` anymore (because more than `n` elements means that all of them come from `iter`).
- As you correctly observed in your original PR, counting `next`ed and `next_back`ed elements makes things easier.

I tried to execute this idea in code, and fwiw, `fn next_back` (as in my edited comment) successfully passes the tests on my machine:

```
        if self.elements_from_next + self.elements_from_next_back < self.elements_required {
            // Note: Same condition as in `fn next`
            let elements_remaining =
                self.elements_required - (self.elements_from_next + self.elements_from_next_back);
            self.elements_from_next_back += 1;
            if self.iter.len() < elements_remaining {
                Some((self.filler)(
                        self.elements_required - self.elements_from_next_back,
                ))
            } else {
                let e = self.iter.next_back();
                assert!(e.is_some()); // If this triggers, incoming ExactSizeIterator is implemented incorrectly. I'd
                e
            }
        } else {
            self.iter.next_back()
        }
```

Of course I'm biased, but to me my `next_back` looks easier than your rewritten function. [^1][^2]

Thinking of it, we may even apply similar logic (i.e. check if we so far generated less than `elements_required` elements and behave accordingly) for `fn next`.

[^1]: Mine has no `saturating_sub`. `saturating_sub` begs the question when *exactly* the operation will actually saturate and when not.
[^2]: Why exactly does yours test `self.iter.len() == 0 && total_consumed >= self.elements_required`? In particular, is `len==0` redundant (because in this case just calling `next_back` would return `None` anyway)?

phimuemue: Speaking of which: When we establish the `elements_from_next + elements_from_next_back < elements_required` thing, couldn't we also use it to straightforwardly implement other operations?

```
fn size_hint(self) {
 if (elements_from_next + elements_from_next_back < elements_required) {
  let elements_remaining = ...;
  let (lo, hi) = iter.size_hint();
  if (lo >= elements_remaining) {
   // iterator itself will supply all remaining elements
   (lo, hi)
  } else {
   // both iterator and filler may supply remaining elements
   ... // TODO
  }
 } else {
  // won't touch filler anymore
  iter.size_hint()
 }
}
```

I know that explicitly listing relevant cases may look ugly, but this way you can keep the simple cases *really* simple, and constrain the complexity to the other cases. I consider this better than relying on several calls to `saturating_[add|sub]` that *still* have case distinctions - but *implicit* ones.

Takashiidobe: Ah I see the edit, the else branch was supposed to be `next_back` instead of None. That makes sense. With that, my approach looks like this:

```rust
    let total_consumed = self.elements_from_next + self.elements_from_next_back;

    if total_consumed >= self.elements_required {
        return self.iter.next_back();
    }

    let elements_remaining = self.elements_required - total_consumed;
    self.elements_from_next_back += 1;

    if self.iter.len() < elements_remaining {
        Some((self.filler)(
            self.elements_required - self.elements_from_next_back,
        ))
    } else {
        self.iter.next_back()
    }
}
```

It's the same implementation except I prefer guard clauses at the top, so I put the else branch at the top to short circuit, and there are no more saturating ops.

For next, I couldn't do exactly the same pattern cause `len` is only for ExactSizedIterator so it's slightly different but the same idea.

```rust
    let total_consumed = self.elements_from_next + self.elements_from_next_back;

    if total_consumed >= self.elements_required {
        self.iter.next()
    } else if let Some(e) = self.iter.next() {
        self.elements_from_next += 1;
        Some(e)
    } else {
        let e = (self.filler)(self.elements_from_next);
        self.elements_from_next += 1;
        Some(e)
    }
}
```

For `size_hint`, I swapped the saturating ops with max. No saturation required because there's no risk of overflow.

```rust
    let total_consumed = self.elements_from_next + self.elements_from_next_back;

    if total_consumed >= self.elements_required {
        return self.iter.size_hint();
    }

    let elements_remaining = self.elements_required - total_consumed;
    let (low, high) = self.iter.size_hint();

    let lower_bound = low.max(elements_remaining);
    let upper_bound = high.map(|h| h.max(elements_remaining));

    (lower_bound, upper_bound)
}
```

src/pad_tail.rs
```diff
@@ -46,38 +53,46 @@ where
...
         G: FnMut(B, Self::Item) -> B,
     {
-        let mut pos = self.pos;
+        let mut pos = self.elements_from_next;
         init = self.iter.fold(init, |acc, item| {
             pos += 1;
             f(acc, item)
         });
-        (pos..self.min).map(self.filler).fold(init, f)
+        (pos..self.elements_required).map(self.filler).fold(init, f)
```
phimuemue: If next_back has already been called, should we really go until elements_required?

Takashiidobe: Yup, I forgot to edit this like I did for `rfold`. I see `rfold` seems to have testing whereas `fold` does not, so I'm going to add some test cases for it in specializations.

Takashiidobe: I've added a block that tests for `fold` just before the `rfold` block in `test_double_ended_specializations`. With the previous (incorrect) implementation:

```rust
let mut pos = self.elements_from_next;
init = self.iter.fold(init, |acc, item| {
    pos += 1;
    f(acc, item)
});

(pos..self.elements_required).map(self.filler).fold(init, f)
```

All the tests passed (so I missed this cause I just find and replaced `self.min` with `self.elements_from_next`).

The test I added in `test_double_ended_specializations` will catch this bug: 

```rust
check_specialized!(it, |i| {
    let mut parameters_from_fold = vec![];
    let fold_result = i.fold(vec![], |mut acc, v: I::Item| {
        parameters_from_fold.push((acc.clone(), v.clone()));
        acc.push(v);
        acc
    });
    (parameters_from_fold, fold_result)
});
```

With the fixed version, 

```rust
init = self.iter.fold(init, |acc, item| {
    start += 1;
    f(acc, item)
});

let end = self.elements_required - self.elements_from_next_back;

(start..end).map(self.filler).fold(init, f)
```

The tests all pass again, since accounts for calls to `next_back` as well as `next` for `fold` just like for `rfold`.

tests/specializations.rs
```diff
@@ -131,6 +131,25 @@ where
     for n in 0..size + 2 {
         check_specialized!(it, |mut i| i.nth_back(n));
     }
+
+    let mut fwd = it.clone();
+    let mut bwd = it.clone();
+
+    for _ in fwd.by_ref() {}
```
phimuemue: Plase make this `while fwd.next_back.is_some` (just as with `bwd`).

src/pad_tail.rs
```diff
@@ -46,38 +53,49 @@ where
...
+
+        if total_consumed >= self.elements_required {
+            self.iter.next()
+        } else if let Some(e) = self.iter.next() {
+            self.elements_from_next += 1;
+            Some(e)
+        } else {
+            let e = (self.filler)(self.elements_from_next);
+            self.elements_from_next += 1;
+            Some(e)
```
phimuemue, line 65: Looks very nice imo. It tells
- `total_consumed` does not overcount `elements_required`, which (presumably) avoids overflows.

src/pad_tail.rs
```diff
@@ -46,38 +53,49 @@ where
...
     }
 
     fn size_hint(&self) -> (usize, Option<usize>) {
-        let tail = self.min.saturating_sub(self.pos);
-        size_hint::max(self.iter.size_hint(), (tail, Some(tail)))
+        let total_consumed = self.elements_from_next + self.elements_from_next_back;
+
+        if total_consumed >= self.elements_required {
+            return self.iter.size_hint();
+        }
```
phimuemue, line 74: Nice and simple. Don't overcount `elements_required`, similar to `next`.

src/pad_tail.rs
```diff
@@ -87,24 +105,34 @@ where
...
-            self.iter.next_back()
-        } else if self.iter.len() >= self.min {
-            self.min -= 1;
-            self.iter.next_back()
+        let total_consumed = self.elements_from_next + self.elements_from_next_back;
+
+        if total_consumed >= self.elements_required {
+            return self.iter.next_back();
+        }
+
```
phimuemue, line 97: Again, nice. Similar to `next`.

src/pad_tail.rs
```diff
@@ -87,24 +105,34 @@ where
...
+        self.elements_from_next_back += 1;
+
+        if self.iter.len() < elements_remaining {
+            Some((self.filler)(
+                self.elements_required - self.elements_from_next_back,
+            ))
         } else {
-            self.min -= 1;
-            Some((self.filler)(self.min))
+            self.iter.next_back()
```
phimuemue, line 104: Please `assert` that the result of `next_back` `is_some`. Otherwise incrementing `elements_from_next_back` does not make sense and we'd possibly overcount `elements_required`.

You could even think about incrementing in each branch explicitly (just as you did in `next`). (Or adjust `next` to match `next_back`: Aligned logic in `next` and `next_back` might be easier to understand.)

src/pad_tail.rs
```diff
@@ -87,24 +105,34 @@ where
...
     fn rfold<B, G>(self, mut init: B, mut f: G) -> B
     where
         G: FnMut(B, Self::Item) -> B,
     {
-        init = (self.iter.len()..self.min)
-            .map(self.filler)
-            .rfold(init, &mut f);
+        let start = self.iter.len() + self.elements_from_next;
+        let remaining = self.elements_required.max(start);
+        let end = remaining - self.elements_from_next_back;
```
phimuemue: I don't understand this. I could not find counterexamples, but the logic here seems to differ from everything else you wrote.

Especially: Why is `remaining = elements_required.max(start)`? Trying to untangle the - implicit - case distinction incurred by `max`:

```
let start = iter.len() + elements_from_next;
if elements_required < start {
 let end = start - elements_from_next_back;
 init = (start..end).rev().map(self.filler).fold(init, &mut f);
} else {
 let end = elements_required - elements_from_next_back;
 init = (start..end).rev().map(self.filler).fold(init, &mut f);
}
```

If `elements_required < start`, doesn't this mean that the `iter` itself will be able to generate more elements than `elements_required`? If so, we don't even need to touch `filler` in that case, and we can simplify to:

```
let start = iter.len() + elements_from_next;
if start <= elements_required {
 let end = elements_required - elements_from_next_back;
 init = (start..end).rev().map(self.filler).fold(init, &mut f);
}
```

There's another implicit case distinction: `start..end` is empty if `start>=end`. Making this case explicit:

```
let start = iter.len() + elements_from_next;
if start <= elements_required {
 let end = elements_required - elements_from_next_back;
 if (start < end) {
  init = (start..end).rev().map(self.filler).fold(init, &mut f);
 }
}
```

Expanding and reformulating the condition `start < end` gives this:

- `start < end`
- `iter.len() + elements_from_next < elements_required - elements_from_next_back`
- `iter.len() + elements_from_next + elements_from_next_back < elements_required`

Now this brings us to known territory - much closer to what you already have in `next[_back]`. Also note that the last line alone is already a stronger condition than the initial `start <= elements_required`, so `start <= elements_required` is not needed.

src/pad_tail.rs
```diff
@@ -46,38 +53,49 @@ where
...
 
     fn fold<B, G>(self, mut init: B, mut f: G) -> B
     where
         G: FnMut(B, Self::Item) -> B,
     {
-        let mut pos = self.pos;
+        let mut start = self.elements_from_next;
         init = self.iter.fold(init, |acc, item| {
-            pos += 1;
+            start += 1;
```
phimuemue: Nitpick: Would this overflow if `iter` itself generated more than `usize::MAX` elements? And if so, would we end up mistakenly calling `filler` at the end of `fold`?

I'm not sure there's a straightforward formulation that would avoid the problem[^1]. I'd be willing to accept it.

[^1]: Apart from messing with `checked_add`, which is sth we do not do often.

phimuemue (changes requested):
Hi there, thanks for pulling this through. I like the idea of not overcounting `elements_required`.

Please either rewrite `rfold` so that it follows the other methods' logic or remove it altogether (it's better to be (explainably) correct than to be fast).

We're not in a rush, so in case of doubt, please take time and make sure all methods follow a coherent pattern (or state why they can't).

Takashiidobe:
Here's the PR once again, with everything checked. I ended up removing `fold` and `rfold` since those should both just call `next` and `next_back` respectively, so as long as `next` and `next_back` are correctly implemented, we're good to go. `size_hint` still needs to be provided so I left that as is.

Here's my explanation for `next` vs `next_back` and why my final implementation of `next_back` turns the way it is. Keep in mind I like the guard clause style here (think of exceptional situations first so you don't forget them later) because that also removes an indent afterwards.

First off is `next`:

There are 3 cases:
1. You're between 0..self.iter.len()
2. You're between self.iter.len()..padded_size
3. You're anywhere from padded_size..usize::MAX.

Kind of visualized like this:

```
|-------------------|------------------------------------|...=usize::MAX
         ^                            ^                     ^
  0..self.iter.len()   self.iter.len()..padding            OOB
```

I do the exceptional case first, so I took 3 first. If the count of calls to next + next_back >= the padded size, you're in the OOB case. So we return `self.iter.next()`.

Next up is either the padding case or iterator case. Thankfully, you can just check `self.iter.next()`. If you get a Some(T), you know you must be between `0..self.iter.len()`, because otherwise the iterator would be exhausted, indicating you need to be in the padding area (because the OOB case is taken care of above). So, in the case you get back something, you return that (since we want to keep using the original iterator) and increment the items we've received from next by one. 

In the other case, if the iterator gives back None, then you need to provided a padded element: in which case, call the filler.

In summary:

1. If OOB: return self.iter.next()
2. If self.iter has something: incr next calls and return that
3. otherwise, incr next calls and return from filler

You can do the guard clause style here where you'd return early, but I just squashed all three cases so I didn't have to write `return self.iter.next()` for the OOB case. That's how I got to this final `next`:

```rust
    let total_consumed = self.elements_from_next + self.elements_from_next_back;

    if total_consumed >= self.elements_required {
        self.iter.next()
    } else if let Some(e) = self.iter.next() {
        self.elements_from_next += 1;
        Some(e)
    } else {
        let e = (self.filler)(self.elements_from_next);
        self.elements_from_next += 1;
        Some(e)
    }
}
```

`next_back` is trickier: If you copy paste the same logic and flip the logic (since you're approaching from the back) you get this:

First, we calculate our current index by doing `padded_size - items_from_next_back - 1`. If the full iterator's size is 5, then it has indexes from 0 to 4. If you never polled it from the back, and don't subtract 1, you'll be asking for index 5, which is out of bounds. So you need to ask for one extra from the back. 

Another thing that's different -- you can't poll the iterator for its next_back, because the would move consume the item from the back. There's no peeking so you have to implement it yourself, which is kind of what I did:

The 3 cases:

1. Guard clause
2. If your index from the back is past the iterator's size, incr calls to `next_back` and return padding
3. otherwise, you're in bounds for the iterator, incr calls to `next_back` and return the iterator's `next_back`. Note that this always has to be Some, since the iterator will always produce an item. Basically, the guard clause splits the iterator between Some/None. 

```rust
    let total_consumed = self.elements_from_next + self.elements_from_next_back;

    if total_consumed >= self.elements_required {
        return self.iter.next_back();
    }

    let index_from_back = self.elements_required - self.elements_from_next_back - 1;
    self.elements_from_next_back += 1;

    if index_from_back >= self.iter.len() {
        Some((self.filler)(index_from_back))
    } else {
        self.iter.next_back()
    }
}
```

So those are my final implementations of `next` and `next_back`. That leaves two questions:

1. There's an `#[inline]` for `next`. I deleted this cause you call the filler in next, so I don't think this is a straightforward inlining case.
2. I'm confused about the need to `assert`, either in this version or the previous version of the code (since they both contain the same logic). self.iter.next_back() is always Some, so I don't know what the assert call does. incrementing self.elements_from_next_back doesn't impact the underlying iterator.

src/pad_tail.rs
```diff
@@ -111,30 +95,15 @@ where
...
-        let elements_remaining = self.elements_required - total_consumed;
+        let index_from_back = self.elements_required - self.elements_from_next_back - 1;
         self.elements_from_next_back += 1;
 
-        if self.iter.len() < elements_remaining {
-            Some((self.filler)(
-                self.elements_required - self.elements_from_next_back,
-            ))
+        if index_from_back >= self.iter.len() {
+            Some((self.filler)(index_from_back))
```
phimuemue, line 102: I think this is incorrect. Comparing `self.iter.len` to `index_from_back` only woks if `elements_next` is zero.

Consider this scenario:

```
let mut it = vec![1,2].pad_using(3, |_| 0);
assert_eq!(it.next(), Some(1));
assert_eq!(it.next_back(), Some(0));
assert_eq!(it.next_back(), Some(2)); // <- This triggers.
```

Please re-instate the previous thing (comparing `iter.len` to `elements_remaining`).

If you want to add a test, look for `fn pad_using` in `specializations.rs` and change it to this (please do it in a separate commit):

```
    fn pad_using(v: Vec<u8>) -> () {
        {
            use std::convert::TryFrom;
            let it = v.iter().copied().pad_using(10, |i| u8::try_from(5 * i).unwrap_or(u8::MAX));
            test_specializations(&it);
            test_double_ended_specializations(&it);
        }
        {
            let only_oks = v.into_iter().map(Ok).collect::<Vec<_>>();
            for elements_required in 0..128 {
                let mut oks_then_errs = only_oks.clone();
                while oks_then_errs.len() < elements_required {
                    oks_then_errs.push(Err(oks_then_errs.len()));
                }
                assert_eq!(
                    only_oks.iter().copied().pad_using(elements_required, Err).collect::<Vec<_>>(),
                    oks_then_errs
                );
                // Check next, next_back. Unexpectedly tricky for pad_using
                let mut iter_check = oks_then_errs.into_iter();
                let mut iter_real = only_oks.iter().copied().pad_using(elements_required, Err);
                let mut rng = rand::thread_rng();
                loop {
                    let element = if rng.gen_bool(0.5) {
                        let element_next_check = iter_check.next();
                        let element_next_real = iter_real.next();
                        assert_eq!(element_next_check, element_next_real);
                        element_next_real
                    } else {
                        let element_next_back_check = iter_check.next_back();
                        let element_next_back_real = iter_real.next_back();
                        assert_eq!(element_next_back_check, element_next_back_real);
                        element_next_back_real
                    };
                    if element.is_none() {
                        break;
                    }
                }
            }
        }
    }
```

phimuemue (changes requested):
Hi there, thanks, but I cannot merge as is. `next_back` is buggy.

src/pad_tail.rs
```diff
@@ -111,30 +95,15 @@ where
...
         } else {
             self.iter.next_back()
```
phimuemue, line 104: > 2\. 'm confused about the need to `assert`, either in this version or the previous version of the code (since they both contain the same logic). self.iter.next_back() is always Some, so I don't know what the assert call does. incrementing self.elements_from_next_back doesn't impact the underlying iterator.

Please add `debug_assert` that the element obtained from `next_back` actually `is_some`. We could leave it out with the argument "garbage in, garbage out", but the `debug_asserts` catches the in-garbage instead of tunneling through wrong results.

phimuemue:
@Takashiidobe Thanks for your effort. Very much appreciated.

I took your commits, re-ordered them, and cleaned up a bit to arrive at https://github.com/rust-itertools/itertools/pull/1082.

Thus, closing this PR.

Issue #1084, opened 2026-02-01
Some bounds in circular_tuple_windows are unnecessary (I think)


sgtatham:
The `circular_tuple_windows` method requires the input _iterator_ to be `Clone` and also `ExactSizeIterator`, as well as the item type being `Clone`.

I recently wanted circular tuple windows of size 2, on an iterator which did not have those additional traits. I implemented it myself using the following code (far less generic than yours, but enough for my purpose at the time), which only requires the item type to be `Clone` and the iterator itself to be `Sized`. Extract the first item from the iterator (if any), clone it and put it at the end again (if any), and then take ordinary `tuple_windows` of the result.

```rust
pub trait MyItertools: Iterator {
    fn cyclic_pairs(mut self) -> impl Iterator<Item = (Self::Item, Self::Item)>
    where
        Self::Item: Clone,
        Self: Sized,
    {
        let first = self.next();
        let first_again = first.clone(); // even if it's None
        first
            .into_iter()
            .chain(self)
            .chain(first_again.into_iter())
            .tuple_windows()
    }
}
```

An extended version of the same technique would surely work to implement circular tuple windows of other sizes without those additional constraints on the iterator.

It's not quite as simple as extending this implementation, because of edge cases when the input list is very short. (If I try to make tuple windows of length 4 by extracting up to 3 initial elements and chaining another copy of them at the end, then I get no output windows at all for a length-1 list, whereas the official `circular_tuple_windows` would give a tuple `(a,a,a,a)`, so some special cases would be needed to decide how many copies of the initial elements to chain at the end.)

So, would it be possible to replace the implementation of `circular_tuple_windows` so as to avoid those extra requirements?

If this seems like a good idea in principle, and has no subtle gotchas that you can foresee, I'd be happy to attempt the main implementation. I don't know that my Rust is good enough to get all the outlying parts right, though. (For example, you'd presumably want to propagate `ExactSizeIterator` from the input _if_ the input has it, and no doubt there are other things of that kind I haven't thought of.)

phimuemue:
Hi @sgtatham, I think your assessment is correct.

However, I suggest moving away from tuples[^1] and going with arrays.

I imagine that `circular_array_windows` could remember the "to-be-cycled items" in an `Option<[Self::Item; N-1]>` that is populated on the first `next` call (using `next_array`), and keep an `index` (into the array) to be incremented once the underlying iterator is exhausted.

[...]

PR #1085, opened 2026-02-04
Deprecate tuple_combinations

phimuemue:
As agreed upon in https://github.com/rust-itertools/itertools/issues/1084#issuecomment-3832305846

### DIFF
--- a/benches/specializations.rs
+++ b/benches/specializations.rs
@@ -355,29 +355,29 @@ bench_specializations! {
         }
         v.iter().copied().update(|x| *x *= 7)
     }
-    tuple_combinations1 {
+    array_combinations1 {
         {
             let v = black_box(vec![0; 1024]);
         }
-        v.iter().tuple_combinations::<(_,)>()
+        v.iter().array_combinations::<1>()
     }
-    tuple_combinations2 {
+    array_combinations2 {
         {
             let v = black_box(vec![0; 64]);
         }
-        v.iter().tuple_combinations::<(_, _)>()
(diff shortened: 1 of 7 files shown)


phimuemue:
@jswrenn gentle nudge.

OT: Should we publish 0.15? We had some bug-fixes and other improvements, plus it would make things easier if CI would not fail on each PR.

Even more OT: Can/should we automate publishing somehow? afaik, I'm not allowed to publish to crates.io, but if I can help, let me know.

jswrenn (approved):

jswrenn:
Yeah, I absolutely should publish 0.15. Release is already semi-automated, but the big manual lift that slows me down is hand-crafting the changelog. I'll bang out a release this week and also get you publish perms so I'm not the bottleneck. :)

phimuemue:
> Yeah, I absolutely should publish 0.15. Release is already semi-automated, but the big manual lift that slows me down is hand-crafting the changelog. I'll bang out a release this week and also get you publish perms so I'm not the bottleneck. :)

@jswren: Hope you're doing well and had a good time at RustWeek. Gentle nudge. Can we include https://github.com/rust-itertools/itertools/pull/1105 and get 0.15 out the door? (This would clean the problematic `semver-checks` in our CI.)

ronnodas:
@phimuemue Is there a reason only `tuple_combinations()` was deprecated here and not the other tuple methods?

phimuemue:
We can deprecate the ones that have an array counterpart that's usable in the same situations.

PR #1086, opened 2026-02-05
New method circular_array_windows().

sgtatham:
This is very like the existing `circular_tuple_windows`, but imposes the minimum possible bounds on the input iterator: it must have cloneable items because each item is returned N times, and it must be Sized so that it can be stored in a struct. Unlike `circular_tuple_windows`, it doesn't require the input iterator itself to have extra traits, like Clone or ExactSizeIterator.

Because the return type is an array (as suggested in #1084), we must handle the zero-length case, because you can't have a constraint `N>0`. In that situation we still read to the end of the input iterator, discard each item as we read it, and return a zero-length array per item, preserving the invariant that this iterator is the same length as the input one.

### DIFF
--- a/src/array_impl.rs
+++ b/src/array_impl.rs
@@ -1,0 +1,230 @@
+use crate::Itertools;
+use std::iter::Fuse;
+
+/// An iterator over all contiguous windows of the input iterator,
+/// producing arrays of a specific size.
+///
+/// See [`.array_windows()`](crate::Itertools::array_windows) for more
+/// information.
+#[derive(Debug, Clone)]
+pub struct ArrayWindows<I, const N: usize>
+where
+    I: Iterator + Sized,
+    I::Item: Clone,
+{
+    iter: Fuse<I>,
+    inner: Option<ArrayWindowsInner<I::Item, N>>,
+}
+
+#[derive(Debug, Clone)]
+struct ArrayWindowsInner<T: Clone, const N: usize> {
+    // `window` stores the `N` items delivered in the most
+    // recent output window. It is stored in the form of a ring
+    // buffer, with `window_start` identifying the element
+    // that logically comes first.
+    window: [T; N],
+    window_start: usize,
+}
+
+impl<T: Clone, const N: usize> ArrayWindowsInner<T, N> {
+    /// Replace the least recent item in `window` with a new
+    /// item.
+    fn add_to_buffer(&mut self, item: T) {
+        if N > 0 {
+            self.window[self.window_start] = item;
+            self.window_start = (self.window_start + 1) % N;
+        }
+    }
+
+    /// Construct an array window to return.
+    fn make_window(&self) -> [T; N] {
+        std::array::from_fn(|i| self.window[(i + self.window_start) % N].clone())
+    }
+}
+
+impl<I, const N: usize> Iterator for ArrayWindows<I, N>
+where
+    I: Iterator + Sized,
+    I::Item: Clone,
+{
+    type Item = [I::Item; N];
+
+    fn next(&mut self) -> Option<[I::Item; N]> {
+        match &mut self.inner {
+            // Initialisation code, when next() is called for the first time
+            None => match self.iter.next_array() {
+                None => {
+                    // The input iterator was completely empty
+                    None
+                }
+                Some(buf) => {
+                    let inner = ArrayWindowsInner {
+                        window: buf.clone(),
+                        window_start: 0,
+                    };
+                    let window = inner.make_window();
+                    self.inner = Some(inner);
+                    Some(window)
+                }
+            },
+            Some(inner) => match self.iter.next() {
+                Some(item) => {
+                    inner.add_to_buffer(item);
+                    Some(inner.make_window())
+                }
+                None => None,
+            },
+        }
+    }
+}
+
+pub fn array_windows<I, const N: usize>(iter: I) -> ArrayWindows<I, N>
+where
+    I: Iterator + Sized,
+    I::Item: Clone,
+{
+    ArrayWindows {
+        iter: iter.fuse(),
+        inner: None,
+    }
+}
+
+/// An iterator over all windows, wrapping back to the first elements when the
+/// window would otherwise exceed the length of the iterator, producing arrays
+/// of a specific size.
+///
+/// See [`.circular_array_windows()`](crate::Itertools::circular_array_windows)
+/// for more information.
+#[derive(Debug, Clone)]
+pub struct CircularArrayWindows<I, const N: usize>
+where
+    I: Iterator + Sized,
+    I::Item: Clone,
+{
+    iter: Fuse<I>,
+    inner: Option<CircularArrayWindowsInner<I::Item, N>>,
+}
+
+#[derive(Debug, Clone)]
+struct CircularArrayWindowsInner<T: Clone, const N: usize> {
+    // `prefix` stores the first `N` items output from this iterator.
+    // If the input contained fewer than `N` items, then it is filled
+    // with clones of the previous items in a cycle.
+    //
+    // `prefix_pos` tracks the number of items that have been _used_
+    // from `prefix`. It begins counting up from 0 once the input runs
+    // out. (So in the case where the input iterator is shorter than
+    // `N`, it will begin counting up before `prefix` has even been
+    // populated during setup.)
+    prefix: [T; N],
+    prefix_pos: usize,
+
+    // For delivering the output arrays, we reuse `ArrayWindowsInner`
+    // unchanged.
+    arraywin: ArrayWindowsInner<T, N>,
+}
+
+impl<I, const N: usize> Iterator for CircularArrayWindows<I, N>
+where
+    I: Iterator + Sized,
+    I::Item: Clone,
+{
+    type Item = [I::Item; N];
+
+    fn next(&mut self) -> Option<[I::Item; N]> {
+        match &mut self.inner {
+            // Initialisation code, when next() is called for the first time
+            None => match self.iter.next() {
+                None => {
+                    // The input iterator was completely empty
+                    None
+                }
+                Some(first) => {
+                    // We have at least one item, so we can definitely
+                    // populate `prefix` (even if we have to make N
+                    // copies of this element).
+
+                    // Construct [Option<T>; N] and convert to [T; N]
+                    // once it's full. TODO: can this be improved?
+                    let mut items = std::array::from_fn(|_| None);
+                    let mut prefix_pos = 0;
+                    if N > 0 {
+                        // The first item stored is the one passed to
+                        // us from our caller.
+                        items[0] = Some(first);
+                    }
+                    for i in 1..N {
+                        // Populate the remaining slots in `items`
+                        // from the input iterator.
+                        items[i] = self.iter.next();
+                        if items[i].is_none() {
+                            // If the input iterator runs out early,
+                            // populate the rest of `items` by
+                            // recycling from the beginning, and set
+                            // `prefix_pos` to indicate that we have
+                            // already consumed those items.
+                            for j in i..N {
+                                items[j] = items[j - i].clone();
+                            }
+                            prefix_pos = N - i;
+                            break;
+                        }
+                    }
+                    let items = items.map(Option::unwrap);
+
+                    let inner = CircularArrayWindowsInner {
+                        prefix: items.clone(),
+                        prefix_pos,
+                        arraywin: ArrayWindowsInner {
+                            window: items,
+                            window_start: 0,
+                        },
+                    };
+
+                    let window = inner.arraywin.make_window();
+                    self.inner = Some(inner);
+                    Some(window)
+                }
+            },
+            Some(inner) => {
+                // Normal case. Read the next item in the logical
+                // input sequence (consisting of the contents of the
+                // input iterator followed by N-1 items recycling from
+                // the beginning), and add it to the ring buffer.
+                let item = if let Some(item) = self.iter.next() {
+                    // Read from the input iterator.
+                    item
+                } else if N == 0 {
+                    return None;
+                } else {
+                    assert!(N == 0 || inner.prefix_pos < N);
+                    if inner.prefix_pos + 1 == N {
+                        // The input iterator has run out, and we've
+                        // emitted as many windows as we read items,
+                        // so we've finished.
+                        return None;
+                    }
+                    let item = inner.prefix[inner.prefix_pos].clone();
+                    inner.prefix_pos += 1;
+                    item
+                };
+
+                if N > 0 {
+                    inner.arraywin.add_to_buffer(item);
+                }
+                Some(inner.arraywin.make_window())
+            }
+        }
+    }
+}
+
+pub fn circular_array_windows<I, const N: usize>(iter: I) -> CircularArrayWindows<I, N>
+where
+    I: Iterator + Sized,
+    I::Item: Clone,
+{
+    CircularArrayWindows {
+        iter: iter.fuse(),
+        inner: None,
+    }
+}
--- a/src/array_impl.rs
+++ b/src/array_impl.rs
@@ -59,7 +59,7 @@ where
                 }
                 Some(buf) => {
                     let inner = ArrayWindowsInner {
-                        window: buf.clone(),
+                        window: buf,
                         window_start: 0,
                     };
                     let window = inner.make_window();
@@ -194,11 +194,9 @@ where
                 let item = if let Some(item) = self.iter.next() {
                     // Read from the input iterator.
                     item
-                } else if N == 0 {
-                    return None;
                 } else {
                     assert!(N == 0 || inner.prefix_pos < N);
-                    if inner.prefix_pos + 1 == N {
+                    if N == 0 || inner.prefix_pos + 1 == N {
                         // The input iterator has run out, and we've
                         // emitted as many windows as we read items,
                         // so we've finished.
@@ -209,9 +207,7 @@ where
                     item
                 };
 
-                if N > 0 {
-                    inner.arraywin.add_to_buffer(item);
-                }
+                inner.arraywin.add_to_buffer(item);
                 Some(inner.arraywin.make_window())
             }
         }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -97,6 +97,7 @@ pub mod structs {
         FilterOk, Interleave, InterleaveShortest, MapInto, MapOk, Positions, Product, PutBack,
         TakeWhileRef, TupleCombinations, Update, WhileSome,
     };
+    pub use crate::array_impl::{ArrayWindows, CircularArrayWindows};
     #[cfg(feature = "use_alloc")]
     pub use crate::combinations::{ArrayCombinations, Combinations};
     #[cfg(feature = "use_alloc")]
@@ -174,6 +175,7 @@ pub use crate::unziptuple::{multiunzip, MultiUnzip};
 pub use crate::with_position::Position;
 pub use crate::ziptuple::multizip;
 mod adaptors;
+mod array_impl;
 mod either_or_both;
 pub use crate::either_or_both::EitherOrBoth;
 #[doc(hidden)]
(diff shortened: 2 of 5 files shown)


sgtatham:
I'm not very happy with this code – I have a nasty feeling it's much longer than it needs to be. But it seemed surprisingly tricky to get all the edge cases right, especially when the input iterator runs out early and you have to recycle elements multiple times!

I'd like to remove the dependency on `Vec`, so that this adapter wouldn't have to be gated on `use_alloc`. But I don't know what the policy is on adding extra dependencies. Something like `arrayvec` would replace `Vec` in a sensible way, I think.

(But you would have to give the arrayvecs capacity N, where you only really need N-1, because you can't do arithmetic on integer type parameters!)

src/array_impl.rs
```diff
@@ -0,0 +1,256 @@
...
+                // The first N-1 items come from `ringbuf`
+                self.ringbuf[(i + self.ringpos) % self.ringbuf.len()].clone()
+            } else {
+                // The last item is the new one we just read
+                new_item.clone()
+            }
+        });
+
+        // Replace the oldest item in `ringbuf` with the new one.
+        self.ringbuf[self.ringpos] = new_item;
```
phimuemue: Couldn't this first modify `self.ringbuf`, and only afterwards compute `window`. This could avoid the `if/else` within the `array::from_fn`.

phimuemue:
Hi @sgtatham, thanks for tackling this.

It'll take some time to do a thorough review, but from what I've seen so far, this always generates as many windows as the original iterator had elements.

I think this method could afford to behave similar to [`slice::array_windows`](https://doc.rust-lang.org/std/primitive.slice.html#method.array_windows) resp. [`slice::windows`](https://doc.rust-lang.org/std/primitive.slice.html#method.windows). They return no windows if `N > underlying_iterator.count()` (which is what `tuple_windows` already does).

I *think* we can exploit the `N > underlying_iterator.count()` to make `prefix` a `Option<[T; N]>` [^1]. Then, the first call to `next()` could call `self.iter.next_array()` to populate `prefix`, and eliminate `prefix_needed`. (`prefix==None` would then possibly replace `CircularArrayWindowsState::NotYetStarted` state.) I'd hope this also allows to eliminate `balanced`.

Moreover, could `ringbuf` be a `[T; N]`? Then - if I'm not mistaken - we could provide `circular_array_windows` without `use_alloc`.

Regarding `CircularArrayWindowsState::Done`: I would hope that this can be derived from the condition `cycledpos==prefix.len()-1` (or something similar, please double check).

If it simplifies your implementation, you can even `panic` if `N==0` (that's what `slice::[array_]windows` does).

[^1]: Ideally `[T; N-1]` but afaik Rust does not allow `N-1` (`generic parameters may not be used in const operations`). `[T; N]` is probably close enough, and we can add a `// TODO`.

sgtatham:
> It'll take some time to do a thorough review, but from what I've seen so far, this always generates as many windows as the original iterator had elements.

Yes – that's the same behaviour as `circular_tuple_windows` which I'm replacing. In cases where you ask for a large window from a small list, the existing function repeats the input elements multiple times. For example, length-7 windows from a list (1,2,3) gives you (1,2,3,1,2,3,1), (2,3,1,2,3,1,2), and (3,1,2,3,1,2,3). The rationale seems to be to regard the input list as cyclic, and return a window of it for each possible start position in the cycle.

There certainly is an argument for the idea that perhaps asking for length-7 windows of a length-3 list should say "don't be silly, there's no such thing". I recently posted a [poll on Mastodon](https://hachyderm.io/@simontatham/115984661514454963) about the case of length-2 windows, and almost _everybody_ thought it was weird that if the input list has length 1 then the window (a,a) is returned. But I think it depends on why you want circular windows in the first place, and some people _will_ want those degenerate elements.

You mention a comparison to `slice::[array_]windows`. But that's not cyclic. I agree it's uncontroversial that a _non_-cyclic windows operator returns no windows if there aren't enough input elements to make a whole one.

> Then, the first call to `next()` could call `self.iter.next_array()` to populate `prefix`, and eliminate `prefix_needed`.

If we change the policy in edge cases, so that returned windows will never contain two copies of the same input item, then I believe that strategy would work, because if you can't `.next_array` a whole N items out of the start of the iterator, you were going to fail anyway and return nothing. So yes, in that case both `prefix` and `ringbuf` would either contain N items or none, and could be `Option<[T;N]>`.

But if we're keeping the behaviour like `circular_tuple_windows`, then that's not good enough, because if you can't fetch N things from the iterator (but can fetch at least one), then you have to keep the <N items you _did_ get, and be prepared to cycle them.

That's where a lot of the complexity in this implementation comes from: the way that sometimes you have to start consuming things from `prefix` before you've even finished _building_ `prefix`. In particular, that prevented me from constructing `prefix` as a fixed-size array using `array::from_fn`, because I would have needed the function to receive not just the current array index, but a slice showing me all the items in previous array slots, so that sometimes I could clone one of them.

phimuemue:
Good point. Then let's try to stay in line with the existing `tuple_circular_windows`.[^1]

Maybe [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=942186a8a552bdd73844faf8cb488e67) is a starting point how the first `next`-call could fill the array. I know it's ugly and inefficient, but it contains the complexity within `fn next`. And it might be replaced by something better one day.

[^1]: We could think about making these special cases explicit (e.g. by offering two different functions, or adding a parameter), but I suggest going with what we have.

sgtatham:
Fair's fair, _my_ current implementation is ugly and inefficient too :slightly_smiling_face:

Yes, I see – avoid having to keep an array of `Option` _inside_ the struct, by instead using a temporary one during setup, and moving the items into the plain `[T;N]` once it's complete. I probably should have thought of that. OK, I'll rework on that basis.

sgtatham:
Thanks for the suggestion! It worked pretty well. Here's the revised implementation, using no `Vec` and no longer depending on `use_alloc`. I've also managed to reduce the special cases for `N<2` to only a few lines, instead of that huge entirely separate code path.

You were right that `balance` wasn't needed. It turns out that an equivalent criterion is that `N-1` items have been consumed from `prefix`, _regardless_ of whether they were all pulled out after the input iterator ended, or whether it ended while they were still being built. So tracking the number of items consumed from `prefix` is enough.

(Obvious with hindsight! We produce `k` windows of length `N`, which cover a total of `k+N-1` items, so we finish precisely when we've used up the input iterator and then consumed `N-1` more items.)

src/array_impl.rs
```diff
@@ -0,0 +1,196 @@
...
+    // running, and then counts up from 0 once the input runs out. (So
+    // in the case where the input iterator is shorter than `N`, it
+    // will begin counting up before `prefix` has even been populated
+    // during setup.)
+    //
+    // If the input iterator contains `k` items, then our `k` output
+    // windows of length `N` will cover `k+N-1` items in total. So we
+    // have output enough windows precisely when `prefix_pos` reaches
+    // `N-1`, whether or not we began incrementing `prefix_pos` during
+    // initial setup.
```
phimuemue: Please boil this down to sth like "prefix stores the first N items (to be used when cycled)" or something.

In particular, the last paragraph may mislead, because even if the input iterator has `k` items, the output iterator does not necessarily have `k` items (e.g. if `N>2*k`).

src/array_impl.rs
```diff
@@ -0,0 +1,196 @@
...
+    prefix: [T; N],
+    prefix_pos: Option<usize>,
+
+    // `ringbuf` stores the _most recent_ N items from the input
+    // iterator, which were delivered in the most recent output
+    // window. It is stored in the form of a ring buffer, with
+    // `ringpos` identifying the element that logically comes first.
+    ringbuf: [T; N],
+    ringpos: usize,
```
phimuemue: Please boil this down to sth like `ringbuf` stores the elements for the current window. Maybe even rename `ringbuf` to `current_window_elements` or something, and comment "`current_window_elements` stored in ringbuffer-fashion".

src/array_impl.rs
```diff
@@ -0,0 +1,196 @@
...
+            prefix_pos,
+            ringbuf: items,
+            ringpos: 0,
+        }
+    }
+
+    /// Read the next item in the logical input sequence (consisting
+    /// of the contents of the input iterator followed by N-1 items
+    /// recycling from the beginning). Add it to the ring buffer.
+    fn read_item(&mut self, iter: &mut impl Iterator<Item = T>) -> bool {
```
phimuemue: Please inline `read_item`: It relies on preconditions that are not evident from the function itself (e.g. it increments `pos` and accesses `self.prefix[*pos]`). Inlining it makes all these information local to `fn next`.

src/array_impl.rs
```diff
@@ -0,0 +1,196 @@
...
+    // `ringpos` identifying the element that logically comes first.
+    ringbuf: [T; N],
+    ringpos: usize,
+}
+
+impl<T: Clone, const N: usize> CircularArrayWindowsInner<T, N> {
+    /// Make a new `CircularArrayWindowsInner`, in which `prefix`
+    /// contains the item `first`, plus `N-1` more items from the
+    /// provided iteraor (or recycle existing items if necessary).
+    fn new(first: T, iter: &mut impl Iterator<Item = T>) -> Self {
```
phimuemue: Please inline this function. It manipulates `iter`, and - if we inline everything into `fn next` - , we have the complicated state manipulation localized.

src/array_impl.rs
```diff
@@ -0,0 +1,196 @@
...
+        // To allow building up `prefix` incrementally, we make it in
+        // the form of an array of `Option`. Once we've finished, and
+        // all its elements are `Some`, we can map it through `unwrap`
+        // to make the unconditional `[T; N]` that goes in the output
+        // struct.
```
phimuemue: Please boil this down to sth like: `construct [Option<T>; N] and convert to [T; N] afterwards // TODO can we improve this?`

src/array_impl.rs
```diff
@@ -0,0 +1,196 @@
...
+            // The first item stored is the one passed to us from our
+            // caller.
+            items[0] = Some(first);
+        }
+        for i in 1..N {
+            // Populate the remaining slots in `items` from the input
+            // iterator.
+            let item = iter.next();
+            if item.is_none() {
+                // If the input iterator runs out early, populate the
```
phimuemue: Can't we assign directly to `items[i]` directly instead of introducing the temporary variable `item`?

src/array_impl.rs
```diff
@@ -0,0 +1,196 @@
...
+
+                    let mut inner = CircularArrayWindowsInner::new(first, &mut self.iter);
+                    let window = inner.make_window();
+                    self.inner = Some(inner);
+                    Some(window)
+                }
+            },
+            Some(inner) => {
+                if let Some(pos) = inner.prefix_pos {
+                    if pos + 1 >= N {
```
phimuemue: Can `pos+1>N` even happen in practice? If not, I suggest `assert(N==0 || pos<N)` (after all we'd otherwise index beyound the array's bounds) or sth similar before the check and only checking `pos+1==N`to make the implementation clear.

src/array_impl.rs
```diff
@@ -0,0 +1,196 @@
...
+            item
+        } else if N > 1 {
+            // The input iterator has run out right now, so clone the
+            // first element of `prefix`, and set cyclepos to point to
+            // the next one.
+            self.prefix_pos = Some(1);
+            self.prefix[0].clone()
+        } else {
+            // Special case if N=1: don't read an item at all if the
+            // input iterator has run out.
```
phimuemue: It says `special case if N=1`, but isn't `N==0` included in this branch?

src/array_impl.rs
```diff
@@ -0,0 +1,196 @@
...
+
+// We return exactly one window per input item, so if the input
+// iterator knows its length, then so do we.
+impl<I, const N: usize> ExactSizeIterator for CircularArrayWindows<I, N>
+where
+    I: Iterator + Sized + ExactSizeIterator,
+    I::Item: Clone,
+{
+    fn len(&self) -> usize {
+        self.iter.len()
```
phimuemue: I think this is wrong for `N>2*self.iter.len()` (the case where we must reuse elements from `prefix` to even create the first window).

Moreover, implementing ExactSizeIterator requires that ["the implementation of `Iterator::size_hint` must return the exact size of the iterator"](https://doc.rust-lang.org/std/iter/trait.ExactSizeIterator.html).

Maybe let's leave out `ExactSizeIterator` for now.

src/array_impl.rs
```diff
@@ -0,0 +1,196 @@
...
+    type Item = [I::Item; N];
+
+    fn next(&mut self) -> Option<[I::Item; N]> {
+        match &mut self.inner {
+            // Initialisation code, when next() is called for the first time
+            None => match self.iter.next() {
+                None => {
+                    // The input iterator was completely empty
+                    None
+                }
```
phimuemue, line 141: Behavior for unfused iterators (iterators that yield `Some(...)` after they already yielded `None`) is questionable here - it would create `inner` on a non-first-call to `next`.

Let's `fuse` the underlying iterator. `windows` for unfused iterators are confusing anyway.

src/lib.rs
```diff
@@ -900,6 +902,38 @@ pub trait Itertools: Iterator {
...
+    /// }
+    /// assert_eq!(v, vec![[1, 2], [2, 3], [3, 4], [4, 1]]);
+    ///
+    /// let mut it = (1..5).circular_array_windows();
+    /// assert_eq!(Some([1, 2, 3]), it.next());
+    /// assert_eq!(Some([2, 3, 4]), it.next());
+    /// assert_eq!(Some([3, 4, 1]), it.next());
+    /// assert_eq!(Some([4, 1, 2]), it.next());
+    /// assert_eq!(None, it.next());
+    /// ```
```
phimuemue, line 1016: Please make the `assert`s uniform: Either go with `assert_eq(..., vec![...])` or with the single `assert_eq`. (I prefer the first one, maybe even condensed to sth like `assert_eq!((1..5).windows().collect(), vec![...])`.)

tests/quick.rs
```diff
@@ -1300,6 +1300,48 @@ quickcheck! {
     }
 }
 
+// array iterators
+quickcheck! {
```
phimuemue, line 1304: We should add tests that verify that `tuple_circular_windows` does the same as `array_circular_windows`.

src/lib.rs
phimuemue, line 1016: Should this describe what happens for e.g. `[1,2].circular_windows<5>()`?

phimuemue (commented):
Hi there, thanks for this! Already much better now that `balanced` et al are out. Also nice to see the `Vec`s gone.

However, I'm wondering if we can simplify a bit further. In particular the case distinction in `read_item` worries me.

As I wrote, I think we should `fuse` the underlying iterator, and I think `fn next` could then do sth like this:

```
fn next() {
 match self.iter.next() {
  None => match self.inner {
   Some(inner) => {// start re-using stuff from inner.prefix}
   None => None
  },
  Some(item) => match self.inner {
   Some(inner) => {update inner},
   None => {initialize inner}
  }
 }
}
```

The story would then be "try `self.iter` as long as possible, and only once it returns `None`, go with `self.prefix`... 

**Which begs one question:** Can we make all of this even simpler by doing a stupid trick: Should we first implement `array_windows`, and then redirect `circular_array_windows` so that the first call to `fn next` sets up `prefix` and chains it after `iter`?

I.e. given `tuple_windows`, could we say the following?

```
enum CircularTupleWindows<N, Iter> {
 NotYetStarted(<Fused<Iter>>),
 Started(
  TupleWindows<N, Chain<Iter, [Iter::Item; N-1]>>
 ),
}

impl Iterator for CircularTupleWindows {
 fn next() {
  match self {
   NotYetStarted(iter) => self=Started(
    {
     let prefix = compute prefix from iter;
     iter.chain(prefix).tuple_windows()
    }
   ),
   Started(tuple_windows_iter) => {}
  }
  let Started(internal) = self else {panic!()}; // or similar
  internal.next()
 }
}
```

I'm sorry this came to my mind just now, but I think it's worth exploring.

As always: Use your judgement, and please excuse if I made obvious mistakes. (It's always too late in the evening when I do reviews...)

src/array_impl.rs
sgtatham: I think it's right for all combinations of N and the input length, because of the characterisation of circular windows I mentioned last time: regard the input iterator as cyclic, and return a window for every possible start point in the cycle.

For example, for input length 1 we always get one window, `[a,a,a,a,...,a]`. For input length 2 we get `[a,b,a,b,...]` and `[b,a,b,a,...]`, exactly 2 windows. Both of these are regardless of N.

src/lib.rs
sgtatham: Ah, I should have known that copying the demo code directly from `circular_tuple_windows` wouldn't be good enough :slightly_smiling_face:

sgtatham:
> **Which begs one question:** Can we make all of this even simpler by doing a stupid trick: Should we first implement `array_windows`, and then redirect `circular_array_windows` so that the first call to `fn next` sets up `prefix` and chains it after `iter`?

I had a try at this, but it didn't go well. After setting up `prefix` you have to chain the first copy as well as the second copy, and the edge cases are at least as painful as doing it directly.

But I did get what looks like a working implementation of `array_windows` itself out of it, and that's obviously worth having, because it would be silly to provide _just_ the circular case and not the easier one!

I'll try again tomorrow with a different approach.

src/lib.rs
phimuemue: > Ah, I should have known that copying the demo code directly from `circular_tuple_windows` wouldn't be good enough 🙂

Haha - (I'm) Guilty as charged. Sorry for being pedantic, but I think if we touch the code, we should make it (and its documentation) as good as we can.

phimuemue:
Hi @sgtatham, I think you are right about my `chain(prefix, iter, prefix)`-idea.

However, I think we still have a bit too much state/case distinctions in there. See [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=3e6feb3b0c6a9d08e5e42227df8292f7) where I adapted your solution. I did not test it or think it too much, so be prepared to find bugs - but it may serve as a start.

And, again, please excuse me being nitpicky, but we've had too much state in our iterators, and it always took a while to untangle them. I want to avoid that in the future.

src/array_impl.rs
phimuemue: I may be wrong, but imagine this:

```
let windows = [1,2].circular_windows<5>();
assert_eq!(windows.len(), 2); // <- that's what I got from your explanation
windows.next(); // yields [1,2,1,2,1], and exhausts self.iter
assert_eq!(windows.len(), 1); // <- this should be true, but the proposed implementation returns 0, because self.iter has been exhausted by the call to next
```

Now that I write it, I think it's not only problematic for `N>2*self.iter.len` (that's just the first counterexample that came to my mind): I think `fn len` would have to check the whole `inner` structure to determine how many items are left.

sgtatham: Ah! Yes, now I understand what I missed. I was thinking only of knowing how many items you're going to return _up front_ – when the `CircularArrayWindows` has just been created, and nobody has called `next` on it yet. But in fact `ExactSizeIterator` requires you to count down as you yield items, so yes, my implementation is wrong.

I could probably still make it work by adding a counter. (That `balanced` thing I took out might have been adaptable to do it!) But in that case I agree it's most sensible to just leave it out.

sgtatham:
> However, I think we still have a bit too much state/case distinctions in there. See [here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=3e6feb3b0c6a9d08e5e42227df8292f7) where I adapted your solution.

Ah, yes. I think that only works _because_ you also fused the input iterator. Half the complexity in my version was trying very hard to avoid calling `.next()` on the input iterator unless I was absolutely sure I had to.

(And, as you pointed out in another comment, I still didn't quite manage it!)

I've pushed a heavily revised version, which I _think_ addresses all your comments above (though Murphy's Law says I probably missed one) and incorporates the control flow change from that gist.

It also includes plain `array_windows`. As it turned out, even though I didn't end up implementing the circular one by literally making an `ArrayWindows`, I was able to share the _inner_ structures' window-construction code, by having `CircularArrayWindowsInner` contain a plain `ArrayWindowsInner` that does half its work for it.

src/array_impl.rs
```diff
@@ -0,0 +1,230 @@
...
+        match &mut self.inner {
+            // Initialisation code, when next() is called for the first time
+            None => match self.iter.next_array() {
+                None => {
+                    // The input iterator was completely empty
+                    None
+                }
+                Some(buf) => {
+                    let inner = ArrayWindowsInner {
+                        window: buf.clone(),
```
phimuemue: Do we really need to `clone` here? I don't see a second use of `buf`, so I think we can move it.

src/array_impl.rs
```diff
@@ -0,0 +1,230 @@
...
+                        // emitted as many windows as we read items,
+                        // so we've finished.
+                        return None;
+                    }
+                    let item = inner.prefix[inner.prefix_pos].clone();
+                    inner.prefix_pos += 1;
+                    item
+                };
+
+                if N > 0 {
```
phimuemue: `if N > 0` is not needed here because `add_to_buffer` does it itself. Or am I missing something?

src/array_impl.rs
```diff
@@ -0,0 +1,230 @@
...
+                // input iterator followed by N-1 items recycling from
+                // the beginning), and add it to the ring buffer.
+                let item = if let Some(item) = self.iter.next() {
+                    // Read from the input iterator.
+                    item
+                } else if N == 0 {
+                    return None;
+                } else {
+                    assert!(N == 0 || inner.prefix_pos < N);
+                    if inner.prefix_pos + 1 == N {
```
phimuemue: Can you merge the `else if N==0` case and the `if inner.prefix_pos + 1 == N` case? This would only leave us with one `return None`.

The result could be sth like this:

```
let item = if let Some(item) = self.iter.next() {
 item
} else {
 assert!(N==0 || inner.prefix_pos < N);
 if (N==0 || inner.prefix_pos + 1 == N) {
  return None; // insufficient elements (N==0 or prefix already processed)
 }
 let item = inner.prefix[inner.prefix_pos].clone();
 inner.prefix_pos += 1;
 item
}
```

src/lib.rs
```diff
@@ -900,6 +902,121 @@ pub trait Itertools: Iterator {
...
+    ///     println!("{a} {b} {c}");
+    /// }
+    ///
+    /// // You can also specify the complete type.
+    /// use itertools::ArrayWindows;
+    /// use std::ops::Range;
+    ///
+    /// let it: ArrayWindows<Range<u32>, 3> = (1..6).array_windows();
+    /// itertools::assert_equal(it, vec![[1, 2, 3], [2, 3, 4], [3, 4, 5]]);
+    /// ```
```
phimuemue: We should not encourage users to explicitly specify the iterator's types. Iterator types quickly tend to become complex.

tests/laziness.rs
```diff
@@ -104,6 +104,12 @@ must_use_tests! {
         let _ = Panicking.circular_tuple_windows::<(_, _)>();
         let _ = Panicking.circular_tuple_windows::<(_, _, _)>();
     }
+    circular_array_windows {
+        let _ = Panicking.circular_array_windows::<0>();
+        let _ = Panicking.circular_array_windows::<1>();
+        let _ = Panicking.circular_array_windows::<2>();
+        let _ = Panicking.circular_array_windows::<3>();
+    }
```
phimuemue, line 118: That's good. Please add `array_windows`, too.

phimuemue (commented):
Hi there, thanks for this. I *think* this implementation is correct.

Can you add the following tests in a new commit (so that I don't have to re-review)?

* Test that the new functions behave similar to their tuple-counterparts
  ```
  fn array_windows_equal_tuple_windows() {
    fn internal(slice: &[usize]) {
        assert_eq!(
            slice.iter().tuple_windows::<(_,)>()
                .map(Into::<[_; 1]>::into),
            slice.iter().array_windows::<1>()
        );
        assert_eq!(
            slice.iter().tuple_windows::<(_,_,)>()
                .map(Into::<[_; 2]>::into),
            slice.iter().array_windows::<2>()
        );
        // ... please extend up to length 10 or so
    }
    for i in 0..100 {
        internal(&(0..i).collect::<Vec<_>>());
    }
  }
  // Please add array_cirular_windows_equal_tuple_circular_windows accordingly
  ```
* Test/document the behavior for `N==0` explicitly.

src/lib.rs
sgtatham: This is another case where I did it the same as the existing code but apparently ought to have realised that that wasn't the right thing to do :slightly_smiling_face: 

I assumed that the purpose of existing examples showing someone how to specify the full type was not to encourage it, but so that they'd know how to do it when they _had_ to, e.g. because they need to store an `ArrayWindows<something>` as a field of a struct and there's no alternative but to explicitly state the type.

But I suppose it's not as if they _can't_ work it out in that situation – the type signature in the rustdoc output tells them what they need to know.

sgtatham:
> Can you add the following tests in a new commit (so that I don't have to re-review)?

Done (or rather, three commits).
 
>     * Test that the new functions behave similar to their tuple-counterparts

I had some tests of this already, but I've replaced them with your version if that's what you prefer.

>     * Test/document the behavior for `N==0` explicitly.

Added documentation and doc tests. But there were behind-the-scenes tests of zero-length windows already, in `tests/arrays.rs`. Both sets are marked with a comment (one of which I only just added, but the other had one already).

phimuemue:
Nice. Thank you.

phimuemue (approved):

PR #1087, opened 2026-02-23
Add flat_map_ok iterator adaptor

### LINKED ISSUE #528: flat_map_ok
Similar to #522, which defines `flatten_ok` there should directly be a `flat_map_ok` which operates on an `IntoIter<Item = Result<T, E>>`, applies a function to `T` if the value is `Ok` and flattens the result of that function. It should be equivalent to `iterator.map_ok(function).flatten_ok()`.

veeceey:
Adds a `flat_map_ok` adaptor that applies a function to `Ok` values and flattens the result, passing `Err` values through unchanged. Basically a combined `.map_ok(f).flatten_ok()` in a single step.

The implementation follows the same pattern as the existing `flatten_ok` adaptor, with support for:

- `DoubleEndedIterator` (when the inner iterator supports it)
- `FusedIterator`
- `Clone` / `Debug`
- `fold` / `rfold` optimizations

Example usage:

```rust
let input = vec![Ok(2i32), Err(false), Ok(3i32)];
let it = input.into_iter().flat_map_ok(|i| 0..i);
// yields: Ok(0), Ok(1), Err(false), Ok(0), Ok(1), Ok(2)
```

Closes #528

### DIFF
--- a/src/flat_map_ok.rs
+++ b/src/flat_map_ok.rs
@@ -1,0 +1,206 @@
+use crate::size_hint;
+use std::{
+    fmt,
+    iter::{DoubleEndedIterator, FusedIterator},
+};
+
+pub fn flat_map_ok<I, F, T, U, E>(iter: I, f: F) -> FlatMapOk<I, F, T, U, E>
+where
+    I: Iterator<Item = Result<T, E>>,
+    F: FnMut(T) -> U,
+    U: IntoIterator,
+{
+    FlatMapOk {
+        iter,
+        f,
+        inner_front: None,
+        inner_back: None,
+        _phantom: std::marker::PhantomData,
+    }
+}
+
+/// An iterator adaptor that applies a function to `Result::Ok` values and
+/// flattens the resulting iterator. `Result::Err` values are passed through
+/// unchanged.
+///
+/// This is equivalent to `.map_ok(f).flatten_ok()`.
+///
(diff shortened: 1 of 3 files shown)


veeceey:
Test results:

```
running 8 tests
test flat_map_ok_all_errors ... ok
test flat_map_ok_collect_ok_reverse ... ok
test flat_map_ok_collect_ok_forward ... ok
test flat_map_ok_collect_mixed ... ok
test flat_map_ok_empty_results ... ok
test flat_map_ok_equivalence_with_map_ok_flatten_ok ... ok
test flat_map_ok_mixed_forward ... ok
test flat_map_ok_mixed_reverse ... ok

test result: ok. 8 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

Full test suite also passes (202 tests).

phimuemue:
Hi there, thanks for this.

I'm reluctant to include even more convenience functions.

Can you elaborate why this new iterator is strictly better than `.map_ok(f).flatten_ok()`? (Apart from being a bit simpler to write.)

veeceey:
That's a fair question — `.map_ok(f).flatten_ok()` does achieve the same result, and this is mainly syntactic sugar. I don't have a strong argument for why a dedicated adapter is strictly necessary here. Closing this — thanks for the consideration.

PR #1094, opened 2026-04-03
Made explicit that duplicates are preserved. 

miselico:
Also changed the example such that there are duplicates. Added a comments to indicate what the content of the three input iterators is.


phimuemue:
Good one, thank you. 

Brings up the question which of the "equal" elements is preferred.

phimuemue (approved):

PR #1097, opened 2026-04-15
Remove authors from Cargo metadata (per RFC 3052)

xtqqczze:
The `authors` field has been deprecated since https://github.com/rust-lang/cargo/pull/15068.


jswrenn:
RFC3052 makes the `authors` field optional, not deprecated. I think it's a nice way to give kudos.

xtqqczze:
According to the [Cargo docs](https://doc.rust-lang.org/cargo/reference/manifest.html#the-authors-field), the `authors` field is deprecated. It has no clear meaning (original authors vs maintainers vs contributors) and doesn’t scale for a project of this size.

PR #1098, opened 2026-04-15
clippy: fix manual_option_zip lint

xtqqczze:
fix warn-by-default lint in Rust 1.95

https://rust-lang.github.io/rust-clippy/master/index.html#manual_option_zip

### DIFF
--- a/tests/test_core.rs
+++ b/tests/test_core.rs
@@ -226,7 +226,7 @@ fn batching() {
     let pit = xs
         .iter()
         .cloned()
-        .batching(|it| it.next().and_then(|x| it.next().map(|y| (x, y))));
+        .batching(|it| it.next_array::<2>().map(|[a, b]| (a, b)));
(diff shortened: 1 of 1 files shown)


xtqqczze:
@phimuemue PTAL

phimuemue:
Hi there, I think your transformation is correct, so thanks for this.

However, it relies on evaluation order, i.e. this:

```
it.next() // must be evaluated first
    .zip(
        it.next() // must be evaluated second
    )
```

From what I glean from https://doc.rust-lang.org/reference/expressions.html#evaluation-order-of-operands, one *can* rely on it, but I don't want to encourage to do it.

Can you maybe change it to `it.next_array::<2>()`?

xtqqczze:
> Can you maybe change it to `it.next_array::<2>()`?

Yeah, that looks better: db613dc3afc7929de3ec963d661fc574141276db

phimuemue (approved):

PR #1099, opened 2026-04-15
clippy: fix collapsible_match lint

xtqqczze:
fix warn-by-default lint in Rust 1.95

https://rust-lang.github.io/rust-clippy/master/index.html#collapsible_match

### DIFF
--- a/src/groupbylazy.rs
+++ b/src/groupbylazy.rs
@@ -197,13 +197,12 @@ where
 
         while let Some(elt) = self.next_element() {
             let key = self.key.call_mut(&elt);
-            match self.current_key.take() {
-                Some(old_key) if old_key != key => {
+            if let Some(old_key) = self.current_key.take() {
+                if old_key != key {
                     self.current_key = Some(key);
                     first_elt = Some(elt);
                     break;
                 }
-                _ => {}
             }
             self.current_key = Some(key);
             if self.top_group != self.dropped_group {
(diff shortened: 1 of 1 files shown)


xtqqczze:
I fixed this manually because the Clippy suggestion introduced an error; issue reported: https://github.com/rust-lang/rust-clippy/issues/16860.

phimuemue:
Hi there, thanks for this, but oh Clippy...

@jswrenn I think we should avoid `_`-wildcard-patterns in `match` unless *strictly* necessary. They don't show the possible states explicitly. I'd even discourage them with `Option`/`Result`/other similarly well-known types (if only for consistency). Same applies to `matches!`/`if let` in my book.

My guideline for reasonably explicit code would be:
- No wildcards in `match` unless absolutely necessary. (Not even for simple types for consistency.)
- No `matches!`.
- Allow `if let` for `Option`.

imho we should bring our code to these standards, and - if Clippy complains too loudly - silence this particular lint.

xtqqczze:
In these cases, we could replace the match with a single if let guard, e.g.:

```rust
            if let Some(old_key) = self.current_key.take() {
                if old_key != key {
                    self.current_key = Some(key);
                    first_elt = Some(elt);
                    break;
                }
            }
```

phimuemue:
> In these cases, we could replace the match with a single if let guard, e.g.:

Yes, please go with that.

phimuemue (approved):
Thank you!

PR #1100, opened 2026-05-01
Fix `IntoChunks` docstring

romanz:
IIUC, `ChunkLazy` is no longer used.


phimuemue:
Looks reasonable, thank you.

Did you grep for other occurrences?

phimuemue (approved):

PR #1101, opened 2026-05-17
take_while_inclusive: tighten FusedIterator to require I: FusedIterator

### LINKED ISSUE #1088: Incorrect FusedIterator implementation on TakeWhileInclusive
As it is currently implemented, `TakeWhileInclusive` is not a fused iterator if the underlying iterator is not also fused. A minimal example is at the bottom of this issue.

Here are some options as I see it:
- Remove the `FusedIterator` implementation
- Edit `TakeWhileInclusive` to make it fused, and document that this is a fused iterator
- Edit the `FusedIterator` implementation to only apply when the underlying iterator is fused

Below is an example:

```rust
use itertools::Itertools;

fn main() {
    let my_iterator = MyIterator { counter: 0 };
    let mut take_while_inclusive = my_iterator.take_while_inclusive(|x| *x < 10);

    // This is okay...
    assert_eq!(take_while_inclusive.next(), None);
    assert_eq!(take_while_inclusive.next(), Some(2));
    assert_eq!(take_while_inclusive.next(), None);

    let my_iterator = MyIterator { counter: 0 };
    let mut fused = my_iterator.take_while_inclusive(|x| *x < 10).fuse();

    // ...but this is bad, since a fused iterator is returning Some after None
    assert_eq!(fused.next(), None);
    assert_eq!(fused.next(), Some(2));
    assert_eq!(fused.next(), None);
}

/// An iterator returning `None` for all odd numbers
struct MyIterator {
    counter: usize,
}

impl Iterator for MyIterator {
    type Item = usize;

        self.counter += 1;

        if self.counter.is_multiple_of(2) {
            Some(self.counter)
        } else {
            None
        }
    }
}
```

c-tonneslan:
Closes #1088.

`TakeWhileInclusive` currently unconditionally implements `FusedIterator`, but its `next()` keeps calling the inner iterator after that inner iterator returned `None` (it only sets `done` when the predicate fails, not when the source runs dry). So if you wrap a non-fused iterator and then `.fuse()` the result, `Fuse` trusts the marker, short-circuits to `None` on its first observed `None`, and the underlying iterator's subsequent `Some` is dropped. The other direction is just as broken: without `.fuse()`, the marker is a lie since `Some` can follow `None`.

`std::iter::TakeWhile` already does the right thing here, only implementing `FusedIterator` when the underlying iterator is fused. Doing the same for `TakeWhileInclusive` keeps the marker honest without changing the runtime behavior for any caller whose source iterator was already fused.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1586,7 +1586,7 @@ pub trait Itertools: Iterator {
     /// [.duplicates()](crate::Itertools::duplicates), but uses the specified hash builder to hash
     /// the elements for comparison.
     ///
-    /// Warning: `hash_builder` is normally randomly generated, and is designed to allow it's
+    /// Warning: `hash_builder` is normally randomly generated, and is designed to allow its
     /// users to be resistant to attacks that cause many collisions and very poor performance.
     /// Setting it manually using this function can expose a DoS attack vector.
     ///


phimuemue:
Closed as per #1088

PR #1102, opened 2026-05-17
InterleaveShortest: don't overflow size_hint lower bound

### LINKED ISSUE #1068: `interleave_shortest`'s size_hint panics with large iterators.
[Playground here](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=bf3fa6b28f0693e46a4ef77710476a65).

You can create an iterator with `interleave_shortest` that is longer than `usize::MAX`. However, since an iterator is lazy, you can use it (I assume you could take `usize::MAX` items from it at least). When you try to call `size_hint` on a long iterator created this way however, it panics, even though you can use the iterator itself.

I assume this is a bug in the `size_hint` method since I don't expect `size_hint` itself to panic -- it should just give me info. In this case, I would prefer `(0, None)` instead of panicking.

c-tonneslan:
`combined_lower` is computed with `saturating_mul`, so it can already sit at `usize::MAX`. The next line then does an unchecked `combined_lower + 1` when the current half has a larger lower bound, which panics in debug builds. Hit it with something like `iter::repeat(0).take(usize::MAX).interleave_shortest(iter::repeat(1))`.

Swap the `+` for `saturating_add`.

Closes #1068.

### DIFF
--- a/src/adaptors/mod.rs
+++ b/src/adaptors/mod.rs
@@ -179,7 +179,7 @@ where
         let (combined_lower, combined_upper) =
             size_hint::mul_scalar(size_hint::min(curr_hint, next_hint), 2);
         let lower = if curr_lower > next_lower {
-            combined_lower + 1
+            combined_lower.saturating_add(1)
         } else {
             combined_lower
         };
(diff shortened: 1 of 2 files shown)


phimuemue:
Closing this as per #1112.

PR #1104, opened 2026-05-18
feat(Itertools): add strip_prefix and strip_prefix_by methods

### LINKED ISSUE #1096: `Itertools::strip_prefix`
```rust
trait Itertools {
    fn strip_prefix(mut self, other: impl IntoIterator<Item = Self::Item>) -> Option<Self>
        where Self::Item: Eq
    {
        let mut other_iter = other.into_iter();

        while let Some(other_item) = other_iter.next() {
            let self_item = self.next()?;
            if self_item != other_item {
                return None;
            }
        }

        Some(self)
    }
}
```

I'm proposing a new `Itertools` method that does what `str::strip_prefix` does, but for iterators.

It takes a second iterator and skips all items in `self` that exactly match the items produced by the second iterator. If the two iterators do not have the same items, the method returns `None` instead.

My use case where it felt natural to reach for this: I'm parsing `object_store` paths that I expect to point to files within a specific directory.

SAY-5:
Adds `strip_prefix` and `strip_prefix_by` to the `Itertools` trait, the iterator analogue of `str::strip_prefix`, as discussed in #1096. `strip_prefix` is implemented in terms of `strip_prefix_by`, which takes an equality predicate so the prefix items may differ in type from `Self::Item`. Includes doctests and quickcheck tests that assert parity with `str::strip_prefix`.

Closes #1096

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -5122,6 +5122,97 @@ pub trait Itertools: Iterator {
             _ => Err(sh),
         }
     }
+
+    /// Removes a prefix from the iterator, returning the rest.
+    ///
+    /// If `self` begins with all the items yielded by `prefix` (in order), this
+    /// returns `Ok` of the iterator advanced past that prefix. Otherwise it
+    /// returns `Err(StripPrefixError { .. })` exposing the partially-consumed
+    /// iterator, the remaining prefix, and the items that failed to match, so
+    /// callers can recover progress made before the mismatch.
+    ///
+    /// See [`strip_prefix_by`](Itertools::strip_prefix_by) for a variant
+    /// taking an explicit equality predicate.
+    ///
+    /// ```
+    /// use itertools::Itertools;
+    ///
+    /// let ok = (1..6).strip_prefix([1, 2]).map(Itertools::collect_vec).ok();
+    /// assert_eq!(ok, Some(vec![3, 4, 5]));
+    /// assert!((1..6).strip_prefix([1, 9]).is_err());
+    /// let empty = (1..6).strip_prefix(std::iter::empty::<i32>()).map(Itertools::collect_vec).ok();
+    /// assert_eq!(empty, Some(vec![1, 2, 3, 4, 5]));
+    /// ```
+    fn strip_prefix<Prefix>(
+        self,
+        prefix: Prefix,
+    ) -> Result<Self, StripPrefixError<Self, Prefix::IntoIter, Self::Item>>
+    where
+        Self: Sized,
+        Prefix: IntoIterator,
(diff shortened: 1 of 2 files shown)


phimuemue:
Hi there, thanks for this.

Regarding the API: Currently this only returns `Some(self_after_prefix_removal)` on success, but the failure case may be equally interesting for callers.

Can we - instead of `Option<Self>` - return a `Result<Self, (Self, Self::Item, J::Item, J)>`, where the `Ok` case resembles the prefix-removed iterator, and the `Err` case holds all cheaply-obtainable components that may be of interest for the caller?
- prefix-partially-removed iterator (`Self`)
- mismatched items (`Self::Item` vs `J::Item`)
- remaining prefix (`J`)

Returning a bespoke `struct` for the error case may be viable, too.

src/lib.rs
```diff
@@ -5122,6 +5122,97 @@ pub trait Itertools: Iterator {
...
+    }
+}
+
+/// The error returned by [`Itertools::strip_prefix`] and
+/// [`Itertools::strip_prefix_by`] when the iterator does not start with the
+/// requested prefix.
+///
+/// All fields are public so callers can recover the partially-consumed
+/// iterators and the mismatched items.
+#[derive(Debug, Clone, PartialEq, Eq)]
```
phimuemue: Let's avoid `PartialEq, Eq` if it's unclear if it is really needed. (Comparing `iterator` is oftentimes not doable via `PartialEq` oftentimes.)

src/lib.rs
```diff
@@ -5122,6 +5122,97 @@ pub trait Itertools: Iterator {
...
+                got => {
+                    return Err(StripPrefixError {
+                        iterator: self,
+                        prefix,
+                        mismatch: (got, wanted),
+                    });
+                }
+            }
+        }
+        Ok(self)
```
phimuemue: We canonically try to use avoid the `while let Some(wanted) = prefix.next` pattern. Can you change it to this?

```
        let mut prefix = prefix.into_iter();
        match prefix.by_ref().try_for_each(|wanted| {
            match self.next() {
                Some(got) if eq(&got, &wanted) => Ok(()),
                got => Err((got, wanted)),
            }
        }) {
            Ok(()) => Ok(self),
            Err(mismatch) => Err(StripPrefixError {
                iterator: self,
                prefix,
                mismatch,
            }),
        }
```

phimuemue (changes requested):
Thank you for this. Could you adopt my changes, then I think we're good to go.

phimuemue (approved):
Thank you.

Issue #1106, opened 2026-05-25
Deprecate `fold_while()` in favor of `Iterator::try_fold()`


ronnodas:
This was previously discussed in #469, but it seems that was mainly about documenting an idiomatic way to rewrite an example using `try_fold()`. Maybe there was already a deprecation that has since been reverted? It's also mentioned in #223 as something that *will be* deprecated and removed.

The [`ControlFlow`](https://doc.rust-lang.org/std/ops/enum.ControlFlow.html) enum, which seems to have the most apt semantics, has been stable since Rust 1.55. `FoldWhile` does have an [`into_inner()`](https://docs.rs/itertools/0.14.0/itertools/enum.FoldWhile.html#method.into_inner) method that doesn't have a direct analog on `ControlFlow` [yet](https://github.com/rust-lang/rust/issues/137461), but this seems a minor ergonomics deficiency.

scottmcm:
Yes, if you check https://github.com/rust-lang/rust/issues/75744#issue-682773330 you'll see that it links https://github.com/rust-itertools/itertools/issues/469 -- `ControlFlow` was added *directly* to be used for this :)

It was https://github.com/rust-itertools/itertools/pull/476 that undeprecated it.

I'm largely peanut gallery here, but it makes sense at least to me that it would be reasonable to re-deprecate now.  With the MSRV at 1.63 should even be able to use `ControlFlow` directly in the implementation.

https://github.com/rust-itertools/itertools/blob/d5897f7bb2283a34b736f539a369fe16dc9f6bb4/Cargo.toml#L19

PR #1107, opened 2026-05-25
Use `ControlFlow` in `fold_while` implementation

scottmcm:
Now that the MSRV is high enough, might as well.

(inspired by https://github.com/rust-itertools/itertools/issues/1106#issuecomment-4536154233)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3169,7 +3169,7 @@ pub trait Itertools: Iterator {
         Self: Sized,
         F: FnMut(B, Self::Item) -> FoldWhile<B>,
     {
-        use Result::{Err as Break, Ok as Continue};
+        use core::ops::ControlFlow::{Break, Continue};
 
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -3169,7 +3169,7 @@ pub trait Itertools: Iterator {
         Self: Sized,
         F: FnMut(B, Self::Item) -> FoldWhile<B>,
     {
-        use Result::{Err as Break, Ok as Continue};
+        use core::ops::ControlFlow::{Break, Continue};
```
scottmcm, line 3172: No idea how this could change the code coverage...

phimuemue (approved):

phimuemue:
@scottmcm thanks.

@jswrenn I hope I don't miss the forest for the trees, but imho we should deprecate (and later remove) `Itertools::{fold_while, FoldWhile}`.

PR #1109, opened 2026-06-19
Use saturating arithmetic to avoid usize overflow panics (#1066, #1067, #1068)

### LINKED ISSUE #1066: Bug in `k_smallest_relaxed`?
If I write this code using `k_smallest`, and ask for `usize::MAX` items, this returns all the items in the iterator, which makes sense.

```rust
use itertools::Itertools;

#[test]
fn k_smallest_memory_is_fine() {
    let iter = vec![1].into_iter();
    let first = iter.k_smallest(usize::MAX);

    itertools::assert_equal(first, vec![1]);
}
```

[I think this is ok because this line only requires memory for an iterator up to size `k`](https://github.com/rust-itertools/itertools/blob/6bd5053fca990a8c7a3de6429faa53fef81de41d/src/k_smallest.rs#L52)

However, when swapping to `k_smallest_relaxed`  given `usize::MAX / 2 + 1` items, it panics:

```rust
use itertools::Itertools;

#[test]
fn k_smallest_memory_does_not_panic() {
    let iter = vec![1].into_iter();
    let first = iter.k_smallest_relaxed(usize::MAX / 2);

    itertools::assert_equal(first, vec![1]);
}

#[test]
#[should_panic]
fn k_smallest_memory_does_panic() {
    let iter = vec![1].into_iter();
    let first = iter.k_smallest_relaxed(usize::MAX / 2 + 1);

    itertools::assert_equal(first, vec![1]);
}
```

[It seems like the problematic line is here](https://github.com/rust-itertools/itertools/blob/6bd5053fca990a8c7a3de6429faa53fef81de41d/src/k_smallest.rs#L102) where `k` is multiplied by 2, unchecked, so usize::MAX / 2 + 1 overflows it and causes the panic.

At the very least, this simple fix causes the test above to pass without panicking, since it does a saturating multiply.

```diff
diff --git a/src/k_smallest.rs b/src/k_smallest.rs
index 7e4ace2..c28831b 100644
--- a/src/k_smallest.rs
+++ b/src/k_smallest.rs
@@ -99,7 +99,10 @@ where
     }
 
     let mut iter = iter.fuse();
-    let mut buf = iter.by_ref().take(2 * k).collect::<Vec<_>>();
+    let mut buf = iter
+        .by_ref()
+        .take(2usize.saturating_mul(k))
+        .collect::<Vec<_>>();
 
     if buf.len() < k {
         buf.sort_unstable_by(&mut comparator);
```

Given that the docs for `k_smallest_relaxed` explicitly call out that you need `2 * k * sizeof(Self::Item) + O(1)` memory, I'm wondering if this change is fine though -- since it explicitly uses more memory than `k_smallest` which is `k * sizeof(Self::Item) + O(1)`.

### LINKED ISSUE #1067: Peeking `usize::MAX` with `peek_nth`
[Running this playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=724625d75ea8a7d8366c4dfed0ab24f6), it looks like you can't `peek_nth` with an argument of `usize::MAX` without panicking.

In a way, this makes sense cause `peek_nth` gets you an item with index in `0..usize::MAX`, so getting the index `usize::MAX` itself is out of bounds. However, if you `peek_nth` past the end, this implementation returns None.  I assume the implicit contract is that it should always return an option, and not panic. 

It's pretty straighforward to special case `usize::MAX` and just return None if its provided as an argument to `peek_nth`. Otherwise, it can be kept as panicking and noted in the docs as an example.

### LINKED ISSUE #1068: `interleave_shortest`'s size_hint panics with large iterators.

patrickwehbe:
Three `Itertools` methods can panic with an overflow in debug builds when
given very large inputs. Each one does an unchecked `+ 1` or `* 2` on a
`usize` that can already be near the maximum. This switches them to
saturating arithmetic.

- `k_smallest_relaxed` (issue #1066): it allocates room for `2 * k` items,
  which overflows once `k` is greater than `usize::MAX / 2`. Now it uses
  `saturating_mul`, so a huge `k` saturates to `usize::MAX` and the whole
  iterator is collected, the same result you already get from `k_smallest`.
- `peek_nth` and `peek_nth_mut` (issue #1067): both compute `n + 1`, which
  overflows when `n == usize::MAX`. Now they use `saturating_add`, so
  peeking that far just returns `None`, which is what peeking past the end
  already does.
- `InterleaveShortest::size_hint` (issue #1068): the lower bound is computed
  as `combined_lower + 1`, which overflows when `combined_lower` is
  `usize::MAX`. Now it uses `saturating_add`. The upper-bound branch a few
  lines down already uses `checked_add`, so this just brings the two into
  line.

I added a regression test for each in `tests/test_std.rs` that calls the
affected method with `usize::MAX`-ish input and checks it returns instead of
panicking.

The reporter (@Takashiidobe) suggested the saturating approach in #1066;
the same idea applies cleanly to all three.

Fixes #1066
Fixes #1067
Fixes #1068

### DIFF
--- a/src/adaptors/mod.rs
+++ b/src/adaptors/mod.rs
@@ -179,7 +179,7 @@ where
         let (combined_lower, combined_upper) =
             size_hint::mul_scalar(size_hint::min(curr_hint, next_hint), 2);
         let lower = if curr_lower > next_lower {
-            combined_lower + 1
+            combined_lower.saturating_add(1)
         } else {
             combined_lower
         };
--- a/src/k_smallest.rs
+++ b/src/k_smallest.rs
@@ -99,7 +99,10 @@ where
     }
 
     let mut iter = iter.fuse();
-    let mut buf = iter.by_ref().take(2 * k).collect::<Vec<_>>();
+    let mut buf = iter
+        .by_ref()
+        .take(2usize.saturating_mul(k))
+        .collect::<Vec<_>>();
 
     if buf.len() < k {
         buf.sort_unstable_by(&mut comparator);
--- a/src/peek_nth.rs
+++ b/src/peek_nth.rs
@@ -69,7 +69,7 @@ where
     /// assert_eq!(iter.peek_nth(1), None);
     /// ```
     pub fn peek_nth(&mut self, n: usize) -> Option<&I::Item> {
-        let unbuffered_items = (n + 1).saturating_sub(self.buf.len());
+        let unbuffered_items = n.saturating_add(1).saturating_sub(self.buf.len());
 
         self.buf.extend(self.iter.by_ref().take(unbuffered_items));
 
@@ -110,7 +110,7 @@ where
     /// assert_eq!(iter.peek_nth_mut(1), None);
     /// ```
     pub fn peek_nth_mut(&mut self, n: usize) -> Option<&mut I::Item> {
-        let unbuffered_items = (n + 1).saturating_sub(self.buf.len());
+        let unbuffered_items = n.saturating_add(1).saturating_sub(self.buf.len());
 
         self.buf.extend(self.iter.by_ref().take(unbuffered_items));
 
--- a/tests/test_std.rs
+++ b/tests/test_std.rs
@@ -87,6 +87,16 @@ fn interleave_shortest() {
 }
 
 #[test]
+fn interleave_shortest_size_hint_does_not_overflow() {
+    // The combined lower bound can exceed `usize::MAX`, which used to overflow
+    // when computing `size_hint`. It should saturate instead of panicking.
+    let i = ::std::iter::repeat(0u8).take(usize::MAX);
+    let j = ::std::iter::repeat(0u8).take(usize::MAX - 1);
+    let it = i.interleave_shortest(j);
+    assert_eq!(it.size_hint(), (usize::MAX, None));
+}
+
+#[test]
 fn duplicates_by() {
     let xs = ["aaa", "bbbbb", "aa", "ccc", "bbbb", "aaaaa", "cccc"];
     let ys = ["aa", "bbbb", "cccc"];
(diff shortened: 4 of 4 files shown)


phimuemue:
Hi there,
can you close this and split these into different PRs so that we review them separately?

I only had a look at the `k_smallest_relaxed` and I am hesitant to include it as proposed:
- afaik `k_smallest_relaxed` benefits from the fact that it allocates `2*k` items:
  - it can cheaply (read `O(k)`) populate the entries at indices `k..2*k`
  - it can cheaply (iirc estimated `O(k)`) rearrange the entries so that it can throw away the upper `k` entries
- if the difference between the elements requested (`k`) and the entries reserved becomes smaller and smaller, this advantage vanishes.
As an extreme case consider `k==usize::MAX-1`. The difference between `k` and "entries reserved" is 1, so we would have to rearrange after each and every item put into the buffer, which might lead to `k^2` runtime.

If my assessment is correct, the better solution might be one of these in case `k` is too large:
- Delegate to the classical `k_smallest`.
- `panic!`, and document this.

Honestly, I would be ok with a documented panic, and see if we run into real-world scenarios where this is problematic.

PR #1111, opened 2026-06-23
Add `pairwise` as an alias for `tuple_windows` when iterating over a pair

### LINKED ISSUE #388: Add pairwise()

George-Ogden:
Closes #388, where most of the discussion is.

### Reasons for this pull request
- lots of people look for `pairwise` because they come from Python
- people reading Rust code will find `pairwise` more intuitive than `tuple_windows`
- it uses `tuple_windows` for the implementation to reduce duplication

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -832,6 +832,24 @@ pub trait Itertools: Iterator {
         tuple_impl::tuple_windows(self)
     }
 
+    /// Alias for `tuple_windows` when iterating over two elements at once.
+    /// ```
+    /// use itertools::Itertools;
+    /// let mut v = Vec::new();
+    ///
+    /// for (a, b) in (1..5).pairwise() {
+    ///     v.push((a, b));
+    /// }
+    /// assert_eq!(v, vec![(1, 2), (2, 3), (3, 4)]);
+    /// ```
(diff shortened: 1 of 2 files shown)


phimuemue:
Hi there, thanks for the effort. However, I'll close this:
- There's [`slice::array_windows`](https://doc.rust-lang.org/std/primitive.slice.html#method.array_windows)
- We lean towards *not* introducing new thin convenience wrappers.
- Unclear if `pairwise` is `tuple_windows` or `array_windows`[^1].

We may want to simplify the transition from Python's to Rust's itertools, but I think we could equally achieve this by providing a cheat sheet that maps Python -> Rust.

[^1]: imho `array_windows` is better than the possibly-to-be-deprecated `tuple_windows` (because adjustable by const generic and not up to tuple size).

PR #1114, opened 2026-07-14
Add TimedIterator that tracks durations of calls to next()

ktausch:
This pull request adds a new iterator adaptor named `TimedIterator` (accessed via `Itertools::timed()`). This iterator wraps each call to `next()` with duration tracking code. In particular, each call to `next()` is forwarded to the wrapped iterator until it returns `None`. In each of these cases, the return of the `<TimedIterator as Iterator>::next()` is `(inner_next_output, duration)` (where `inner_next_output` is an `Option`). Once these iterations are done, the `TimedIterator` will return `None` on every call to `next()` until it is dropped.

Items considered:

- [x] `TimedIterator` implements `Debug` and `Clone` (assuming the wrapped iterator does).
- [x] `TimedIterator` implements `FusedIterator`, notably even if the inner iterator does not.
- [x] `TimedIterator` cannot implement `ExactSizeIterator` even if the inner iterator does because it makes the adaptor is a longer iterator than the wrapped one (see [the docs for `ExactSizeIterator`](https://doc.rust-lang.org/std/iter/trait.ExactSizeIterator.html#when-shouldnt-an-adapter-be-exactsizeiterator) for details).
- [x] `TimedIterator` is lazy. This behavior is tested in `tests/laziness.rs` and the `must_use` macro decorates the struct.
- [x] `TimedIterator` is behind the `use_std` feature because it requires `Instant` and `Duration` from `std::time`.
- [x] The `Itertools::timed()` method consumes `self`.
- [x] The `Itertools::timed()` method is tested in `tests/test_std.rs` in the cases of (1) an infinite iterator (2) an empty iterator (3) a finite (implicitly fused iterator) and (4) a specially crafted unfused iterator

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -143,6 +143,8 @@ pub mod structs {
     pub use crate::take_while_inclusive::TakeWhileInclusive;
     #[cfg(feature = "use_alloc")]
     pub use crate::tee::Tee;
+    #[cfg(feature = "use_std")]
+    pub use crate::timed::TimedIterator;
     pub use crate::tuple_impl::{CircularTupleWindows, TupleBuffer, TupleWindows, Tuples};
     #[cfg(feature = "use_std")]
     pub use crate::unique_impl::{Unique, UniqueBy};
@@ -238,6 +240,8 @@ mod sources;
 mod take_while_inclusive;
 #[cfg(feature = "use_alloc")]
 mod tee;
+#[cfg(feature = "use_std")]
+mod timed;
 mod tuple_impl;
 #[cfg(feature = "use_std")]
 mod unique_impl;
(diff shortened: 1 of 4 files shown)


scottmcm:
*Specifically* coupling this to *timing* makes no sense to me.  It adds a `std` dependency and doesn't work if you want to use something other than `Instant`.  If this is worth doing, it'd IMHO need to be a more-generic "before and after" almost aspect-like adapter.  (I'd also consider whether the `(Option<T>, Duration)` is even a good idea, compared to say just `(T, Duration)` that can better forward calls, plus a `fn none_time(self) -> Option<Duration>` for the final call.)

But also you can get most of the value here from
```rust
let iter = std::iter::from_fn(|| {
    let start = Instant::now();
    let item = original_iter.next()?;
    let duration = start.elapsed();
    Some((item, duration))
});
```
and if `next` is expensive enough to be worth timing, I don't know that you're really getting anything from the extra effort of making a whole type.

phimuemue:
Hi there, thanks for the idea, but I agree with @scottmcm.

And, if we do this, I'm not sure if checking for `exhausted` in this iterator is wise. My guess would be: No, just keep pulling from the underlying iterator and do before/after stuff.

ktausch:
I made changes to the implementation in my fork of the repository that:
1. switch from yielding `(Option<T>, Duration)` to `(T, Duration)` with a `none_duration()` method on the `TimedIterator` class as mentioned by @scottmcm
2. Automatically implementing the `ExactSizeIterator` when the underlying iterator does (now that the `TimedIterator` yields elements one-to-one with its inner iterator)

However, I get that you guys aren't interested in the change, and that's ok! I just wanted to thank you for your feedback and let you know I'm using your suggestions in my code (I'll use `std::iter::from_fn` in situations where `none_time` isn't required, too).

> If this is worth doing, it'd IMHO need to be a more-generic "before and after" almost aspect-like adapter.

I'll think about this for the future and will let you know if I come up with something with a good API.

Thanks again, and cheers!

PR #1116, opened 2026-08-07
Fix typo in `array_windows()`

ronnodas:
The signs in the length were flipped, if the window is longer then there are fewer windows to yield. Now the formula also matches with the next remark about `N == 0`.


phimuemue (approved):
Thanks.

PR #1121, opened 2026-08-17
Improve Unique[By]/Duplicates[By] debug messages

FHTMitchell:
Updates the `Debug` impls of `Duplicates[By]` and `Unique[By]` to be more descriptive. Hides the implementation detail that `seen` is a `HashMap` under the hood for `Unique[By]` and removes the `meta` redirect in `Duplicates[By]`.

Open questions:

1. `Unique` doesn't actually get any benefit from having an `inner: UniqueBy` field since it has it's own `next` impl. Should we just lift those fields into the struct?
2. I renamed `used` to `seen` which feels like a more correct name, but I can revert. 
3. I made the `*By` structs use `finish_non_exhaustive` which adds a `..` to the representation. That's not something anything using the macro gets but useful information I feel. Might be worth adding a `..` option to the macro?
4. Should the values in `seen` in `Duplicates[By]` become a 2-valued enum (not a bool) to make it clearer what's going on? Probably not worth it...

```
 // Before

  Unique { iter: UniqueBy { iter: Iter(["b", "c"]), used: {"a": ()} } }
  UniqueBy { iter: Iter(["b", "c"]), used: {1: ()} }
  DuplicatesBy { iter: Iter([3]), meta.used: {1: false, 2: true} }
  DuplicatesBy { iter: Iter([3]), meta.used: {0: true, 1: false} }

// After

  Unique { iter: Iter(["b", "c"]), seen: {"a"} }
  UniqueBy { iter: Iter(["b", "c"]), seen: {1}, .. }
  Duplicates   { iter: Iter([3]), seen: {1: false, 2: true}, pending: 1 }
  DuplicatesBy { iter: Iter([3]), seen: {0: true, 1: false}, pending: 1, .. }

### DIFF
--- a/src/duplicates_impl.rs
+++ b/src/duplicates_impl.rs
@@ -19,13 +19,34 @@ mod private {
         pub(crate) meta: Meta<Key, F, S>,
     }
 
-    impl<I, V, F, S> fmt::Debug for DuplicatesBy<I, V, F, S>
+    impl<I, V, S> fmt::Debug for DuplicatesBy<I, V, ById, S>
     where
         I: Iterator + fmt::Debug,
         V: fmt::Debug + Hash + Eq,
         S: BuildHasher,
     {
-        debug_fmt_fields!(DuplicatesBy, iter, meta.used);
+        fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+            f.debug_struct("Duplicates")
+                .field("iter", &self.iter)
+                .field("seen", &self.meta.seen)
+                .field("pending", &self.meta.pending)
+                .finish()
+        }
+    }
+
+    impl<I, V, G, S> fmt::Debug for DuplicatesBy<I, V, ByFn<G>, S>
+    where
+        I: Iterator + fmt::Debug,
+        V: fmt::Debug + Hash + Eq,
+        S: BuildHasher,
+    {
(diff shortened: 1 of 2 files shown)


phimuemue:
Hi @FHTMitchell, thanks for this. I'm gonna use some maintainers discretion here:

- Let's not start writing custom `Debug` implementations. They are for developers, and developers can be assumed to parse them if necessary. I only want to diverge if we have evidence that developers desperately need "better" `Debug` implementations.
 If you want to tackle this, please adjust `impl<I> Debug for ExactlyOneError<I>` accordingly so that it does what everyone else is doing. (If it bothers you: Eventually `ExactlyOneError` should have something nicer than `Option<Either<[Item;2], Item>>` e.g. something like an `Array<2>::Iterator`.) 
- `used` -> `seen`: Fine with me, but please in a separate commit. More important is uniformity between `Unique` and `Duplicates`.
- `finish_non_exhaustive`: If this is really important enough, then it is not specific for `Unique[By]` and should be part of `debug_fmt_fields!`. If I am not mistaken, we prefer `Derive(Debug)` and use `debug_fmt_fields` precisely where we do not have `Debug` for all fields. So maybe `debug_fmt_fields` should have it?
- `bool` vs bespoke `enum`: I just skimmed the code and I think `bool` is ok.

PR #1122, opened 2026-08-20
Derive `Copy` for `EitherOrBoth`

ronnodas:
It's been missing since the original implementation but I see no reason it shouldn't be `Copy` when both `A` and `B` are.


phimuemue:
Result implements Copy, too, so let's follow that. Thanks!

phimuemue (approved):

Issue #595, opened 2022-01-06
only / single


gilescope:
https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.single?view=net-6.0

is there a way to do a filter and err if the result of the filter is not exactly one (ie error/none if not there or if many)? Quite often we want / expect that there is only one of something.

phimuemue:
Do you know about 'exactly_one'? Using it, you could do 'iterative.filter(predicate).exactly_one()'?

gilescope:
Apparently not :-) - many thanks!

Issue #566, opened 2021-07-29
Feature request: single.  


JCBurnside:
I propose two additional functions.

`fn single(self) -> Result<Self::Item, MutlipleResultError>` which will return Ok(Self::Item) if there is a single element in the iterator.

and 

`fn filter_single<F>(self, filter:F) -> Result<Self::Item, MutlipleResultError> where F: Fn(&Self::Item)->bool` which will be basically an alias for `self.filter(filter).single()`

Reasoning:
I am basing this off [C# linq's single function](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.single?view=net-5.0) which will throw an error if there is more than a single element in the enumerable.  There are cases where I want there to be a single element in the iterator and if multiple are found that is a logical error.

phimuemue:
Hi! Isn't `single` the same as [`exactly_one`](https://docs.rs/itertools/0.10.1/itertools/trait.Itertools.html#method.exactly_one)? (To be fair, I think `single` may be a better name.)

As for `filter_single`, I do not know if this shortcut is worth an own function, as this rises the question why `filter` in connection with `single`/`exactly_one` is more special than e.g. `filter` with other functions. Is it a well known or very common pattern to do `filter().single()`?

JCBurnside:
Welp I had missed exaclty_one when looking through the docs.   The `filter_single` is based of an overload of c#'s single which takes a function.

scottmcm:
Note that Rust generally doesn't need the special-cased terminals the way C# does.  For example, C# has an `Enumerable.Count` overload that takes a predicate, but `.filter(p).count()` in Rust works well for that -- thanks to overloadable static dispatch, there's a special codepath to optimize for that combination: https://doc.rust-lang.org/1.45.0/src/core/iter/adapters/mod.rs.html#1012-1019

JCBurnside:
I see.  Perhaps should consider a rename of `exactly_one` to `single`

Issue #334, opened 2019-02-25
single?


phimuemue:
I find myself repeatedly in need of a function that determines if an iterator yields a single element.

I found a crate doing this: https://github.com/CAD97/rust-single and also [asked](https://github.com/CAD97/rust-single/issues/1) the maintainer if we should add this functionality to itertools. What do you think about this?

bluss:
collect tuple can do this (using a singlet), but it makes sense to give it its own name.

phimuemue:
Good point, thanks.

Still, I still do feel that - in case of failure - it might be advantageous to know the reason.

Do you think changing from `Option<...>` to `Result<..., CollectTupleErr>` (where `CollectTupleErr` is an `enum` with `Empty` and `TooMany`) is a viable option?

bluss:
I think I recognize this discussion, not sure from which issue. Do you want to help maintain this crate?

phimuemue:
Ah, I see that there was some previous discussion at https://github.com/bluss/rust-itertools/pull/233#issuecomment-342578731. So I think I'll revisit my use case and see if I really need a `Result` carrying additional information or if I can get away with an option.

As for the maintainership offer: Thank you for thinking of me, but I think I wouldn't make a good maintainer at the moment (having a full-time job and a time-consuming hobby), so I have to politely decline.

phimuemue:
I just wanted to give an update on the `Option` vs `Result` issue here: In many uses cases, I would actually benefit if I got a `Result` telling me why `single` did not work out.

In particular, I often want to allow "no element or one element only", but not multiple elements.

I see that we do not really have methods returning `Result`, but we have e.g. `MinMaxResult` which may be a similarly special.

phimuemue:
I just observed that we have `exactly_one`, which I can probably use. Though it offers the contents of the whole original iterator, its usage seems imho a bit worse than what was originally suggested in this issue.

Closing this for now, as it covers my basic request.

phimuemue:
``

PR #1125, opened 2026-08-31
Add doc alias "single" to "exactly_one"

nicopap:
"Single" seems like an obvious name for "exactly_one", so I feel like adding this alias makes it more discoverable.

It's notable that "single" doesn't appear in the documentation for exactly_one either.

I found a few issues that show that indeed this change might help people:

* https://github.com/rust-itertools/itertools/issues/595
* https://github.com/rust-itertools/itertools/issues/334
* https://github.com/rust-itertools/itertools/issues/566


phimuemue (approved):
Very good, indeed. Thanks.

